Issue #119, opened 2014-10-11
Integrate with the official julia-mode rather than replacing it


PythonNut:
An official `julia-mode` already exists. Would it be possible for ESS to simply extend the origional instead of replacing it entirely?

(This arose as a product of [JuliaLang/julia #8635](https://github.com/JuliaLang/julia/issues/8635))

mturok:
@vspinu - I see that @Wilfred submitted a pull request to unify ESS and the official julia-mode some time ago:  https://github.com/emacs-ess/ESS/pull/122

Anything one can do to get this moving forward?

Thanks,
Michael

mmaechler:
IIUC, your change's consequence would be that a julia + emacs-ess user would also need to install  "julia-mode" from the julia sources / website (or other ways such as MELPA)? 
I'm just asking... if it is true, I'd want to make sure that would be as seemless as possible ... and for now, 
(where julia is still quite a bit of a moving target!)  even consider bundling  julia-mode with the ESS tarball in our releases.

@wilfred, @mturok : I have added two "inline" comments in your file changes, can you have a look?

vspinu:
Sorry folks. I got completely overwhelmed with two jobs and a reconstruction of the new house.  It's getting a bit easy on my side these days and II hope to look at it soon now.

> 'm just asking... if it is true, I'd want to make sure that would be as seemless as possible

It's true. If ESS is installed from MELPA the julia dependency will be pulled automatically.  We might consider having a contrib directory with up to date julia mode as a fall-back in case machine default julia mode is not has not been found.

mturok:
@vspinu Not at all - and congratulations with the new house.    Happy to lend a hand....but saw that some tests were failing currently (and looks like the pull request is on a very old revision).

mmaechler:
fixed by 7bfcbbda4146e7375a48c4e1fba3582784c49b8f

PR #122, opened 2014-11-20
Depend on julia-mode.el

Wilfred:
This PR changes ESS to use julia-mode.el for syntax highlighting and indentation, fixing #119. It also changes the ESS Julia commands to consistently use an `ess-` prefix to avoid name collisions.

I'm completely unfamiliar with the ESS codebase, so please let me about any/all the silly mistakes I've made and what we need to get this merged.

Thanks :)


spencerlyon2:
Ping.

I would really like to see this happen! Any way I can help?

Wilfred:
I believe these changes are suitable for being merged, FWIW. I'm not aware of any outstanding issues.

gragusa:
I would love too see this  to happen as well.

mturok:
Anything that can be done to help here?

lisp/ess-julia.el
```diff
@@ -281,19 +71,19 @@ VISIBLY is not currently used."
...
+(defun ess-julia-manual-lookup-function (&rest args) ; args are not used
   (interactive)
   "Look up topics at http://docs.julialang.org/en/latest/manual/"
   ;; <li class="toctree-l1"><a class="reference internal" href="introduction/">Introduction</a></li>
-  (let* ((pages (or julia--manual-topics
-                    (setq julia--manual-topics
-                          (julia--retrive-topics "http://docs.julialang.org/en/latest/manual/"))))
+  (let* ((pages (or ess-julia--manual-topics
+                    (setq ess-julia--manual-topics
+                          (ess-julia--retrive-topics "http://docs.julialang.org/en/latest/manual/"))))
```
mmaechler: Hmm.... you mean to use help in julia's  ess-integrated mode one would access the internet ?

No way now,.... maybe in ten years.... but honestly I'd even hate it then.
Hej, this is free software so with free docs...  and that free means I want the relevant stuff on my hard disk.

mmaechler:
Reply to my "inline" (code line) comments?

lisp/ess-julia.el
Wilfred: I didn't add this. I only renamed the function. This functionality is already part of ESS: https://github.com/emacs-ess/ESS/blob/e16e13d36a1e3fce928d1434f07e0dfd96cdf218/lisp/ess-julia.el#L294

If you'd like to remove it, it's your code :). I don't think it's relevant to this pull request though.

Wilfred:
>  We have M-x R and so I want to keep M-x julia 

OK, I'll add an alias. I was just worried about conflicts with `julia-mode` itself, as ESS doesn't own the julia prefix. No-one else currently defines a `julia` command so it should be fine.

Wilfred:
OK, I've merged ESS master into my branch, so this is ready to merge! Let me know if I've missed anything.

mturok:
@Wilfred @vspinu  Thank you!   Looking forward!

lisp/ess-julia.el
mmaechler: @Wilfred : you are right.. it's already there {though I'm still puzzled about how github does show it - to my eyes - as if it was new.  Then, there's also `-retrive-` which should really be `-retrieve-` ... but again that's not part of your pull request.

spencerlyon2:
:fireworks: :tada:

vspinu:
Thanks @Wilfred, but this will break MELPA users with missing julia-mode.el. 

I haven't had a chance to look at the new changes. I assume it's working for you.

vspinu:
> but this will break MELPA users with missing julia-mode.el. 

Never mind, I just fixed that (hopefully). 

It will still break all other users installing directly, but that we can fix later. ESS has been always self contained.

mmaechler:
@vspinu :  I have added   'make julia-mode.el'  and as a target (gets it from the julia sources)
and also added the file to the tar ball.
So whenever we *release* ESS in the traditional way, we will have a current version bundled.

There are other problems though.. the look and feel is still worse than in the terminal, e.g., because name completion (in *julia*, i.e. the inferior julia process buffer) does not work...
and other things ... "as always with julia"  in my experience.

vspinu:
> So whenever we release ESS in the traditional way, we will have a current version bundled.

This is not ideal. We don't want to overload julia-mode.el if it's already installed on the machine. So we need to have a conditional load in `ess-site.el`. 

We need to have  julia-mode.el (and probably others in the future) as a git submodule in contrib folder.   

>  because name completion (in julia, i.e. the inferior julia process buffer) does not work...

Everything was working when I originaly wrote it. Someone has to maintain `ess/ess-julia.jl` or ask julia folks to provide that functionality in the core and don't break it in the future.

mmaechler:
because name completion (in julia, i.e. the inferior julia process buffer) does not work...

> Everything was working when I originally wrote it. Someone has to maintain etc/ess-julia.jl or ask julia > folks to provide that functionality in the core and don't break it in the future. 
Indeed, @vsipnu .. I've been very grateful for your work  ((maybe should have said it louder at the time..)!
But, with which version of julia did it work? ;-)   They have been changing julia so often..
Also the released version tends to be ok...  If you want to use extensions (say in statistics) it is always "the next" version that you are supposed to use instead of the released one.

I have not been a very regular user of the ESS julia interface, but have tested it many times, and I think it never worked perfectly for both "julia released" and "julia devel" in all the cases I had tried.
At the moment object completion seems to fail for both versions of julia I tried.

Apropos @vspinu: have you seen the discussion about the issue that our ESS part of the interface currently seems to need internet access ... which I'm finding very wrong.

vspinu:
> At the moment object completion seems to fail for both versions of julia I tried.

That has been broken for a while. Each time I fix it, next time they change their help system and everything fails again. I did it like 3 times already and I am now so out of julia that I have no immediate plans to fixit again. The code in ess-julia.jl is trivial. Anyone with basic of knowledge of julia should be able to fix that in minutes. 

>  @vspinu: have you seen the discussion about the issue that our ESS part of the interface currently seems to need internet access ... which I'm finding very wrong.

That discussion was about ESS pinging some external server on starup. I don't see how this is possible. We didn't receive a confirmation for that. Some functionality, like apropos search are built to work through the web but those are user commands.

vspinu:
> We need to have julia-mode.el (and probably others in the future) as a git submodule in contrib folder. 

It's actually not possible to get only one file as git subtree or submodule due to how git work. So I am leaving this as it is. Those installing from source will have to install `julia-mode.el` from MELPA.

mmaechler:
Well, I know you don't like "make"  but in this case you should see how I get a current julia-mode.el in the Makefile.... and also add it to the tarball.
Instead of in ./lisp/ you may put it in ./lisp/contrib/ or whatever and then ensure we load it conditionally only in the case where there is no "system" julia-mode.el

vspinu:
> Instead of in ./lisp/ you may put it in ./lisp/contrib/ or whatever and then ensure we load it conditionally only in the case where there is no "system" julia-mode.el

I have done that half way and then abandoned it. There are complications with `make install` which I don't have time to figure out right now. We basically need to track yet another directory (contrib) (besides `lisp` and `etc`) without including it into the load path.

lisp/ess-julia.el
```diff
@@ -39,224 +39,14 @@
 
 (require 'compile); for compilation-* below
 (require 'ess-utils)
+(require 'julia-mode)
```
lionel-: @vspinu @mmaechler 

The code for generating tests stopped working because it cannot load `julia-mode`. I installed the package but it still cannot find it.

Would it be reasonable to wrap this as follows?
```lisp
(unless (bound-and-true-p 'ess-test-script)
  (require 'julia-mode))
```
Or is it a problem with my Emacs installation that I should solve?

vspinu: If you installed from MELPA then the test generation code will not work.  

Don't know how to properly go about it. Can you for now run `make julia-mode.el` from `ESS/lisp/`?

We can also do `(when (locate-library "julia-mode.el") ..)` for a conditional install.

lionel-: > Don't know how to properly go about it. Can you for now run make julia-mode.el from ESS/lisp/?

alright, thanks for the easy workaround.

lionel-: And I just added `julia-mode.el` to `.gitignore` so it can stay there and not interfere with git.

spencerlyon2:
I just tried a fresh install of ess and julia-mode and am seeing `ESS[julia-derived]` as the major mode, but am unable to do unicode substitution (e.g. `\alpha[TAB]` doesn't replace with a unicode alpha). 

Is this a bug or known limitation, or am I doing this wrong somehow?

PR #147, opened 2015-02-17
Fix bug in fill paragraph with roxy

davidshepherd7:
Currently if you attempt to call fill-paragraph with point in position `X` below

    #'X

an error

    Invalid search bound (wrong side of point)

occurs. This is problematic when using [aggressive-fill-paragraph](https://github.com/davidshepherd7/aggressive-fill-paragraph-mode) which continuously attempts to fill comment paragraphs.

This patch fixes the issue.


vspinu:
Thanks! Looks good to me.

PR #150, opened 2015-03-04
Add r-completion.elc to Makefile

dickoa:
ess-r-completion is not byte-compiled therefore not sent to /usr/share/emacs/site-lisp for people using make install it creates some problem

### DIFF
--- a/lisp/Makefile
+++ b/lisp/Makefile
@@ -48,7 +48,7 @@ ELC = $(CORE) ess-comp.elc ess-custom.elc \
 	ess-omg-l.elc ess-omg-d.elc \
 	ess-bugs-l.elc ess-bugs-d.elc ess-jags-d.elc \
 	ess-noweb.elc ess-noweb-mode.elc ess-noweb-font-lock-mode.elc \
-	ess-eldoc.elc ess-roxy.elc ess-rutils.elc \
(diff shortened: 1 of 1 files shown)


mmaechler:
Thank you. 
The reason this was never detected: It seems that all ESS core developers work out of the build directory and never use 'make install' ... which is at least true for me. ;-)

vspinu:
Thanks Ahmadou. I always keep forgetting about makefiles :(

mmaechler:
Actually.  There's more broken:

```
cd lisp
\rm -f *.elc
make
```
very quickly ends in an error that  `cl-lib is not found .... which is a bit strange, as interactively it is found well  when   (require 'cl-lib)   is encountered inside ess-completion.el

I'd be grateful for another patch ... as I'm not detecting where the problem is  (Emacs 24.3).

Martin

dickoa:
I can't reproduce this error on a fresh install from the latest commit.

My specs are the following
Emacs version: GNU Emacs 24.4.1
Linux version (Arch Linux): Linux dickoa 3.18.6-1-ARCH #1 SMP PREEMPT Sat Feb 7 08:44:05 CET 2015 x86_64 GNU/Linux

But I will continue to dig further into this but

vspinu:
I think `cl-lib` was introduced in 24.4. I have changed it to the old `(require 'cl)`.

PR #151, opened 2015-03-18
Make buffer display customizable

sam-s:
* lisp/ess-inf.el (ess-show-buffer-action): Add user option.
(ess-show-buffer): Use it instead of a hard-coded value.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1043,6 +1043,11 @@ If TOGGLE-EOB is given, the value of
 ;;       (display-buffer buf nil (get-frame-for-buffer buf))
 ;;     (display-buffer buf nil t)))
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(defcustom ess-show-buffer-action
+  '((display-buffer-use-some-window display-buffer-pop-up-window))
+  "Actions for `ess-show-buffer', passed to `display-buffer'."
+  :group 'ess
+  :type 'list)
 (defun ess-show-buffer (buf &optional visit)
   "Ensure the ESS buffer BUF is visible.
 The buffer, specified as a string, is typically an iESS (e.g. *R*) buffer.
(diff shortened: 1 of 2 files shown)


vspinu:
:+1:

lisp/ess-inf.el
```diff
@@ -1043,6 +1043,11 @@ If TOGGLE-EOB is given, the value of
 ;;       (display-buffer buf nil (get-frame-for-buffer buf))
 ;;     (display-buffer buf nil t)))
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(defcustom ess-show-buffer-action
+  '((display-buffer-use-some-window display-buffer-pop-up-window))
```
vspinu: You have reverted the order of actions. People already started complaining of the changed behavior. I didn't notice that during the merge. What was the reason for the change?

sam-s: > * Vitalie Spinu <abgvsvpngvbaf@tvguho.pbz> [2015-03-26 09:10:48 -0700]:
>
>> @@ -1043,6 +1043,11 @@ If TOGGLE-EOB is given, the value of
>>  ;;       (display-buffer buf nil (get-frame-for-buffer buf))
>>  ;;     (display-buffer buf nil t)))
>>  ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
>> +(defcustom ess-show-buffer-action
>> +  '((display-buffer-use-some-window display-buffer-pop-up-window))
>
> You have reverted the order of actions. People already started
> complaining of the changed behavior. I didn't notice that during the
> merge. What was the reason for the change?

The new behavior is closer to the standard Emacs behavior with other
inferior modes (lisp, python &c)

-- 
Sam Steingold (http://sds.podval.org/) on darwin Ns 10.3.1344
http://www.childpsy.net/ http://think-israel.org http://jihadwatch.org
http://mideasttruth.com http://camera.org http://dhimmi.org http://ffii.org
To avoid fatigue, one must sleep 8 hours every day. And 8 hours every night too.

PR #153, opened 2015-03-20
Fixed install failure and missing file

dkulp2:
Not sure if this is specific to Mac, but doc was missing the .dvi files after make and also there was no README or ANNOUNCE, so make install was failing.  I just removed them.
Also, there's a julia.jl file in etc, that wasn't installed, but needed when calling `julia`.


vspinu:
I think it's indeed mac/tex distribution specific. Thanks.

PR #155, opened 2015-04-05
make-variable-buffer-local should not be called at other than top-level

syohex:
I got following warning with original code.
```
In msdos-minor-mode:
msdos.el:149:9:Warning: `make-variable-buffer-local' not called at toplevel
```


vspinu:
I didn't even know that we have `msdos.el` in ESS :). Thanks.

PR #168, opened 2015-05-02
Implement ess-arg-function-offset-continued

lionel-:
Hello,

This implements `ess-arg-function-offset-continued`, a companion to `ess-arg-function-offset-new-line`. While the latter acts when no function arguments have been entered, the new setting allows to control the indentation for continuations of an argument list.

The behaviour is similar to existing parameters:
```r
# nil (the default, same indentation as before the patch)
fun(arg1,
    fun2(arg2,
         arg3
    ))

# 2
fun(arg1,
      fun2(arg2,
             arg3
      ))

# '(2)
fun(arg1,
  fun2(arg2,
    arg3
  ))
```

This feature was [requested] (https://stat.ethz.ch/pipermail/ess-help/2011-December/007344.html) a while ago on the mailing list. Also, setting this parameter to `'(2)` produces code in line with Hadley Wickham's coding style.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -671,12 +671,12 @@ This is in addition to ess-continued-statement-offset.")
 (defvar ess-arg-function-offset 2
   "Extra indent for internal substatements of function `foo' that called
 in `arg=foo(...)' form.
-If not number, the statements are indented at open-parenthesis following foo.")
+If not number, the statements are indented at open-parenthesis following `foo'.")
 
 (defvar ess-arg-function-offset-new-line '(2)
   "Extra indent for function arguments when ( is folowed by new line.
 
-If nil, the statements are indented at open-parenthesis following foo:
+If nil, the statements are indented at open-parenthesis following `foo':
 
   a <- some.function(other.function(
                                     arg1,
@@ -686,16 +686,16 @@ If a list of the form '(N) where N is a number, the statements
 are indented at the previous line indentation + N characters:
 
   a <- some.function(other.function(
-     arg1,
-     arg2)
+      arg1,
+      arg2)
 
 
 If a number N, the statement are alligned at the beginning of
 function call + N characters:
 
   a <- some.function(other.function(
-                       arg1,
-                       arg2)
+                         arg1,
+                         arg2)
 
 
 For inner function arguments the behavior is unchanged:
@@ -717,6 +717,39 @@ some.function(arg1,
                      a,
                      b,
 
+")
+
+(defvar ess-arg-function-offset-continued nil
+  "Extra indent for function arguments when ( is folowed by arguments and
+a new line. This setting is mainly intended for compatibility with Rstudio
+indentation rules. The default (nil) is recommended.
+
+If nil, the statements are indented at open-parenthesis following `foo':
+
+  a <- some.function(other.function(arg1,
+                                    arg2)
+
+If a list of the form '(N) where N is a number, the statements are indented
+at the previous line indentation + N characters, as in Rstudio:
+
+  a <- some.function(other.function(arg1,
+      arg2)
+
+
+If a number N, the statement are alligned at the beginning of
+function call + N characters:
+
+  a <- some.function(other.function(arg1,
+                         arg2)
+
+Note that this setting only applies to function calls, not functions
+declarations. For declarations, arguments are always aligned with
+the opening parenthese:
+
+  a <- function(arg1,
+                arg2) {
+      body
+  }
 ")
 
 ;;added rmh 2Nov97 at request of Terry Therneau
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -415,6 +415,9 @@ Variables controlling indentation style:
    `foo'.
  `ess-arg-function-offset-new-line'
    Extra indent for function arguments when ( is folowed by new line.
+ `ess-arg-function-offset-continued'
+   Extra indent for function arguments when ( is folowed by arguments and a
+   new line.
  `ess-expression-offset'
     Extra indent for internal substatements of `expression' that specified
     in `obj <- expression(...)' form.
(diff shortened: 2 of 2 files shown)


eddelbuettel:
I honestly do not understand why anyone would want style `# 2` above, let alone `# 3` which is equally terrible but more condensed.  

I guess if you waste all those columns by inserting underscores everywhere, you feel forced to make it up via two-column indenting.  Textbook definition of a 'loose-loose' situation ...

I am half-joking, of course, but I really to think that mode `# 1` is the best thing ever for R code.  And a lot of existing code seems to agree.  But you can't argue style or preferences...

lionel-:
I added `#2` because it was trivial to implement and it makes this setting consistent with existing parameters.

Emacs is about customisation to personal preference. It seems to me that ESS should be able to write the kind of code used in all those Hadley packages.

(If we really need to compare coding style, I think camelCase code has often longer function names than under_score code because of this Java-influenced culture of having thoroughly explicit names.)

eddelbuettel:
> I think camelCase code has often longer function names

That conjecture would need empirics for a real fight. I salute you for refering to Java, I tend to do that too when I want to annoy people.  But my camelCase comes from years (decades) of C/C++ where it is common.  Even Hadley acknowledged (in private chat) that he uses camelCase there and that it was seeping into some R he wrote as he had been doing so much C++ lately...

And just to be plain, for the same verb and noun, `doThing` will always be shorter than `do_thing`. 

Now let's put this to rest.  As you say, Emacs is all about customization, and luckily the people setting default values have decent taste ;-)

lionel-:
Anyway, I configured emacs to display dashes instead of underscores :-)

eddelbuettel:
:+1:

vspinu:
> Emacs is about customisation to personal preference.

Well, while this is certainly true, indentation would better be standard across people and systems.

> produces code in line with Hadley Wickham's coding style

It's not part of @hadley style guide but rather how R Studio indents it.  I completely agree with @eddelbuettel that  `#2` and `#3` are rather terrible examples of indentation. I don't think we want to promote that. 

@rstudio indents differently arguments in function declarations and calls.   See for example @hadley's [guide](http://r-pkgs.had.co.nz/style.html) where arguments in `function` declaration are indented at the opening `(`:

```
long_function_name <- function(a = "a long argument", 
                               b = "another argument",
                               c = "another long argument") {
  # As usual code is indented by two spaces.
}
```

So I would say it's an inconsistency with R studio indentation.

lionel-:
yes but in the mean time I have a hard time contributing to those packages because Emacs won't indent my code according to the house rules.

Besides, ESS indentation code is very flexible and allows far more ugly style in other areas.
(Plus, I've grown to like this kind of indentation and I don't agree it's so horrible.)

lionel-:
I don't want to use Rstudio to contribute code to some package. For better or worse, it's become part of idiomatic R.

lionel-:
Vitalie, you had a good point about the difference between function calls and function declarations. This is now fixed.

```r
fun <- function(arg1,
                arg2) {
  stuff
}

call(long_argument = "long stuff",
  long_argument2 = "still more stuff")
```

lisp/ess-custom.el
```diff
@@ -686,8 +686,8 @@ If a list of the form '(N) where N is a number, the statements
 are indented at the previous line indentation + N characters:
 
   a <- some.function(other.function(
-     arg1,
```
vspinu: Why has this indent been changed?

lisp/ess-custom.el
```diff
@@ -719,6 +719,38 @@ some.function(arg1,
 
 ")
 
+(defvar ess-arg-function-offset-continued nil
+  "Extra indent for function arguments when ( is folowed by arguments and
+a new line.
+
+If nil, the statements are indented at open-parenthesis following foo:
```
vspinu: Would you please add a note here that this variable is here only for the compatibility with R studio and is not recommended by ESS developers. 

Also please enclose "foo" in quotes of some kind.

lisp/ess-custom.el
lionel-: I thought it was a documentation mistake. The original example had indents of 3 spaces whereas 2 spaces are used in the rest of the examples.

Or am I missing something?

vspinu:
I am still very hesitant about merging this. ESS currently honors `ess-arg-function-offset-new-line` which indents as R-studio but only if you "newline" immediately after `(`: 

```
a <- some.function(other.function(
    abc = 1,
    xyz= 2)
```
I think this should be enough if the host of the project that you contribute to really insists on such indentation.

The need to split arguments occurs only with many arguments or long argument names. So ending with `(` is really not a big deal. It's also considerably more readable.  Compare:

```
a <- some.function(other.function(abc = 1,
    xyz= 2)
```

So, even in this case, ESS indentation is consistent and versatile enough to fit your need. We are already customizing indentation much more than any other system out there and having many more bugs with corner cases as a result of that. I really would prefer not to manage yet another quirky style. 

Let's see that other people think.

lisp/ess-custom.el
lionel-: should I also enclose foo in other docstrings?

vspinu: Ah. I see. It was a mistake. It should be 4 (the default).

lisp/ess-custom.el
vspinu: Ah. I see. It was a mistake. It should be 4 (the default).

lisp/ess-custom.el
vspinu: Yes, please enclose. You can use backtick to emphasize that it's a code symbol without highlighting it as a lisp symbol.

eddelbuettel:
If RStudio standards really mattered, could you not simply use Yihui's formatr package?

I also don't think that the conventions are that rigid--for the various Rcpp-related projects, JJ, Kevin and I never had a problem.

lionel-:
> I think this should be enough if the host of the project that you contribute to really insists on such indentation.

I think it is not enough because I often need to keep my code manually indented, which is a real chore...

I understand completely the point about code complexity, but I would say it is worth it. Besides the issue of contributing to other packages, there is also the issue of collaborating with people. What if most of your colleagues or your boss use Rstudio? Since Rstudio is probably the most used R editor, it's important to be able to write its R idiom smoothly.


> If RStudio standards really mattered, could you not simply use Yihui's formatr package?

That is a terrible user experience... I just want to write code.

lisp/ess-custom.el
lionel-: Here you go. I used an opening backtick and a closing quote as I saw this was used elsewhere.

kevinushey:
When we talk about RStudio indentation styles, there's really only one 'knob' that you can turn -- either arguments to functions are vertically aligned, or they're not. I believe we default to 'on' for vertical alignment now, but of course users are free to switch that off.

With vertical alignment on, similar to what ESS does now, RStudio only does vertical alignment if there is at least one argument following the opening `(`. (or `[`, `[[`, for that matter). This allows a 'fall-back' to no-vertical-alignment even when that's on, since it's unlikely that users want to write e.g.

```
someTerriblyLongFunctionName(
                             n = 1
)
```

So, IIUC, this PR intends to support the case of RStudio-style indentation where vertical alignment of function arguments is off.

Personally, when I first began writing R code, I preferred no vertical alignment (felt like I was wasting whitespace); now I much prefer it as it keeps things more readable. I know we have users who keep that knob off and prefer the 'less-whitespace' indentation rules.

I do agree that the Emacs philosophy here means making more indentation rules available -- it's definitely not unheard of; e.g. C / C++ indentation is very customizable in Emacs. Of course, I am not an ESS maintainer / regular contributor so it's not my choice to make.

Anyway, that's my two cents as the current 'maintainer' of indentation in RStudio...

eddelbuettel:
Three cheers to @kevinushey for piping in.

lionel-:
Development continues in #169

PR #171, opened 2015-05-03
[Fix #170] Indentation of continuation lines inside braces

### LINKED ISSUE #170: Indentation of continuation statements inside braces
`ess-continued-statement-offset` is not taken into account inside braces:
```r
# outside braces
first %>%
  second %>%
  third

# inside braces
{
  first %>%
    second %>%
      third
}
```

lionel-:
This is a really simple fix that looks a bit naive but I cannot find any side effect.

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -1052,12 +1052,12 @@ Return the amount the indentation changed by."
 (defun ess-calculate-indent--closing-paren ()
   (search-forward ")")
   (backward-sexp)
-  (or 
+  (or
    ;; If the cursor is in between parents we indent as normal text to avoid
    ;; annoyance with paired parents
    (and (looking-at-p "[( \t\n]+)")
         (ess-calculate-indent--after-last-open-paren))
-   
+
    (cond ((numberp ess-close-paren-offset)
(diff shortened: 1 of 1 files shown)


vspinu:
> ess-continued-statement-offset is not taken into account inside braces:

I am afraid I don't see that. 

I myself have 

````
(add-to-list 'ess-style-alist
             '(my-style
               (ess-indent-level . 4)
               (ess-first-continued-statement-offset . 2)
               (ess-brace-offset . -4)
               (ess-expression-offset . 4)
               (ess-else-offset . 0)
               (ess-close-brace-offset . 0)
               (ess-continued-brace-offset . 0)
               (ess-arg-function-offset . 4)
	       (ess-arg-function-offset-new-line . '(4))
               ))

```

which indents as in your first example, but also correctly indents within braces. 

Also the change that you propose is not relevant to the indentation at all. `ess--continued-statement` still returns nil after `{` because all conditions except the last `(t nil)` are not satisfied. The first condition is simply a shortcut not to have those regexp matches in the second condition.

lionel-:
oooh it was all because of that redefinition of `ess--continued-statement` that was hidden in one of my config files since a very long time and that I had completely forgotten about until yesterday...

sorry about that!

Issue #174, opened 2015-05-05
Test cases for indentation


lionel-:
Since the indentation code is rather meaty, it is difficult to change it without breaking things. It would be great to have systematic tests for indentation to reduce the probability of regressions. I will write a PR for this later this week.

I put together this collection of test cases, but I probably forgot many things. Feel free to complete the list:
```r
### Function declarations

fun <- function(arg,
                arg) {
    body
}

function(
    arg,
    arg
    )
    {
        body
    }

function(arg_fun(sub_arg,
                 sub_arg),
         arg) {}

function(arg, x = fun(
                  arg),
         arg) {}



### Function calls

fun_call(arg,
         arg)

fun_call(
    arg,
    arg
    )

fun_call(arg = (
    stuff         # bug?
    ),            #
         arg)

fun_call(arg = fun(
             stuff
    ),
         arg)



### Blocks

function()
    {
        body
    }

fun_call(arg = {
             stuff
         },
         {
             stuff
         }, arg = {
             stuff
         }, {
             stuff
         }
         arg)

fun <- call({
    stuff
}, {
    stuff
},
            {          # bug
                stuff  #
            }          #
            )          #

fun_call(function(x) {
             body
         },
         function(x) {
             body
         })

{
    {
        stuff
    }
}

{{
     stuff
 }
}


### Control flow

if (cond) {
    stuff
} else {
    stuff
}

if (cond) {
    stuff
}
else {
    stuff
}

if (cond)
    {
        stuff
    } else
        {
            stuff
        }

if (cond)
    {
        stuff
    }
else
    {
        stuff
    }

for (seq)
    {
        stuff
    }

for (seq) {
    stuff
}



### Continuation lines

stuff %>%
    stuff %>%
        stuff

{
    stuff %>%
        stuff %>%
            stuff
} %>%
    stuff %>%
        stuff

(
    stuff %>%
        stuff %>%
            stuff
    ) %>%
    stuff %>%
        stuff


[
 stuff %>%
     stuff %>%
         stuff
 ] %>%
     stuff %>%
         stuff

stuff %>%
    stuff %>%
        if (cond) {
            stuff %>%
                stuff %>%
                    stuff
        } else {
            stuff %>%
                stuff %>%
                    for (vec) {
                        stuff
                    } %>%
                        stuff %>%
                            stuff
        } %>%
            stuff %>%
                stuff


ggplot() +
    geom(lhs -
             rhs
         ) +
        geom()

{
    ggplot() +
[...]

PR #175, opened 2015-05-08
ERT testing framework and indentation test cases

lionel-:
Here is the testing framework discussed in #174.

The tests can be run with `./run-tests` (needs to be launched from the test directory) and the indentation test files can be regenerated with `./run-tests --generate-indent-cases`.

The latter will:
* (re)generate test files based on all configurations contained in the variable `ess-test-style-alist`.
* take the `indentation-test-RRR.R` file as a model so any addition or modification of the test cases should be done in this file.

Currently the tests are broken because I have added this `test-misc1` configuration that makes ESS indentation fail with an error :-). It fails on the test cases 11, 13, 17 and 20.

So this PR can be merged and we can correct this bug later. We'll have `./run-tests` to check everything is fixed.

### DIFF
--- a/test/ess-tests.el
+++ b/test/ess-tests.el
@@ -1,0 +1,81 @@
+;; ess-tests.el --- Tests for ESS
+;;
+;;
+;; Filename: ess-tests.el
+;; Created: 07-05-2015 (ESS 15.09)
+;; Keywords: tests, indentation
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; This file is *NOT* part of GNU Emacs.
+;; This file is part of ESS
+;;
+;; This program is free software; you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation; either version 3, or
+;; (at your option) any later version.
+;;
+;; This file is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; A copy of the GNU General Public License is available at
+;; http://www.r-project.org/Licenses/
+;;
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; Commentary:
+;;
+;; To run these tests:
+;;   All tests: M-x ert t
+;;
+;; To apply styles for manual testing:
+;;   M-: (let ((ess-style-alist ess-test-style-alist))
+;;         (ess-set-style 'test-misc1))
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; Code:
+
+(require 'ert)
+
+
+;;; Indentation tests
+
+(defvar ess-test-style-alist
+  ;; New test-RRR style containing additional settings.
+  ;; Compared to RRR, it makes sure personal configuration does not
+  ;; interfere with tests
+  `(,(assq 'RRR ess-style-alist)
+    (misc1
+     (ess-indent-level . 6)
+     (ess-first-continued-statement-offset . 3)
+     (ess-continued-statement-offset . 0)
+     (ess-brace-offset . -6)
+     (ess-arg-function-offset . 0)
+     (ess-arg-function-offset-new-line . 6)
+     (ess-expression-offset . 0)
+     (ess-else-offset . 6)
+     (ess-close-brace-offset . 6)
+     (ess-brace-imaginary-offset . 3)
+     (ess-continued-brace-offset . 3)
+     (ess-close-paren-offset . '(3)))))
+
+(defun ess-test-R-indentation (file style)
+  (let ((ess-style-alist (append ess-test-style-alist ess-style-alist)))
+    (with-temp-buffer
+      (insert-file-contents-literally file t)
+      (let ((expected (buffer-string)))
+        (R-mode)
+        (setq ess-fancy-comments t)
+        (ess-set-style style)
+        (indent-region (point-min) (point-max))
+        (should (equal (buffer-string) expected))))))
+
+(ert-deftest test-ess-R-indentation-RRR ()
+  (ess-test-R-indentation "styles/RRR.R" 'RRR))
+
+(ert-deftest test-ess-R-indentation-misc1 ()
+  (ess-test-R-indentation "styles/misc1.R" 'misc1))
--- a/test/run-tests
+++ b/test/run-tests
@@ -1,0 +1,16 @@
+#!/usr/bin/env emacs --script
+
+;; This script must be run from the test directory
+
+(let ((current-directory (file-name-directory load-file-name)))
+  (setq ess-test-path (expand-file-name "." current-directory))
+  (setq ess-root-path (expand-file-name "../lisp/" current-directory)))
+
+(add-to-list 'load-path ess-root-path)
+(add-to-list 'load-path ess-test-path)
+
+(require 'ess-site)
+(load (expand-file-name "ess-tests.el" ess-test-path) nil t)
+
+;; run tests
+(ert-run-tests-batch-and-exit t)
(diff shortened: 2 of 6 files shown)


test/indentation-test-RRR.R
```diff
@@ -0,0 +1,349 @@
...
+
+## 28
+for (sequence)
+    stuff
+
+
+
+### Continuation lines
+
+## 29
```
vspinu: I think it would wiser to start new numbering in each section. We will surely be adding stuff in the future and it would be nice to keep clean numbering without the need to modify everything else.

test/run-tests
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env emacs --script
+
+;; Adapted from Cider tests
```
vspinu: No need to specify this. It's a fairly standard run-tests file.

test/indentation-test-RRR.R
lionel-: very good point

test/run-tests
```diff
@@ -0,0 +1,47 @@
...
+
+(require 'ess-site)
+(load (expand-file-name "ess-tests.el" ess-test-path) nil t)
+
+
+;;; (Re)generate files with test cases for indentation
+
+;; The RRR file is taken as a model, so modify this file to add or
+;; adjust test cases.
+(when (member "--generate-indent-cases" argv)
```
vspinu: How about a separate ./generate-indent-case file? It doesn't feel right to mix it with ./run-tests, unless you don't want to add other files in the git repo. But why wouldn't you? I think if all style files are in repo it can be used by people to compare styles and we can reference directly patterns in  those files.

test/indentation-test-RRR.R
lionel-: I'll adjust the RRR file, misc1 will be updated when indentation stops failing on this file.

test/ess-tests.el
```diff
@@ -0,0 +1,86 @@
...
+(require 'ert)
+
+
+;;; Indentation tests
+
+(defvar ess-test-style-alist
+  ;; New test-RRR style containing additional settings.
+  ;; Compared to RRR, it makes sure personal configuration does not
+  ;; interfere with tests
+  `(,(append '(test-RRR)
```
vspinu: Would it be posible to make this part a bit more readable. Something like 

```
(test-RRR ,(append ...)
```

test/indentation-test-RRR.R
```diff
@@ -0,0 +1,349 @@
+
```
vspinu: I think these files should have -test- in their names. They are not tests as such.  It would also be nice to use these files as references.  I would simply call them `indentation-RRR.R` or `style-RRR.R`. Maybe even place them in a separate subfolder `samples/` or `styles/` and keep only the base name `RRR.R`.

test/ess-tests.el
lionel-: I tried but I'll try again. I don't program in Lisp often so I'm still struggling with list manipulation.

test/indentation-test-RRR.R
lionel-: alright. Should I put the styles folder in `etc/` then?

test/ess-tests.el
lionel-: I can't get it to work. I keep getting this extra layer of list around the settings, and the only way I found around this was to merge everything inside `(append)`

Here is what I get:
```lisp
Value: ((test-RRR
  ((ess-indent-level . 4)  ;; shouldn't have this extra layer here
   (ess-first-continued-statement-offset . 0)
   (ess-continued-statement-offset . 4)
   (ess-brace-offset . 0)
   (ess-arg-function-offset . 4)
   (ess-arg-function-offset-new-line quote
                                     (4))
   (ess-expression-offset . 4)
   (ess-else-offset . 0)
   (ess-close-brace-offset . 0)
   (ess-continued-brace-offset quote
                               (0))
   (ess-close-paren-offset . 0)
   (ess-fancy-comments . t)))
 (test-misc1
  (ess-indent-level . 6)
  (ess-first-continued-statement-offset . 3)
  (ess-brace-offset . -6)
  (ess-arg-function-offset . 0)
  (ess-arg-function-offset-new-line . 6)
  (ess-expression-offset . 0)
  (ess-else-offset . 6)
  (ess-close-brace-offset . 6)
  (ess-brace-imaginary-offset . 3)
  (ess-continued-brace-offset . 3)
  (ess-close-paren-offset quote
                          (3))
  (ess-fancy-comments)))
```

lionel-:
> It fails on the test cases 11, 13, 17 and 20.

The first three are linked to block arguments. I have a fix ready that I will upload in #169. I'll look into the last one now.

test/indentation-test-RRR.R
vspinu: I have no definite opinion on that. But given that etc is so cluttered and that these files are generated by a a script in test/ folder I think it would be better to leave them in test/ folder (for now at least).

test/ess-tests.el
vspinu: > I can't get it to work. I keep getting this extra layer of list around

:) You need to splice with `,@`. Something like this:

```

(defvar ess-test-style-alist
  ;; New test-RRR style containing additional settings.
  ;; Compared to RRR, it makes sure personal configuration does not
  ;; interfere with tests
  `((test-RRR
     ,@(append (cdr (assq 'RRR ess-style-alist))
	       '((ess-brace-imaginary-offset . 0)
		 (ess-continued-brace-offset . '(0))
		 (ess-close-paren-offset . 0)
		 (ess-fancy-comments . t))))
    (test-misc1
     (ess-indent-level . 6)
     (ess-first-continued-statement-offset . 3)))))
```

You also don't need the `append` there:

```
(defvar ess-test-style-alist
  ;; New test-RRR style containing additional settings.
  ;; Compared to RRR, it makes sure personal configuration does not
  ;; interfere with tests
  `((test-RRR
     ,@(cdr (assq 'RRR ess-style-alist))
     (ess-continued-brace-offset . '(0))
     (ess-close-paren-offset . 0)
     (ess-fancy-comments . t))
    (test-misc1
     (ess-indent-level . 6)
     (ess-first-continued-statement-offset . 3))))
```

BTW, this is precisely why lisp and code as data idea is so great!

lionel-: nice! thanks

lionel-:
Here you go, I amended the commit with all your remarks.

lionel-:
By the way, since you mentioned in the other PR that you prefer to merge a single-commit branch, I amended the commit rather than pushing new ones.

But would you have preferred to see the individual commits first before I squash them together? I just realised it would have been easier for you to check the changes this way.

vspinu:
>  to see the individual commits first before I squash them together?

I just go directly to the global changes. There is usually no much point to look at something that will be subsequently changed, unless the commits are unrelated of course.

test/ess-tests.el
```diff
@@ -0,0 +1,86 @@
...
+(require 'ert)
+
+
+;;; Indentation tests
+
+(defvar ess-test-style-alist
+  ;; New test-RRR style containing additional settings.
+  ;; Compared to RRR, it makes sure personal configuration does not
+  ;; interfere with tests
+  `((test-RRR
```
vspinu: But this will still generate `styles/test-RRR.R`, right?

Except `ess-fancy-comments` all other variables should be part of the style.  Especially RRR style should be the most complete one. Then you can use `RRR` in this alist without further ado. Would you care adding a separate commit for this?

I also think we should set ess-fancy-comment directly in the `ess-test-R-indentation` just before `ess-set-style`. Then styles can  overwrite its value.

test/ess-tests.el
```diff
@@ -0,0 +1,86 @@
...
+     ,@(cdr (assq 'RRR ess-style-alist))
+     (ess-brace-imaginary-offset . 0)
+     (ess-continued-brace-offset . '(0))
+     (ess-close-paren-offset . 0)
+     (ess-fancy-comments . t))
+    (test-misc1
```
vspinu: Is this truly random style or it is something useful?

Once I merge your #168 PR would be nice to create RStudio style that would be as close to RStudio as we can get.

lionel-: It's not completely random, I tried to make it as different from RRR as possible.
Hopefully we'll come up with a better strategy later.

I was thinking the same about the Rstudio style.

lionel-:
Here you go

lionel-:
since everything is auto-generated, we could just as well use all ESS styles as test cases? From a testing point of view it's not necessary because most of the settings are redundant across styles, but it makes sense if we consider the test files as style references.

vspinu:
> but it makes sense if we consider the test files as style references.

Yes. That's the plan. I am just thinking to have inline documentation of what setting does what and then substituted the actuall setting in each file. The simplest aproach is to simply pretty print `ess-style` alist in the beginning of each files, but I am sure we can do better;)

PR #176, opened 2015-05-11
roxygen editing enhancements

lionel-:
This introduces several changes and features to the auto-filling code for roxygen doc blocks.

The most useful feature in that PR is that paragraph filling of fields now produces an indent. This produces more readable code. I use `ess-indent-level` to set the indent size but maybe this needs to be a new setting?

Secondly I modified the `(move-beginning-of-line)` advice to make its behaviour more consistent with the rest of Emacs.

* It now goes to the beginning of line and not to the first non-blank character. I added a `(ess-roxy-first-non-blank)` function to make up for this. Those of you using Evil can bind it to `(evil-first-non-blank)`.

* Calling this command repeatedly would make the cursor switch from the roxy prefix to the zeroth column. I removed this behaviour because I used this function in the code and needed it to be consistent. Also as a user this behaviour felt odd because Ctrl-A did not feel like going against a 'hard wall'. However I can revert this change and find another strategy for the code that relies on this.

I also added the ROxygen fields as paragraph delimiters. This way we can use `(forward-paragraph)` to move from field to field. This also allows to remove the advice around `(mark-paragraph)`.

Finally I modified the `(newline-and-indent)` advice to make it produce an extra indent when the point is in a field. This makes it easier to write long parameter descriptions. On the other hand this also produces an unwanted indentation when we want to write a new field, so I'm not sure if this change is good. Now that field filling will create this indentation, it's also easy to write the field and fill it afterwards.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -35,6 +35,10 @@ or @code{'(t)} to refer to the base indentation level
 as possible R files indented via RStudio. To reproduce the setup of some
 RStudio users, you will also need to set @code{ess-offset-arguments} to
 @code{'(t)}.
+@item @ESS{[R]}: Roxygen fields will now be indented on paragraph
+refilling in order to make the documentation more readable. You can also
+refill commented lines in the @code{@examples} field without squashing
+the surrounding code in the comments.
 @end itemize
 
 Changes and New Features in 15.03-1:
--- a/etc/ess-roxy-tests.R
+++ b/etc/ess-roxy-tests.R
@@ -1,10 +1,10 @@
 ##' \code{loadings(object)} and then design your own plotting method.
 ##' @title Side by side scores and loadings plot
 ##' @usage slplot(object, pcs=c(1,2), scoresLoadings=c(TRUE, TRUE),
-##' sl="def", ll="def", hotelling=0.95, rug=TRUE, sub=NULL,...)
-##' @param object 
-##' @param pcs 
-##' @param tjo 
+##'   sl="def", ll="def", hotelling=0.95, rug=TRUE, sub=NULL,...)
+##' @param object
+##' @param pcs
+##' @param tjo
 ##' @return None, used for side effect.
 ##' @export
 ##' @author Henning Redestig
@@ -18,12 +18,12 @@ setMethod("slplot", "pcaRes",
 ##' .. content for \description{} (no empty lines) ..
 ##'
 ##' .. content for \details{} ..
-##' @title 
-##' @param a 
-##' @param b 
-##' @param d 
-##' @param asd 
-##' @return 
+##' @title
+##' @param a
+##' @param b
+##' @param d
+##' @param asd
+##' @return
 ##' @author Henning Redestig
 trickyInArgsComments <- function(a,#comment
                                  b,#hejhopp trams
@@ -34,11 +34,11 @@ trickyInArgsComments <- function(a,#comment
 ##' .. content for \description{} (no empty lines) ..
 ##'
 ##' .. content for \details{} ..
-##' @title 
-##' @param a 
-##' @param b 
-##' @param cc 
-##' @return 
+##' @title
+##' @param a
+##' @param b
+##' @param cc
+##' @return
 ##' @author Henning Redestig
 withdef <- function(a, b=c("asd","ffd", "asd", "ffd",
                          "asd",
@@ -54,18 +54,18 @@ setClass(Class="inference", representation=representation(model="character"
                                    , two.sided="logical"
                                    , ci.level="numeric"), contains=c("matrix"))
 
-##' .. content for \description{} (no empty lines) 
-##' 
+##' .. content for \description{} (no empty lines)
+##'
 ##' .. content for \details{} ..
 ##' @title asd
-##' @param a 
-##' @param asdsd 
-##' @param sd 
-##' @param ... 
+##' @param a
+##' @param asdsd
+##' @param sd
+##' @param ...
 ##' @return s
 ##' @author Henning Redestig
 tempFixNasFunction <- function(a,asdsd, sd, ...) {
-  asds 
+  asds
 }
 
 setGeneric("updateMu", function(respM, gamma, ...)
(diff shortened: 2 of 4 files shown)


lionel-:
> I removed this behaviour because I used this function in the code and needed it to be consistent.

Also it is not very useful to go to the zeroth column except for removing the roxygen prefix. But we have `(ess-roxy-toggle-roxy-region)` to accomplish this.

lionel-:
oh I forgot: The `@examples` field does not get indented. And you can `(fill-paragraph)` the comments in example code and everything will work as expected.

lionel-:
> On the other hand this also produces an unwanted indentation when we want to write a new field

We could have a command to create a new field? It would make a new line without indentation and write `@` directly.

vspinu:
> We could have a command to create a new field? I

No, no. Too many commands. The stuff is already complex. Maybe we don't advice `new-line-and-indent` but rather tackle this directly in our indentation code.

lionel-:
There is also this one.

I still think that making fields paragraphs is the way to go. It is consistent with other parts of Emacs (e.g. Auctex) and it makes sense to have `(mark-paragraph)` mark fields.

I understand it can be surprising to have `(forward-paragraph)` baby-stepping through one-liner fields. On the other hand, it makes sense that paragraphs should not be defined structurally (in terms of whitespace) but semantically.

vspinu:
> I understand it can be surprising to have (forward-paragraph) baby-stepping through one-liner fields

Yes. This is why I am hesitating. Given that's it's so easy to redefine paragraph regexp I am inclined to live this complex alterantives to the users. I am not a fun of auctex ways and I am redefining my paragraphs there.

There are two problems with the semantic navigation. First, is way stop short of using semantic navigation for the code then (function, blocks etc)? There is no limit of how reffind you can go. Second, semantic navigation leads to inconsistent behavior. I think consistency accros different buffers, modes and smaller context is a very high priority, and Auctex is more of an exception than a rule.

I think it would be the best if we simply define empty lines within comments (not only roxygen comments) as paragraph delimiters. It would lead to predictable and consistent behavior across the whole buffer. This also leaves an additional flexibility that users can control where paragraphs end and start themselves. if you hardwire small paragraphs such refined control over the layout is lost. Users will likely to get irritated and will change their paragraph settings anyways.

One exception might be `@examples` secsion because examples is the predictably large block and are commonly at the end, thus it's a natural marker between code and roxygen documentation.

lionel-:
it's true that it's easy enough to set the paragraph regexp in `init.el`.

I'll revert the commit then. Or should I not restore the original defadvice around `(mark-paragraph)`? I think all `-paragraph` functions should be consistent with each other.

What about the indent on newline issue? I think I should revert this one too because it will be frustrating to write new fields. It's easy to fill the paragraph (all adjascent fields then...) a posteriori to indent the field.

lionel-:
> One exception might be `@examples` secsion

I don't think we need an exception for this one because it is usually made up of several paragraphs already, so paragraph motions will not stop very far from the field marker.

vspinu:
> Or should I not restore the original defadvice around (mark-paragraph)? 

Yes please. Either both or none. I hope this will not affect roxygen functionality in case `mark-paragraph` was used by roxygen itself.

>  I think I should revert this one too because it will be frustrating to write new fields. 

Yes. Most of the roxygen fields are one liners anyways. Longer fields could be refilled. But does the refill work as expected now (indenting subsequent lines)?

lionel-:
@vspinu It's actually not easy to make the filling code work this way... it makes the code complicated.

And I just discovered the variable `use-hard-newlines`. It allows to have semantic paragraphing and still move the cursor on structural paragraphs as you like. Would this be ok for you?

lionel-:
you may be right that semantic paragraphing is not common outside Auctex but I wish it were. If I'm refilling a YAML field, it doesn't make sense to me that the whole thing gets squashed together. Same thing if I'm refilling a markdown footnote or title. To me all modes for structured text editing should define semantic paragraphs.

Of course the decision is yours, just arguing my case ;)

lionel-:
hmmm I thought `use-hard-newlines` only had an effect on motions, but of course it has also an effect on refilling since the latter relies on `(forward-paragraph)`.

lionel-:
oops

vspinu:
I see you managed to add indentation on filing. Great!

Could you please add a news entry on this? I usually document major changes myself before the release but on more complex changes it's always good if the original author documents them.

lionel-:
here you go.
Feel free to ping me when users file issues about indentation, I'll try to maintain the code.

vspinu:
> Feel free to ping me when users file issues about indentation, I'll try to maintain the code.

Thanks. I will certainly do.

vspinu:
There is a problem with move-beginning-of-line advice in roxy comments.

> Calling this command repeatedly would make the cursor switch from the roxy prefix to the zeroth column. 

This is not happening right now. You are just stuck on the end of roxygen comment. So there is no way to get to bol at all right now.

vspinu:
I have quicly fixed that by making it go to real bol on repeated invocation. 

I also think that if we are trying to be smart here, we should be smart with plain comments as well.  Frankly, I don't like the idea of modifying such basic UI at all.  There are plenty of tools out there users can use to customize their navigation commands globally, not only in ESS.

lionel-:
> This is not happening right now. You are just stuck on the end of roxygen comment.

That was on purpose. I think it's not a nice behaviour to go back to bol, I want ctrl-a to feel like I'm going against a hard wall.

lionel-:
To be clearer, when I said,

> Calling this command repeatedly would make the cursor switch from the roxy prefix to the zeroth column.

I was refering to the previous behaviour that I modified. No big deal for me, I can just redefine the advice for my own use I guess. But if I remember correctly, some of my code expects that this command always goes to bol + roxy-prefix, so your revert might introduce some bug.

vspinu:
>  I think it's not a nice behaviour to go back to bol, 

Nice or not, it's not how other emacs works. I missed this part completely as we focused on paragraphs. 

I am personally used to going to bol and deleting lines and selecting regions from there. Not being able to go to bol is frustrating. Just like you said, "it's hard wall" that you cannot smash, and often you need to smash it. I understand that there are many ways to smash it, but users are used to some ways, and this change was badly interfering with those ( my habits at least). As I said, we should not mess up with basic UI at all, but historically we do, so  let's not make it worse than it already is.

> But if I remember correctly, some of my code expects that this command always goes to bol + roxy-prefix, so your revert might introduce some bug.

It's not good to use user level navigation commands in programs. Users might advice them.

lionel-:
Then my preference would be to go to bol all the time. Previously, it went to bol+roxy-prefix, then bol if invoked a second time. So I just made it more consistent and for me it's more useful to go to roxy-prefix. I understand that other workflows might be different and indeed if there is any controversy about this we should make it go to the real bol as it will be more consistent with other parts of Emacs.

vspinu:
> Previously, it went to bol+roxy-prefix, then bol if invoked a second time.

This setup has been there for ages, and how it works now.

>  For me it's more useful to go to roxy-prefix.

For me as well (and I guess for other users also). So I am inclined to keep it as it is.

What don't you like about the current setup besides being inconsistent with other emacs?

We can also make an option for this and make it behave exactly the same for both roxy and standard comments.

lionel-:
> What don't you like about the current setup besides being inconsistent with other emacs?

It just seemed to denature the command, but that may just be me ;)

> We can also make an option for this and make it behave exactly the same for both roxy and standard comments.

hmm, this sounds nice, but this sounds like a feature that should be Emacs-wise rather than mode-wise.

PR #180, opened 2015-05-21
Fix various indentation bugs introduced in code refactor

lionel-:
I think all examples in `R-ESS-bugs.R` work fine now and I spotted a few more bugs related to block indentation. But we still have the issue mentioned in the other PR. I copy it here to pursue the discussion:

I think we have a dilemma here. Before we had
```r
    my.long.Expression <- expression(
        x[a[j]] == exp(theta[1] + theta[2]^2),
        x[b[i]] == sin(theta[3] ~~ theta[4])
        )
```
but now with `ess-offset-arguments-newline` set to `t` we have:
```r
    my.long.Expression <- expression(
                              x[a[j]] == exp(theta[1] + theta[2]^2),
                              x[b[i]] == sin(theta[3] ~~ theta[4])
                          )
```

One way out would be to implement an idea that I had for the future, which is to allow different styles depending on the depth of embeddedness of expressions. For example we could decide to have `'(t)` at level 0 and `t` beyond. This way we'd have:
```r
object <- fun_call(
  arg1,
  arg2
)
```
and
```r
object <- fun_call(
  arg1, other_call(
          arg2,
          arg3
        ),
  arg4)
```
Under this scheme we'd solve the problem above. What do you think?

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -1102,8 +1102,7 @@ Returns nil if line starts inside a string, t if in a comment."
     (let* ((indent-point (point))
            (state (syntax-ppss))
            (containing-sexp (cadr state))
-           (prev-containing-sexp (car (last (butlast (nth 9 state)))))
-           ind)
+           (prev-containing-sexp (car (last (butlast (nth 9 state))))))
       (back-to-indentation)
       (cond
        ;; return nil (in string) or t (in comment)
@@ -1116,10 +1115,8 @@ Returns nil if line starts inside a string, t if in a comment."
        ((ess-block-opening-p)
         (ess-calculate-indent--block-opening))
        ;; Block: Continuations
-       ((setq ind (ess-calculate-indent--continued-block containing-sexp))
-        (+ ind ess-indent-level))
-       ((setq ind (ess-calculate-indent--continued containing-sexp))
-        (+ ind (ess-offset 'continued)))
+       ((ess-calculate-indent--continued-block containing-sexp))
+       ((ess-calculate-indent--continued containing-sexp))
        ((null containing-sexp)
         0)
        ;; Block: Closing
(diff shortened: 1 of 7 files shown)


vspinu:
> hich is to allow different styles depending on the depth of embeddedness of expressions

I think this is an additional complexity that we don't want to have unless really necessary. Current status is just the right level of complexity. 

How about generalizing  `ess-indent-from-outer-parameter` to apply not only to parameter declaration but to assignment as well? That is, make it operate on `lhs` and call it `ess-indent-from-lhs`? 

Then you would have:

```

## ess-offset-arguments-newline = nil  ess-indent-from-lhs = nil
object <- some_function(other_function(
                                       arg1,
                            arg2))

some_function(object <- other_function(
                                       arg1,
                                       arg2))

object <- other_function(
                         arg1,
                         arg2)


## ess-offset-arguments-newline = t  ess-indent-from-lhs = nil
object <- some_function(other_function(
                            arg1,
                            arg2))

some_function(object <- other_function(
                            arg1,
                            arg2))

object <- other_function(
              arg1,
              arg2)


## ess-offset-arguments-newline = t  ess-indent-from-lhs = t
object <- some_function(other_function(
                            arg1,
                            arg2))

some_function(object <- other_function(
                  arg1,
                  arg2))

object <- other_function(
    arg1,
    arg2)


## ess-offset-arguments-newline = (t)  ess-indent-from-lhs = t
object <- some_function(other_function(
    arg1,
    arg2))

some_function(object <- other_function(
    arg1,
    arg2))

object <- other_function(
    arg1,
    arg2)
```

I think this is even more consistent than it's right now as all types of `lhs` are treated identically and we solve our problem without even adding a new var. Win win for everyone, no? 

We can even have `ess-indent-from-lhs` to take the 3rd and 4th value `=only` and `<-only` to accommodate  people's fanciness even further.

lionel-:
> How about generalizing ess-indent-from-outer-parameter

that's exactly what I was thinking on the ride home :)
I like how you framed this in terms of lhs. The whole thing feels right and consistent.

lionel-:
should we generalise it further to horizontal continuations? e.g.:

```r
object <- stuff %>% call() %>%
                        call()

```
instead of how it's hardcoded now:
```r
object <- stuff %>% call() %>%
    call()
```

hmm... we can first implement your design with only lhs for now and then if some people ever want to indent as in the first snippet above it's still time to generalise it then.

vspinu:
> then if some people ever want to indent as in the first snippet above it's still time to generalise it then.

Yes. Let's not complicate things for now; you have just cleaned thing up :)

I am merging this one for now. Thanks.

PR #181, opened 2015-05-22
Generalise ess-indent-from-outer-parameter to ess-indent-from-lhs

lionel-:
This should complete the refactoring.


vspinu:
Looks good. Thanks for the great job! It's probably the biggest refactoring ESS has seen so far:)

lionel-:
and thanks for the great reviewing! you were very helpful.

PR #182, opened 2015-05-26
Fix indent issues

lionel-:
Fixes for the issues reported in #174


vspinu:
Thanks for the quick fix!

PR #188, opened 2015-06-01
Handle trailing operators in evaluated inputs

lionel-:
Hello,

This creates a setting to trim trailing operators from evaluated inputs. This is particularly useful for working with magrittr pipes or ggplot. For example, with this setting enabled you can select and eval the first two lines of the following snippet without having to think about the trailing `+`:
```r
ggplot(...) +
  geom1(...) +
  geom2(...)
```

I personally find this very useful (maybe because I use vim bindings with which it is natural to select whole lines, I know nothing about regular Emacs workflow). I set it to `nil` by default nonetheless because in case this breaks someone's workflow.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -1983,6 +1983,12 @@ of Emacs until the code has been successfully evaluated."
   :group 'ess-proc
   :type  'number)
 
+(defvar ess-eval-without-trailing-operator nil
+  "If non-nil, trailing operators will be trimmed from evaled input.
+This is useful to evaluate partially a magrittr pipeline or a
+ggplot specification.")
+
+
  ; System variables
 
 ;;*;; Variables relating to multiple processes
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1312,6 +1312,15 @@ STRING). In all other cases the behavior is as described in
       string
     (concat string "\n")))
 
+(defun ess--delete-trailing-operator-maybe (input)
+  (if ess-eval-without-trailing-operator
+      (replace-regexp-in-string (concat "\\(%[^ \t]%"
+                                        "\\|[-:+*/><=&|~]\\)"
+                                        "[ \t\n]*\\'")
+                                "" input)
+    input))
+(add-hook 'ess-presend-filter-functions 'ess--delete-trailing-operator-maybe)
+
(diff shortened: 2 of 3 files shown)


lisp/ess-inf.el
```diff
@@ -1312,6 +1312,14 @@ STRING). In all other cases the behavior is as described in
       string
     (concat string "\n")))
 
+(defun ess--delete-trailing-operator-maybe (input)
+  (when ess-eval-without-trailing-operator
```
vspinu: This will fail if ess-eval-without-trailing-operator is nil.  Presend functions must return a string.

lisp/ess-custom.el
```diff
@@ -887,7 +887,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
      (ess-indent-with-fancy-comments . t))
-
```
vspinu: This commit introduces unrelated white space changes.

lisp/ess-inf.el
lionel-: oops. This is fixed

lisp/ess-custom.el
lionel-: should I disable `(delete-trailing-whitespace)` while working on ESS?

vspinu: It's all about synchronization with others. If we can activate that settings for all files in ESS directory that would be good. 

 I was not using `whitespace-cleanup` myself because of the inability to set it per project basis. But now I see that there is a third party [`whitespace-cleanup-mode`](https://github.com/purcell/whitespace-cleanup-mode) and also [`ws-butler`](https://github.com/lewang/ws-butler).  

I will add a setting to `.dir-locals`.

lionel-: this mode looks very nice. Especially the fact that it only cleans whitespace if the buffer was already clean when it was opened.

I reverted the whitespace and added a bullet to the news file.

vspinu: >>> Lionel Henry on Mon, 01 Jun 2015 14:10:18 -0700 wrote:

 LH> this mode looks very nice. 

Yes ... only if it wasn't buggy. I have already opened two issues for it.

PR #189, opened 2015-06-01
Indentation of chained calls

lionel-:
This should do it.

I also adjusted `misc1` to have cascading continuations, since they are not the default anymore.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -16,31 +16,30 @@ customisation is now centered around 3 settings:
 @code{ess-offset-arguments}, @code{ess-offset-arguments-newline} and
 @code{ess-offset-block}. The first two settings control how arguments
 are indented inside function calls and index brackets, while the latter
-controls the indentation of statement blocks. They all take either a
-number @code{N} (indentation relative to innermost function call), a
-list @code{'(N)} containing a number (indentation relative to last line
-indentation as in RStudio) or a @code{nil} value (indentation relative
-to opening delimiter). See the documentation strings of those variables
-for more information. Indentation of delimiters refer to these three
-values depending on the context in order to offer a consistent
-indentation.
-
+controls the indentation of statement blocks. They can be set to
+@code{prev-call} (indentation relative to innermost function call),
+@code{prev-line} (indentation relative to last line indentation as in
+RStudio) or @code{open-delim} (indentation relative to opening
+delimiter). See the documentation strings of those variables for more
+information. Indentation of delimiters refer to these three values
+depending on the context in order to offer a consistent indentation.
 @code{ess-arg-function-offset} has been replaced by
-@code{ess-indent-from-lhs} and has been generalised to
+@code{ess-indent-prev-call-lhs} and has been generalised to
 assignements. This setting works with both statement blocks and
-expressions. It takes only effect when the expression offsets are set to
-a number in order to produce a consistent indentation.
+expressions. It now only takes effect for offsets set to
+@code{prev-call} in order to produce a consistent indentation.
+
 @item @ESS{[R]}: @code{ess-brace-imaginary-offset},
-
 @code{ess-expression-offset} and all delimiter offsets are deprecated.
+The settings for indentation of continued statements have been replaced
+by @code{ess-offset-continuations}. It can be set to either
+@code{cascade} or @code{straight} (the default).
 
-@item @ESS{[R]}: All indentation offsets can now be set to @code{t}
-or @code{'(t)} to refer to the base indentation level
-@code{ess-indent-offset}.
 @item @ESS{[R]}: A new RStudio style is provided to mimic as closely
 as possible R files indented via RStudio. To reproduce the setup of some
 of the RStudio users, you will also need to set
-@code{ess-offset-arguments} to @code{'(t)}.
+@code{ess-offset-arguments} to @code{prev-line}.
+
 @item @ESS{[R]}: Roxygen fields will now be indented on paragraph
 refilling in order to make the documentation more readable. You can also
 refill commented lines in the @code{examples} field without squashing
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -833,6 +833,27 @@ If t:
 
 See `ess-style-alist' for for an overview of ESS indentation.")
 
+(defvar ess-indent-from-first-chained-call t
+  "This switch adjusts the behaviour of
+ess-offset-arguments(-newline) and ess-offset-block when they are
+set to `t'. In this case, arguments are indented starting from
+the function call. When ess-indent-from-first-chained-call is `t'
+as well, chained calls will be treated as if they were one call
+and indentation will start from the first one.
+
+For example, with ess-offset-arguments-newline set to `t' and
+ess-indent-from-first-chained-call set to `nil', we have:
+
+  some_function(other_function(
+                    argument
+                ))
+
+And when the latter is set to `t' instead:
+
+  some_function(other_function(
+      argument
+  ))")
+
 (defvar ess-offset-continued 2
   "Extra indent for lines not starting new statements.
 You can refer to `ess-indent-offset' by setting this parameter to
@@ -875,6 +896,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . t)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . t))
 
     (C++
@@ -886,6 +908,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . t)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . t))
     
     ;; CLB added rmh 2Nov97 at request of Terry Therneau
@@ -898,6 +921,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . 4)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . t))
 
     (GNU
@@ -909,6 +933,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . t)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . t))
     
     (K&R
@@ -920,6 +945,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . t)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . t))
     
     ;; R added ajr 17Feb04 to match "common R" use
@@ -932,6 +958,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . t)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . t))
     
     (RStudio
@@ -943,6 +970,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
      (ess-offset-continued-first . t)
      (ess-indent-function-declaration . t)
      (ess-indent-from-lhs . t)
+     (ess-indent-from-first-chained-call . t)
      (ess-indent-with-fancy-comments . nil))
     
     (DEFAULT
@@ -952,6 +980,7 @@ See `ess-style-alist' for for an overview of ESS indentation."
       (ess-offset-arguments-newline . ,(default-value 'ess-offset-arguments-newline))
       (ess-indent-function-declaration . ,(default-value 'ess-indent-function-declaration))
       (ess-indent-from-lhs . ,(default-value 'ess-indent-from-lhs))
+      (ess-indent-from-first-chained-call . ,(default-value 'ess-indent-from-first-chained-call))
       (ess-offset-continued . ,(default-value 'ess-offset-continued))
       (ess-offset-continued-first . ,(default-value 'ess-offset-continued-first))
       (ess-indent-with-fancy-comments . ,(default-value 'ess-indent-with-fancy-comments))))
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -652,7 +652,77 @@ See `ess-style-alist' for all available offsets.")
 
 (define-obsolete-variable-alias 'ess-indent-level 'ess-indent-offset "15.09")
 
-(defvar ess-offset-block '(t)
+(defvar ess-offset-arguments 'open-delim
+  "Indent for function arguments or bracket indexing.
+This variables has an effect only when the ( or [ are not
+directly followed by a new line. See
+`ess-offset-arguments-newline' for indentation after closing
+newline.
+
+If 'open-delim, the arguments are indented at the opening
+delimiter following foo:
+
+  object <- some_function(other_function(arg1,
+                                         arg2,
+                                         arg3)
+
+
+If 'prev-call, the arguments are aligned at the beginning of the
+closest function call + N characters:
+
+  object <- some_function(other_function(arg1,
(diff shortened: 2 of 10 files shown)


lionel-:
hold on a minute, I will upload another fix soon.

vspinu:
I keep thinking about your `(t opt1 opt2)` idea. It has one major benefit that we get rid of the impure `indent-from-lhs` and `chained-calls` variables. I wonder if it can work better if we combine it with named options that we discussed before. 

For example with  `ess-offset-arguments` we can have `(prev-line N)`, `(lhs N)`, `(chain N)`, `(lhs-chain N)` and `(open-paren N)`? In most common case when N is `t` the user can simply specify `prev-line`, `lhs`, `chain` etc. An obvious advantage of this is that we can painlessly extend this system in the future if we need so. 

What do you think?

lionel-:
sounds good, but I'm not sure I understand your proposed settings. Here is how I see it:
```r
(ess-offset-arguments  .  (prev-line t))              ;; Rstudio
(ess-offset-arguments  .  (prev-call t :lhs :chain))  ;; with :lhs and :chain optional
(ess-offset-arguments  .  (open-paren t))
```

but in the last case what would t mean ? An even simpler way would be to assume `t` where it makes sense (the first two), and have a `:size` key to optionally specify a different size:
```r
(ess-offset-arguments  .  (prev-line :size 2))      ;; size 2 instead of t
(ess-offset-arguments  .  (prev-call :lhs :chain))  ;; :size t is implied here
(ess-offset-arguments  .  (open-paren))
```

vspinu:
> (ess-offset-arguments  .  (prev-call t :lhs :chain))  ;; with :lhs and :chain optional

I meant to reduce this one to three settings - `lhs`, `prev-call-chain` and `prev-call`.  All other combinations between them and  with `prev-line`/ `open-paren` are meaningless. 

I also suggested that we need not use the list for settings in the common case. That is, `prev-line` is equivalent to `(prev-line t)`. 

> and have a :size key to optionally specify a different size:

This is difficult in terms of customize interface. Also it leaves the same problem as we have right now - optional settings for some values and no settings for others. It's marginally better than the current situation.

> but in the last case what would t mean ? 

This is awkward indeed. 

Some time ago you suggested enforcing `t` on these offsets.  Maybe we just go for it and be done with this ugly customization business. No-one will use it anyways. We leave the flexibility tor ess-offset-continued and  ess-offset-continued-first where it's not problematic. Contrary to what I though before, this doesn't affect our default styles.

lionel-:
>  It's marginally better than the current situation.

as I see it, it's much better because it's hierarchically structured. You cannot set `:lhs` if you don't set the offset to `prev-call` first. It also means it's easier to comprehend, because you first learn about three well separated types of offsets, and then learn about ways of customising some of them. This would be lost with a flat hierarchy.

But your suggestion is good as well.

lionel-:
> reduce this one to three settings - lhs, prev-call-chain and prev-call

but how do you combine call-chain and lhs?

vspinu:
> you first learn about three well separated types of offsets, and then learn about ways of customising some of them. 

How do you see the customization interface for this? I guess customization could be done yeararchicaly, but if you customize it in elisp, nothing stops you from doing `(:prev-line :lhs)`.

> but how do you combine call-chain and lhs?

Forgot about that one. Something like `prev-callchain-lhs`. Essentially it comes to the same thing as `(:prev-call :lhs :chain)` but it's much easier in terms of customization - it's a flat `(choice ...)` dropdown which, IMO, is easier to maintain and comprehend. 

The advantage of a list is that you can include the size, and you need not even prefix it with `:size` just include a number in that list. The disadvantage is that the interface is pretty non-standard.

lionel-:
aaah you're right... I never use the customisation interface so I totally forgot about it :/

lionel-:
> nothing stops you from doing (:prev-line :lhs)

well I can stop anyone from doing that :). Just have to add informative errors in `(ess-extract-offset)`. By the way I was thinking of `(prev-line :lhs)` with the first argument without colon to distinguish between offset type and parameters.

In the end I trust your judgement so we'll go with what you think is best.

lionel-:
> Some time ago you suggested enforcing t on these offsets. Maybe we just go for it and be done with this ugly customization business. No-one will use it anyways.

And if anyone wants a different indent size, they can advise the correct indentation function to redefine `ess-indent-level`.

If we go this way, I think we should also turn the two continued statements settings into a single boolean.

vspinu:
> And if anyone wants a different indent size, they can advise any indentation function to redefine ess-indent-level.

Is the indentation engine that clean now? That you can  advice a function for each indentation type?

> f we go this way, I think we should also turn the two continued statements settings into a single boolean.

That sounds appealing but it would break styles. I also think that making continuation lines indent by a different offset can be very informative.

lionel-:
> Is the indentation engine that clean now? That you can advice a function for each indentation type?

We really only have arguments, arguments-newline and block left to customise so yes. Well I guess you cannot make a different indent size for the argument offsets since they are handled in the same function.

> I also think that making continuation lines indent by a different offset can be very informative.

We could use the same scheme as for other offsets. With your scheme:

```
(ess-offset-continuations  .  cascade)
(ess-offset-continuations  .  straight)
```
and optionally:
```
(ess-offset-continuations  .  (cascade t))
(ess-offset-continuations  .  (straight 5))
```

lionel-:
this discussion is a bit exhausting but we better get it right because it will probably stay that way for years to come :)

vspinu:
> this discussion is a bit exhausting but we better get it right because 

Agreed. I plan for a rewrite of R-mode very soon and I would very much like that we  fix indentation now

> (ess-offset-continuations  .  cascade)
> (ess-offset-continuations  .  straight)
> (ess-offset-continuations  .  (cascade t))
> (ess-offset-continuations  .  (straight 5))

This is nice. I like it.  

>  the way I was thinking of (prev-line :lhs) with the first argument without colon to distinguish between offset type and parameters.

After yet another half an hour of thinking I have a bunch of more reasons for flat lists. It's idiomatic emacs  which people are used to. People are also used to parlists like `(prev-line :lhs t)` and `(prev-line :lhs)` feels slightly unfamiliar. While customizing in elisp, when the whole symbol is explicit like `prev-call-lhs`, it feels simple and well defined because individual parts need not have independent semantic meanings. Those are simply tags. Whereas `(prev-call  :lhs)` has two moving parts,  which can vary independently. Hence the cognitive burden of distinguishing `(prev-call  :lhs)` vs `(prev-line  :lhs)` is much higher.  By adding yet another dimension `:chain` you get exponential increase in potentially valid combinations. With flat names this problem doesn't occur at all (at least with my brain). Finally, with new `ess-offset-continuations` the whole system will consists of exactly the same customization pattern `foo` or `(foo t)`, nice and uniform IMO.

lionel-:
alright, if you don't find my proposition intuitive chances are many people won't. The other advantage is that your way is easier to implement.

Then the last question to sort out: should we drop support for custom indent size? Currently we can customise arguments, arguments-newline, block and continuations. If we remove this possibility, people can still advise the arguments, block and continuations functions to adjust `ess-indent-level`.

I don't think it's very important to drop that support, it would maybe make the system feel more simple and consistent, as well as simplifying the code a bit.

lionel-:
> you get exponential increase in potentially valid combinations

I just want to add that parametrising the offset type is actually a way of handling this potential increase. With your scheme we increase the number of actual settings exponentially.

For example, we talked about the possibility of a setting for indenting from the start of horizontal continuations. With the flat setting structure it would be too expensive to introduce because you then need to create all combinations with lhs and chains.

Not that it matters so much because I don't think we will have to add many settings.

lionel-:
Now that we decided on giving names to offset types, we could go back to the idea of boolean switches since the name can be indicative of what offset type it applies to.

so:
```
(ess-offset-block              .  prev-indent)
(ess-offset-arguments          .  open-paren)
(ess-offset-arguments-newline  .  prev-call)
```
and then:
```
(ess-indent-prev-call-chains  .  t)
(ess-indent-prev-call-lhs     .  t)
```

This would avoid the exponential blowout of combined settings because we effectively parametrise the offset type, but with an easy familiar interface.

vspinu:
> This would avoid the exponential blowout of combined settings because we effectively parametrise the offset type, but with an easy familiar interface.

That's a briliant idea. Cleaner than that is impossible.

But also with optional numeric values, right?

lionel-:
> But also with optional numeric values, right?

For the offset types? yes unless we drop support for custom indent size.

vspinu:
> For the offset types? yes unless we drop support for custom indent size.

Let's keep. It doesn't add programming complexity but allows for all the flexibility one would need. 

In terms of programming, because all offsets are the same we can simply have `ess-normalize-offset` which would always return a canonical pair `(foo N)` where N is a number.

lionel-:
perfect, this should be easy to do, I'll try to do it tonight.

vspinu:
Thanks. No hurry.

lionel-:
@vspinu finally done.

I fixed many issues and rewrote the code for indentation of continued statements to handle more corner cases and avoid unnecessary loops.

I also implemented a new setting `ess-indent-override-ifelse` to align vertically nested `ifelse()` calls:
```r
data %>%
  mutate(
    var = ifelse(condition1, out1,
          ifelse(condition2, out2, out3))
  )
```

Finally, the default for continuations of logical operators is now to conform to `ess-offset-continuations`. Should we add a setting `ess-indent-override-logical-ops` to reproduce the old behaviour? If we do, I think it should be consistent and apply in all situations, not only inside `if()` calls. Before we had:
```r
if (cond1 ||
    cond2 && (cond3 ||
                  cond4)) {
    cond5 ||
        cond6
}
```
But I think it should be
```r
if (cond1 ||
    cond2 && (cond3 ||
              cond4)) {
    cond5 ||
    cond6
}
```
or maybe only inside parentheses.

However, now that continuations are set to `straight` by default, I'm not sure if this old tweak is still needed.

vspinu:
> @vspinu finally done.

Great!

> I also implemented a new setting ess-indent-override-ifelse to align vertically nested ifelse() calls:

This might apply to other functions. Let's have a list of  names of functions that should be aligned to the previous call. Or what's the exact generic pattern here? What exactly `ess-indent-override-ifelse` overrides? I don't like this name, it can mean anything.

> Should we add a setting ess-indent-override-logical-ops to reproduce the old behaviour? If we do, I think it should be consistent and apply in all situations, not only inside if() calls.

Yes. I never liked special treatment of if. 

I am not sure about `ess-indent-override-logical-ops`. Can we makes continuation symbols customizable and don't include logicals by default?

> or maybe only inside parentheses.

I am not sure we need that special case. The only standard case beside `(` is lhs pattern. Right now we have (is this a bug?):

```

out <-
    test1 | test2 |
    test3 | test4

funcall(arg1 =
            test1 | test2 |
            test3 | test4
        )

```

The above is fine with me. So If we have a contextless configuration for logicals we will probably be fine.

lionel-:
> is this a bug?

hmm good question. Right now we have:
```r
## 15
object <-
    stuff1 +
    stuff2 ~
        stuff3 +
        stuff4 :=
            stuff5 +
            stuff6 =
                stuff7 +
                stuff8
```
but maybe each time a new continuation is started with a definition-op, we should apply a new indent on the second lines? I think it's probably more aesthetic the way it is now though. Hrm, just found a bug:
```r
object <- stuff1 +
    stuff2 + stuff3 +
    stuff4 ~ stuff5 +
    stuff6 + stuff7 +
    stuff8 := stuff9 +
                stuff10 + stuff11 +
                stuff12 = stuff13 +
                stuff14 + stuff15 +
                stuff16
```
Should be:
```r
object <- stuff1 +
    stuff2 + stuff3 +
    stuff4 ~ stuff5 +
        stuff6 + stuff7 +
        stuff8 := stuff9 +
            stuff10 + stuff11 +
            stuff12 = stuff13 +
                stuff14 + stuff15 +
                stuff16
```

>  Can we makes continuation symbols customizable and don't include logicals by default?

I'm not sure I understand what you mean. But in any case I think the default should be that logicals behave just as any other operator.

> The above is fine with me.

Good! so it's ok if they do get indented inside `if()` calls? 

> This might apply to other functions.

yes, what about: `ess-indent-align-nested-calls  .  (ifelse myfun)`

lionel-:
> what about: ess-indent-align-nested-calls . (ifelse myfun)

should `nil` be the default by the way?

vspinu:
>>   Can we makes continuation symbols customizable and don't include logicals by default?

> I'm not sure I understand what you mean. 

Right now the continuation oepprators are hardcoded like `(looking-back "[-:+*/><=&|~]" ...)`. I thought of making this a custom variable and not to include logicals by default. 

> so it's ok if they do get indented inside if() calls?

No, no. That's a thorny issue. It used to be in the past for short period when all continuations were treated identically and independently of the context. People started complaininng very quickly. No one seems  to like offsetting of continuation lines in if statements.

So if we decide to treat logicals specially and contextlessly then  the default should be no offset.

> should nil be the default by the way?

I think it should be `(ifelse)` by default. I think nested`ifelse` pattern is pretty common and it's conceptually similar to `switch` and `if .. else if ... else if` statements. If it's not the default most users will not use this feature.

vspinu:
> I think it's probably more aesthetic the way it is now though. 

I agree.

>  definition-op,

That's a good name. We can have a var that would hold a regexp or a list of definition ops and then our lhs options would apply to all definition ops. Would that make sense?

lionel-:
@vspinu 

your idea of leaving logical operators out of the regex pattern mostly works, but there are some issues:
```r
out <-
    test1 | test2 |
test3 | test4
```

Not sure what to do. It's probably better to have an exception for continued logical statements inside parentheses.

vspinu:
> It's probably better to have an exception for continued logical statements inside parentheses.

How about ignoring `ess-offset-continued-first` for inside parenthesis then?  No exception for logicals.

lionel-:
nope that would cause turmoil :)

think of all formulaes, numeric and matrix operations etc

lionel-:
@vspinu Do you know of an alternative to `(or)` that evaluates all its arguments (so that side effects can take place), even after encountering one that is non-nil? Something like R's `any()`. I thought it should be a pretty standard function but I can't find any.

vspinu:
> think of all formulaes, numeric and matrix operations etc

I think I would like it better actually. This is what we would get: 

```

arg1 + arg2 + (arg3 +
               arg4 +
               arg5)    
       
## instead of           
arg1 + arg2 + (arg3 +
                   arg4 +
                   arg5)


arg1 + arg2 +
    (arg3 +
     arg4 +
     arg5)

## instead of

arg1 + arg2 +
    (arg3 +
         arg4 +
         arg5)


funcall(arg1 +
        arg2)

## instead of

funcall(arg1 +
            arg2)

```

I don't think this is bad at all, and we don't need exceptions for logicals.


> Do you know of an alternative to (or)

I don't think there is anything like that. The best I can think of is 

```
(cl-every 'identity
	  (list 100
		nil
		(message "here")))
```

You can add a function if it's really useful:

```
(defun ess-or (&rest args)
  (every 'identity args))
```

vspinu:
Oups, that's the implementation for `and` not `or`. It should have been:

```
(defun ess-and (&rest args)
  (every 'identity args))

(defun ess-or (&rest args)
  (some 'identity args))
```

lionel-:
cool, thanks

what should be the setting name for the override? `ess-indent-continuations-at-open-delim` ?

vspinu:
> ess-indent-continuations-at-open-delim ?

Sounds good. But the delim separated by new line should probably not affect the continuation.  The following should still indent as usual:

```
{
    arg1 +
        arg2
}
```

lionel-:
@vspinu it should be good now.

lionel-:
um, there is a merge conflict, can you revert Martin's commit first?

lionel-:
It's not clear what to do with:
```r
fun_call(arg1 +
         arg2, arg3 +
                  arg4)
```

vspinu:
> now fixed, and I've rebased on master

Great! I bet you are exhausted.

> It's not clear what to do with:

This sucks indeed. As does:

```
fun_call(arg1 +
         arg2,
         arg3 +
             arg4)
```

The previous version is not much better but at least it is consistent.

```
fun_call(arg1 +
             arg2, arg3 +
                       arg4)
```

There is one good thing about previous indentation. Notice that in both following cases `arg4` is indented at the same level:

```

fun_call(arg1 +
             arg2,
         arg3 +
             arg4)

fun_call(arg1 +
             arg2,
         par = arg3 +
             arg4)
```

I think `if`, `ifelse` and `(...)` (what else?) are pretty special because only the first (or the only) argument should ignore continuations. Shall we instead of `ess-indent-align-braced-continuations` have `ess-indent-ignore-continuations-in-calls` list of functions? That would imply only the minor modification of the current code.

lionel-:
> I bet you are exhausted.

yeah, this is harder than I expected. hopefully the code will be robust enough that new issues won't pop up all the time

> That would imply only the minor modification of the current code.

ok, I'll adjust this along the ifelse blocks issue, probably in two weeks.

Issue #192, opened 2015-06-25
No indentation for dplyr chains beginning on the first line of a script


karawoo:
I was testing out some of the recent changes to indentation in ESS with a blank R script, and it appears that dplyr chains beginning on the first line of the script are not indented on the second line, but are indented in subsequent lines:

```r
iris %>%
filter(Species == "setosa") %>%
    select(Sepal.Width) %>%
    head
```

If I start the chain on the second line of the script or later, it behaves as I'd expect:

```r
iris %>%
    filter(Species == "setosa") %>%
    select(Sepal.Width) %>%
    head
```

ggplot calls with `+` seem to be indented normally regardless of where they start.

This isn't likely to affect many people's code since most people won't (and shouldn't) write their R scripts this way. But it did confuse me for a while and strikes me as odd behavior.

lionel-:
thanks I'll look into it.

PR #195, opened 2015-07-02
Fix #190 #192

### LINKED ISSUE #190: Distinguish different types of blocks for indentation?
Currently we treat all blocks similarly. Particularly with `ess-offset-block` set to `(t)` we get:

```
r <- funcall(if (cond1)
    stuff1
else if (cond2)
    stuff2
else
    stuff3)
```

How big of a problem is this? Do we really need to differentiate different blocks? And if so, how?

lionel-:
Here are the updates for the last discussions including the one in #189.

About the overrides for blocks, I implemented it as a list because it was easy enough, and on second thought I think it can be useful. This way we allow both if-else blocks and function declaration blocks to be aligned vertically. It makes sense that some users may want indentation at prev-line of blocks in general (for the `test_that` DSL for example) but still want their lambda functions to be indented in a compact way. In particular, some users like to have code in function arguments (cf Martin's example with an if-else block as default function argument), and it looks better to have vertical alignment in this case.

But if you object I can change it to a boolean.


vspinu:
Everything looks ok except my minor comments.

> This way we allow both if-else blocks and function declaration blocks to be aligned vertically. 

List is ok with me. Could you please give an example of `fun-decl` settings? I am not sure I understand what you mean.

lionel-:
> Could you please give an example of fun-decl settings?

For example with `arg1` here:
```r
{
   fun <- function(arg1 = function() {
                             body
                          },
                   arg2 = if (cond) {
                              stuff1
                          } else {
                              stuff2
                          })
}
```

vspinu:
So `fun-decl` refers to the function declaration inside a function declaration only?

How about:

```
fun <- fun_call(arg1 = function() {
    body
})
```
?

vspinu:
More concretely, what is the relationship to `ess-indent-align-declaration-args`. Do we still need it if we have `fun-decl` option?

lionel-:
> So fun-decl refers to the function declaration inside a function declaration only?

No, to any function call, so your example becomes

```r
fun <- fun_call(arg1 = function() {
                           body
                       })
```

lionel-:
the `misc1.R` contains some examples of that

vspinu:
Ok. I am clarified now. We basically converged on having `ess-indent-align-xxx` to fine tune other offsets in a specific vertical pattern.  That's not bad, I think.

lionel-:
now that we have a bunch of `align-` setting, I wonder if we should rename them to `ess-align-` for the sake of shorter variable names?

vspinu:
> now that we have a bunch of align- setting, I wonder if we should rename them to ess-align- for the sake of shorter variable names?

Yes. I don't mind. It's also better for the documentation We can document them inside a separate group in `ess-style-alist`.

lionel-:
Here you go Vitalie

vspinu:
Great! Thanks.

With this we are pretty much done, right? Or, is there any other urgent indentation issues that should be fixed by the next `15.09` release?

lionel-:
we're good for now! :)

PR #202, opened 2015-07-26
Fix issue with indentation of backquoted :=

lionel-:
here you go @vspinu


vspinu:
Thanks for the quick fix!

PR #209, opened 2015-08-09
Slight improvements to *R dired* buffer

vapniks:
1) new sortable column displaying object sizes (in bytes)
2) new command/keybinding `ess-rdired-reverse` to reverse sort order
3) keybinding for `hide-lines` command if it is available (for filtering lines)


vspinu:
Thanks. Looks good.

PR #216, opened 2015-08-13
Fix indentation at prev-call

lionel-:
Here is a fix for @mmaechler's recent issue as well as some fixes for indentation at `prev-call`.

Regarding the latter, we now have 
```r
object <- lapply(object, function(x)
                         {
                             body
                         },
                 argument)

test_that("it works", {
                          test()
                      })
```

Altough in the second case I'm not sure whether we should consider that the previous-call is `test_that()` instead of the opening naked `{`.  cc @vspinu

### DIFF
--- a/etc/R-ESS-bugs.R
+++ b/etc/R-ESS-bugs.R
@@ -1,5 +1,4 @@
-#### File showing off  things that go wrong or *went* wrong in the past
-#### -- with R-mode (mostly coded in ../lisp/ess-mode.el )
+#### File showing off  things that go wrong or *went* wrong in the past #### -- with R-mode (mostly coded in ../lisp/ess-mode.el )
 
 ### NOTE: this file is indented with RRR style !!!!!
 ### but do not change indentations anymore of anything in here:
@@ -789,6 +788,13 @@ read.csv('file.csv') %>%
                dim = d[1:2])
 }
 
+
+### MM: not ok yet
+## {from real code in R sources}
+{
+    obj <- obj && (condition1 || class2 %in% .BasicClasses ||
+                                             condition3)
+}## not ok yet
 
 
 
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -1399,13 +1399,26 @@ Returns nil if line starts inside a string, t if in a comment."
                                     (looking-at "function\\b")))))
                     'decl
                   t)))
-    (if (not (eq block-type 'own))
-        (ess-calculate-indent--args offset (ess-offset-type 'block)
-                                    containing-sexp indent-point block)
-      ;; Block is not part of an arguments list
-      (ess-climb-block)
-      (+ (current-indentation)
-         offset))))
+    (cond
+     ;; Alignment at 'prev-call
+     ((and (eq (ess-offset-type 'block) 'prev-call)
+           (memq block-type '(opening body))
+           (or (eq block 'decl)
+               (looking-at "{")))
+      ;; In case of opening delimiter on its own line, jump to
+      ;; function decl
+      (when (and (eq block-type 'opening)
+                 (eq block 'decl))
+        (goto-char indent-point)
+        (ess-backward-sexp 2))
+      (+ (current-column) offset))
+     ;; Generic case
+     ((memq block-type '(opening body))
+      (ess-calculate-indent--args offset (ess-offset-type 'block)
+                                  containing-sexp indent-point block))
+     ;; Block is not part of an arguments list
+     ((eq block-type 'own)(ess-climb-block)
+      (+ (current-indentation) offset)))))
 
 (defun ess-calculate-indent--comma ()
   (let ((indent (save-excursion
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -1210,13 +1210,13 @@ be advised"
     (cond ((looking-at "(")
            (when (and (ess-backward-sexp)
                       (looking-at "if\\b"))
-             ;; Check for `if else'
+             ;; Check for `else if'
              (prog1 t
                (ess-save-excursion-when-nil
                  (let ((orig-line (line-number-at-pos)))
                    (and (ess-backward-sexp)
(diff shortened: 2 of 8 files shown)


vspinu:
:+1:

vspinu:
> Altough in the second case I'm not sure whether we should consider that the previous-call is test_that() instead of the opening naked {.

Hm. I think I have merged this too fast.I thought we have agreed that test_that type of calls should be shifted with open `{`. Why have you decided to "fix" this more concretely?

lionel-:
I don't recall discussing that. I was pondering the introduction of a stretched style (see the issue I opened today) and I was not sure what we should do there.

Though now I see that (a) the RRR style also uses `prev-call` (I had forgotten about that) and (b) indenting from `{` is ugly in that situation I think. Since RRR also uses `prev-call`, that means that even the lambdas (first case above) should be indented from the embedding call I guess. So feel free to revert the commit e3b3a9e.

But it'd be good to be able to allow "hanging" lambdas one way or another since a few users expressed a preference for that style (Dirk Eddelbuettel and Kevin Ushey).

vspinu:
> I don't recall discussing that.

We didn't discuss this issue specifically but the whole refactoring started with people complaining of too deep indentation of the `function() {}` withing other calls. 

> But it'd be good to be able to allow "hanging" lambdas one way or another since a few users expressed a preference for that style (Dirk Eddelbuettel and Kevin Ushey).

Hm. I was under the impression that we allow for that  style with appropriate config.

lionel-:
> Hm. I was under the impression that we allow for that style with appropriate config.

I don't think so, but we could tweak indentation at `prev-line` with something like `ess-indent-prev-call-from-decl`?

Or maybe the hanging style (better name than "stretched") should be the job of `(ess-offset-block . open-delim)`?

lionel-:
> Or maybe the hanging style (better name than "stretched") should be the job of (ess-offset-block . open-delim)?

well judging from `misc1.R` it actually is! Sorry for the confusion.

So now I can ask the question again. With block indentation at `open-delim`, should naked argument blocks such as in `test_that()` be indented from the opening-delim or from the call? I feel like the former is ugly and would not be used by anybody.

By the way, this discussion makes me think we could rename `from-delim` to `hanging`? That would make sense for indentation of arguments at opening paren as well as for indentation of blocks from function declarations.

Sorry to drag the discussions. You must be as tired as I am of indent settings, but I'd like to tackle the last corners before the release.

vspinu:
> So now I can ask the question again. With block indentation at open-delim, should naked argument blocks such as in test_that() be indented from the opening-delim or from the call? I feel like the former is ugly and would not be used by anybody.

I agree the former is ugly. Let's not spend too much time on it.

> By the way, this discussion makes me think we could rename from-delim to hanging? 

I am lost here. Do you mean `open-delim` setting in `ess-offset-arguments`? I fail to see how "hangling" is more suggestive.

> but I'd like to tackle the last corners before the release.

Thanks for this. It's  definitely useful, but I am starting to forget things a bit.  If you consider some change being uniformly better, just go for it.

> should be the job of (ess-offset-block . open-delim)?

I vaguely understand what you mean here. But yes. it seems right to use `open-delim` for this situation. Wasn't that the original intent for `ess-offset-block`?

BTW, the doc of `ess-offset-block` is super outdated. It doesn't mention the new symbol scheme. Could you sweep through the docs and see if any other historical trails were left?

lionel-:
> I fail to see how "hanging" is more suggestive.

hmm it seems to me that this kind of terminology is used in Latex, e.g. hanging list items.

I like it because it works both for arguments aligned at open paren, and blocks aligned at function declaration. I would also add a "hanging" theme with all offsets set to `open-delim` (or `hanging` if we proceed to rename it).

lionel-:
> Wasn't that the original intent for ess-offset-block?

I also forget things ;)

vspinu:
> hmm it seems to me that this kind of terminology is used in Latex, e.g. hanging list items.

I don't think this is mainstream. It doesn't ring a bell with me as a LaTeX user either :(  Let's stick with the self explanatory `open-delim`.

> works both for arguments aligned at open paren, and blocks aligned at function declaration.

I see the problem with `(setq ess-offset-block 'open-delim)` which indents to the previous open paranthesis as in 

``` 

object <- lapply(object, function(x) {
                     body
                 }, argument)
```

You seem to suggest that the right semantics is to align at `function(x)` for `open-delim`:

```
object <- lapply(object, function(x) {
                             body
                         }, argument)
```

at `object` for `prev-call` 

```
object <- lapply(object, function(x) {
                     body
                 }, argument) 
````

and all the way to bol for `prev-line`

```
object <- lapply(object, function(x) {
    body
}, argument) 
```

I think the above makes sense because it would be in line with the same symbolic names for `ess-offset-arguments`.  On the other hand `open-delim` would lead to the ugly:

```
test_that("it works", {
                          test()
                      })
```

This is why you want to introduce yet another hangling option?  Is my understanding correct?

I think we can simply fix this by being explicit in the documentation of `ess-offset-block` that the alignment is with respect to the block as a whole and not with respect to the block's inner expressions. That is, `open-delim` refers to the "stuff outside" of the block and not to the delimiter of the block itself.

**Comment by vspinu** [comment]
BTW, `(setq ess-offset-block 'prev-call)` doesn't work as expected:

```
object <- bbb(lapply(object, function(x) {
    body
}, argument)) 
```

**Comment by lionel-** [comment]
> (setq ess-offset-block 'prev-call) doesn't work as expected:

Hmm it does if you add
```lisp
(setq ess-indent-prev-call-lhs nil)
(setq ess-indent-prev-call-chains nil)
```

> This is why you want to introduce yet another hangling option?

No no I was confused, now I don't want to add anything else, just clarify the naked `{` behaviour (and I think you're right, let's indent from surrounding context in this case) and find a more suggestive name for `open-delim` if possible.

For instance if we indent a naked block from the surrounding call it's not from the delimiter at all but from the previous call... Would renaming `open-delim` to `aligned` work? Other idea: `rightmost`.

PR #218, opened 2015-08-13
Revert "Fix indentation at prev-call"

lionel-:
This reverts commit e3b3a9e7ff0844b2afcf29cebe182c88c8fcc9fa.

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -1403,26 +1403,13 @@ Returns nil if line starts inside a string, t if in a comment."
                                     (looking-at "function\\b")))))
                     'decl
                   t)))
-    (cond
-     ;; Alignment at 'prev-call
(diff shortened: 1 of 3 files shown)


lionel-:
@vspinu 

Do you know the purpose of `(ess-line-end-position)` ?
See https://github.com/emacs-ess/ESS/blob/master/lisp/ess-mode.el#L44

It wraps an interactive function and doesn't seem to be doing anything different from `(line-end-position)`.  It's used in the utilities for finding a function. Do we want to rely on user advices here?

vspinu:
Such functions come from the time when ESS was supporting Xemacs, or when
the corresponding function was available only in the newest emacs. In any
case, XEmacs is no longer supported and the backward compatibility should
be done differently with `fboundp` check. All such duplicated functions
should be cleaned up some day.

PR #220, opened 2015-08-19
Filling call arguments

lionel-:
The first commit fixes some bugs where refilling would overlap over the code following a roxygen block.

The second implements a new setting `ess-fill-calls` to allow refilling of call arguments. When it is non-nil, calling `(fill-paragraph)` when point is inside a call will indent and format all arguments according to `fill-column`. When called with a prefix, it formats them according to one argument per line.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -44,6 +44,13 @@ of the RStudio users, you will also need to set
 refilling in order to make the documentation more readable. You can also
 refill commented lines in the @code{examples} field without squashing
 the surrounding code in the comments.
+
+@item @ESS{[R]}: A new setting @code{ess-fill-calls} allows function
+calls to be formatted with @code{(fill-paragraph)}. The function's
+arguments will be arranged so as to not exceed @code{fill-column}.
+When @code{(fill-paragraph)} is called with a prefix, the call is
+formatted with one argument per line.
+
 @end itemize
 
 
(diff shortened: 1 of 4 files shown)


vspinu:
Is this related and fixes this issue https://stat.ethz.ch/pipermail/ess-help/2015-August/010641.html ?

I couldn't reproduce it.

lionel-:
Hmm it looks related but I can't reproduce it either, with master or this branch. Though this branch does fix an infinite loop occurring when I `(fill-paragraph)` the generated roxy block.

I didn't know ESS could generate roxygen blocks btw, that's nice!

I found a but with call filling, I'll upload a fix.

lionel-:
> I found a but with call filling, I'll upload a fix.

Fixed.

lionel-:
When I have time I may do the same for continuations. So that long lines like this can be refilled easily:

```
fit_outreach <- lm(registered ~ outreach_action + log_income + urban + retired_pc + single_pc + region, data_eu)
```

And as for calls, without prefix it would arrange the continuations according to `fill-column`, while with prefix it would arrange them as one expression per line.

vspinu:
> When I have time I may do the same for continuations

That would be great. Filling of code is a big weakness with ESS.

vspinu:
This change is awesome. Thanks.

PR #222, opened 2015-08-21
Indentation updates

lionel-:
This fixes a few bugs and refactors indentation of blocks, which I splitted in several functions to make things more manageable. Indentation of blocks at `open-delim` and `prev-call` should be a bit more consistent now (though sometimes uglier, but there are switches like `ess-align-blocks` to prevent this).


lionel-:
@vspinu fwiw, I just found another mode with semantic paragraphing: magit message editing ;)

vspinu:
Cool. I see that you keep discovering weird edge cases :)

PR #227, opened 2015-08-26
Add new indentation setting: ess-indent-block-lhs

lionel-:
Hello,
I'm opening a pull request for this one because it adds a new setting, `ess-indent-block-lhs`.

This controls whether to indent assigned function blocks from the lhs. When set to t (default ESS):

```r
  object <-
      function(argument)
  {
      body
  }
```

When set to nil (default RStudio):
```r
  object <-
      function(argument)
      {
          body
      }
```

 cc @vspinu @mmaechler

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -427,7 +427,7 @@ indentation style. At present, predefined style are `BSD', `GNU', `K&R', `C++',
        (cons "/" ""))
   ;; timer
   (add-hook 'ess-idle-timer-functions 'ess-synchronize-dirs nil 'local)
-  ;;; extras
+;;; extras
   (ess-load-extras)
   ;; SJE Tue 28 Dec 2004: do not attempt to load object name db.
   ;; (ess-load-object-name-db-file)
@@ -1542,7 +1542,13 @@ Returns nil if line starts inside a string, t if in a comment."
                      (goto-char containing-sexp)
                      (and (looking-at "{")
                           (ess-climb-block-opening))))
-                   (some 'looking-at (ess-overridden-blocks)))
+                   (some 'looking-at (ess-overridden-blocks))
+                   ;; This ensures that we indent call-prefixed blocks
+                   ;; from their lhs if they have one, even when
+                   ;; `ess-align-blocks' says to align
+                   (not (save-excursion
+                          (and (some 'looking-at ess-block-funs-patterns)
+                               (ess-looking-back-definition-op-p t)))))
           (+ (current-column) offset))))))
 
 (defun ess-calculate-indent--naked-block ()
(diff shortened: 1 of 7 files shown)


vspinu:
Do we really need it? Shall we go with the second example (RStudio) for all styles?

lionel-:
I also prefer the Rstudio one but I saw on the mailing list that Martin strongly prefers the first one.

vspinu:
> Do we really need it? Shall we go with the second example (RStudio) for all styles?

Actually that would depend on what happens in the standard case of 

```
object <-  function(argument)
  {
      body
  }
```

lionel-:
this is still indented from `object` in all cases because I supposed that no one wants to indent like that.

vspinu:
Is this (`function`) case affected only? It started looking like nitpicking now. I am afraid if we want to handle all small R-studio differences like this, we might end up introducing many more variables like this.  

One big advantage of the first case is that you don't have to re-indent the body after moving `function` to a new line.

lionel-:
> Is this (function) case affected only? It started looking like nitpicking now.

Yes though it could be easily extended to other block calls.

> One big advantage of the first case is that you don't have to re-indent the body after moving function to a new line.

The disadvantage is that to some people it might not look good, or even might look inconsistent. I think the second one is better yet couldn't hardwire it because of backward compatibility.

> if we want to handle all small R-studio differences like this, we might end up introducing many more variables like this.

I don't want to follow RStudio 100%, I found some small corner cases where ESS just make more sense. But where RStudio is consistent I'd like ESS to be able to reproduce it.

I agree this isn't a critically important case though. But now it's implemented we might just as well use it ;)

vspinu:
There is an unfrtunate overlap with the `fun-delc` in `ess-align-blocks`.  

I think indentation of the whole block should be consistent with or without `{` on a new line. So the difference in the following is a bad design:

```

object <-
    function(argument) {
        body
    }


object <-
    function(argument)
{
    body
}
```

Let's stick to the second case without introducing new var. There is no question of backward compatibility. The indentation has been so much rewriten that there will be a lot of differences with the old code. Good thing is that all those differences are in the right direction :)

mmaechler:
I agree with @vspinu (4 hours ago).  @lionel- : I am pretty sure there was a misunderstanding somewhere (here or there) when/if  I gave the impression of prefering the current default for the 
```
function(..)
 {  
}
```
case.  I think I did like the behavior if *inside* an outer block, e.g., typically for `switch(..)` inside a function.  There it helps saving horizontal space if you get

{
     a <- switch(myvar, 
                       "case_1" = 
                          {
                                 ...
                          },
                        "case_2" = 
                           {

lionel-:
Thanks for the comments. It's now hardwired.

I also hardwired an exception for `ess-align-blocks`. Even when function declarations are overridden to be vertically aligned, they still will be indented from the LHS if there is one:

```r
object <- function()
{
    body
}
```
instead of:
```r
object <- function()
          {
              body
          }
```
Since I don't think the second one would appeal to anyone.

vspinu:
> aargh... github markdown does not quite render this as I meant it ... but you get the idea)

:) Yeh ... usual pain. It's because of the tab length. I usually do `M-x untabify` to get rid of tabs and then paste to github.

PR #228, opened 2015-08-27
New hanging/stretched style

lionel-:
Here is the new indentation style we've talked about.

Compared to RRR, it doesn't try to save horizontal space as much. For example prefixed blocks such as lambda functions are indented from the opening parenthesis:
```r
list <- lapply(list, function(x) {
                  body
              },
              argument)
```

Bare blocks are indented from the nearest function call:
```r
some_call(other_call({
              stuff
          },
          {
              stuff
          }))
```

These can be even more shifted to the right by adding `fun-decl` and `bare-blocks` to `ess-align-blocks`. In that case you get:
```r
list <- lapply(list, function(x) {
                        body
                     },
               argument)

some_call(other_call({
                         stuff
                     },
                     {
                         stuff
                     }))
```

We still need to figure out the name. I quite like "hanging". This term is used in the Latex world for bullet lists. The idea is that the body of the block seems to "hang" from the block opening (in the example above, `function(`). An alternative is "stretched" but, thinking about it more, I don't like it because I think it evokes a change of shape rather than position. Other ideas: "rightmost", "easterly", ...

Also, maybe the style should include either `bare-blocks` or `fun-decl` into `ess-align-blocks`?

It'd be nice to have opinions from others on this. Cc @eddelbuettel and @kevinushey since you were interested in a style that doesn't save horizontal space. It must be a C++ thing ;)

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -731,7 +731,7 @@ the closest function call:
                            ))
 
 You can control the details of indentation at `prev-call' with
-`ess-indent-prev-call-lhs' and `ess-indent-prev-call-chains'.
+`ess-indent-from-lhs' and `ess-indent-from-chain-start'.
 
 
 When set to `prev-line', arguments on a new line are indented
@@ -779,7 +779,7 @@ closest function call:
                  }))
 
 You can control the details of indentation at `prev-call' with
-`ess-indent-prev-call-lhs' and `ess-indent-prev-call-chains'.
+`ess-indent-from-lhs' and `ess-indent-from-chain-start'.
 
 
 When set to `prev-line', blocks are indented relative to the
@@ -893,7 +893,7 @@ aligned vertically. With `fun-decl', the body of a function
 declaration will always be aligned with the call to
 `function'.")
 
-(defvar ess-indent-prev-call-lhs nil
+(defvar ess-indent-from-lhs nil
   "When non-nil, indent arguments from the left-hand side of an assignment.
 
 This setting only has an effect when indentation of arguments or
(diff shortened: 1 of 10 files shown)


vspinu:
How about "Greedy" ? I think it's more suggestive of "doesn't try to save space" than "hanging". 

It might be just me who doesn't find "hanging" suggestive. If others like it, let's go for it.

lionel-:
I also considered greedy, but it could be viewed either way: could be that the style wants to eat all the spaces because it's greedy ;)

lionel-:
what about "lax" or "loose"?

vspinu:
How about simply `aligned`?

mmaechler:
I'd also prefer something different than "hanged" ...

But let me re-iterate what I said on the related issue:  The real  RRR style (in the sense of R core, and the R manual) was what ESS did about 10-15 years ago (and still did about 4-5 yrs ago)... in the case it worked well.
I don't want to go back to that version of ESS at all, but let me say that style with that ESS version did try to align more (including probably also in the above case) than the current RRR in current ESS.

lionel-:
`aligned` is the most straightforward name we can find I think.

I changed indentation of bare blocks again because I'd rather have alignment at open delim completely consistent throughout arguments and blocks. From
```r
some_call(argument, {
    stuff
},
{
    stuff
})
```
to
```r
some_call(argument, {
              stuff
          },
          {
              stuff
          })

```
Also I found another interface dilemma. It occurred to me that the two `prev-call`-specific switches for LHS and chained calls could be relevant to `open-delim`. So how about renaming `ess-indent-prev-call-lhs` to `ess-indent-from-lhs` and `ess-indent-prev-call-chains` to `ess-indent-from-first-of-chain`? This way we leave open the possibility of having these switches affect blocks at open-delim in the future, if someone should ask for that.

vspinu:
> Also I found another interface dilemma. It occurred to me that the two prev-call-specific switches for LHS and chained calls could be relevant to open-delim. 

Makes sense, indeed. 

> So how about renaming ess-indent-prev-call-lhs to ess-indent-from-lhs and ess-indent-prev-call-chains to ess-indent-from-first-of-chain? 

I like the proposed names better. They are a bit more intuitive. The old names emphasize that the setting is for `prev-call` only.  If that's not true anymore, they should be definitely renamed.

vspinu:
> ess-indent-from-first-of-chain

`ess-indent-from-chain-start` sounds a bit better to me.

lionel-:
> ess-indent-from-chain-start sounds a bit better to me.

Perfect, thanks.

PR #243, opened 2015-09-20
Add RStudio-unaligned style

lionel-:
This style reproduces RStudio indentation when "Vertically align arguments" is not activated.

For example Hadley's packages use this style so I think it's important to include it.


vspinu:
I don't quite follow. Hadley is using R-studio, so his code should comply with our R-studio style. 

What is the difference between R-studio and R-studio-unaligned more concretely? If the difference is small, maybe we should keep only one of them.

lionel-:
The difference is that `ess-offset-arguments` is set to `prev-line` instead of `open-delim`. This is because Hadley leaves this checkbox unticked:

rstudio

But it is ticked by default so most of the time, when you collaborate with RStudio users you will use the `RStudio` style. And when you work on Hadley's packages, `RStudio-unaligned` will conform to the house rules.

lionel-:
I think we should have both of them. For historical reasons we are stuck with many styles with very small differences that are not really useful, but these two RStudio styles actually match code indentation of hundreds of packages.

vspinu:
Makes sense. I don't quite like the name though. Can we do better? Would it make sense to call this RStudio and the another one RStudio-alligned in order to parallel RRR naming?

lionel-:
We could, but wouldn't this encourage users to use the unaligned version? That would go against ESS policy to favour open-delim alignment of arguments (which is also RStudio's policy for that matter).

We could call it `'hadley ;)

lionel-:
or `'RStudio-Hadley`

vspinu:
> We could, but wouldn't this encourage users to use the unaligned version? That would go against ESS policy to favour open-delim alignment of arguments (which is also RStudio's policy for that matter).

Ok. This is getting very confusing. I ghouthRRR an `unaligned` style as compared to  `RRR-aligned`. 

So the `alignment` in RStudio-unaligned and RRR-aligned would mean very different things. This is bad. I think `RStudio-unaligned` roughly corresponds to settings all `ess-align-` variables to `nil`. 

We should probably rename `RRR-aligned` into more specific `RRR-aligned-blocks`.

> or 'RStudio-Hadley

I think we can do better.

lionel-:
> This is getting very confusing. I ghouthRRR an unaligned style as compared to RRR-aligned.

The thing is that `'open-delim` is in itself an alignment of arguments. For that reason I wondered before if we should have it named `aligned` but open-delim is probably more suggestive of what it does.

> RRR-aligned-blocks

Makes sense.

> I think we can do better.

I'm out of new ideas. `RStudio-prev-line-args` is not great plus it'd be good to avoid lengthy names.

lionel-:
* RRR and RRR-aligned-blocks
* RStudio and Rstudio-unaligned-args ?

vspinu:
This is quite good. But is it only about args? I see in your code that almost all `ess-aligne-` are nil.

lionel-:
Well the difference with the RStudio style is only about args.

vspinu:
How about RRR/RRR2, RStudio/RStudio2?

lionel-:
It's good enough for RStudio but somehow something bothers me for RRR.

Can we rename RRR-aligned to `shifted`?

vspinu:
Why shifted?

vspinu:
I would expect shifted is opposite of aligned. I even thought about RRR-vblocsk vs RStudio-sargs. v stands for vertical and s stands for shifted.

lionel-:
I thought about it as shifted from the normal indentation, but your logic makes more sense indeed. 
Maybe simply RRR-aligned and RStudio-shifted? I prefer not too many particles.

Or we can settle on RStudio/RStudio2 and if we can't find a good name for RRR-aligned, we can just leave it like this. Last ideas: `pushed` or `nudged`.

vspinu:
Long names are good because if the user (or us) would want to make another style with some changed alignment it would be consistent. 

The problem is that whatever way we go even long names are not representative. RRR-aligned-blocks tells you that `ess-align-blocks` is set to t but it doesn't tell you that `ess-indent-from-chain-start` is also changed. 

So I think we probably should simply go for RRR/RRR2, RStudio,RStudio2 and describe them in the documentation. Or maybe RRR/RRR+, RStudio,RStudio-.

lionel-:
I like the last idea!

lionel-:
:+1:

mmaechler:
>>>>> Lionel Henry <notifications@github.com>
>>>>>     on Sun, 20 Sep 2015 15:43:41 -0700 writes:

    >> This is getting very confusing. I ghouthRRR an unaligned
    >> style as compared to RRR-aligned.
    > The thing is that `'open-delim` is in itself an alignment
    > of arguments. For that reason I wondered before if we
    > should have it named `aligned` but open-delim is probably
    > more suggestive of what it does.

    >> RRR-aligned-blocks

    > Makes sense.

    >> I think we can do better.

    > I'm out of new ideas. `RStudio-prev-line-args` is not
    > great plus it'd be good to avoid lengthy names.

Yes, please *DO* avoid lengthy names.
For my taste, I found  'RRR-aligned'  already long...!

vspinu:
We ended up with `RRR/RRR+, RStudio/RStudio-`. Good enough? I am still not sure about those. Those +/- give me strange feelings. An alternative is a simple `RRR/RRR2, RStudio/RStudio2`.

lionel-:
> I am still not sure about those. Those +/- give me strange feelings

oh, I like them a lot. Simple and informative.

vspinu:
> Simple and informative.

Well. I guess they are indeed. Let's stick to them then.

PR #245, opened 2015-09-23
add support for ?topic documentation in julia REPL

dellison:
This change adds support for using '?[topic]' in the Julia REPL to display documentation.

It adds a function ess-julia-input-sender that checks for a ? at the beginning of the input in a Julia REPL. If it's there, it calls ess-display-help-on-object to show the documentation (if available), and otherwise defers to the default inferior-ess-input-sender.

It's more or less a simplified version of ess-R-input-sender (in ess-inf.el), but it doesn't attempt to sanitize the input or check for any other special cases. It appeared to me that sanitizing the input isn't necessary in this case, since ?if and ?when don't cause the process to hang.

### DIFF
--- a/lisp/ess-julia.el
+++ b/lisp/ess-julia.el
@@ -90,6 +90,16 @@ VISIBLY is not currently used."
          (page (ess-completing-read "Lookup:" pages nil t)))
     (browse-url (get-text-property 1 :manual page))))
 
+(defun ess-julia-input-sender (proc string)
+  (save-current-buffer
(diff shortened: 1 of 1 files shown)


dellison:
Thanks!

mmaechler:
well, I should have thanked *you* for the good code contribution.

dellison:
My pleasure! I really enjoy using Julia and ESS, so I'll be sure to stay watchful for other places where I can help things continue to improve. It's exciting for me to be able to contribute in a small way to a great project like ESS :)

PR #250, opened 2015-09-30
Insert argument offset before delimiter on newline

lionel-:
A huge frustration of mine when I write R code is that `(newline-and-indent)` inside a call rarely does what I want. The advice in that PR fixes that.

It adds an offset only when:
* point is in a call
* the char after point is a `)`
* the line ends with a `,` so we know a new argument is incoming

This way, a newline at:
```r
  fun_call(argument,|)
```
produces:
```r
fun_call(argument,
  |)
```
instead of:
```r
fun_call(argument,
|)
```

Do we need to add a setting to turn this advice off, in case this would interfere with other package or user advices?

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -2197,6 +2197,18 @@ otherwise nil."
                             (ess-climb-name)
                             (setq finished t))))))))
 
+(defadvice newline-and-indent (around ess-newline-and-indent)
+  ad-do-it
+  (when (and (eq major-mode 'ess-mode)
+             (string= ess-dialect "R")
+             (looking-at ")")
+             (ess-point-in-call-p)
+             (save-excursion
+               (ess-skip-blanks-backward t)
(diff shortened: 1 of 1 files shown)


lionel-:
hmmmm....

actually why not just make a new indentation rule for closing `)` after a comma....

lionel-:
Here it is: https://github.com/emacs-ess/ESS/commit/15dc4c7241f2db7e1c39298d37c96d770d897cc6

Don't know why I didn't think of that earlier.

vspinu:
Have you removed the advice? We shouldn't interfere with the user's settings and I bet a lot of users already have some sort of advice for their RET. If the "after comma" trick does the job we should stick to that. Let's try to not pile up advices, but remove those instead if we can.

lionel-:
yes it works as it should with a simple indentation rule.

vspinu:
Great!

PR #256, opened 2015-10-13
Update Makefile

dickoa:
There is an error for user relying on Makefile to install ESS on their system. The file ess-r-syntax.el need to be compiled to solve this problem on my setup. Someone can confirm that this change solves this problem and doesn't break anything ?


mmaechler:
thank you; you are entirely right.

PR #259, opened 2015-10-14
names -> fieldnames in Julia 0.4.

denizyuret:
Fixed another Julia v0.4 compatibility issue.


denizyuret:
OK, I made some backward compatibility fixes, the latest version should work on both 0.3 and 0.4.

PR #261, opened 2015-10-23
Display knitr output buffers only on error

Jehops:
The patch to `ess-swv-run-in-R` will only display a new window when
there is an error in the R code and will display a message in the
minubuffer when the weaving/tangling/kniting/purling is complete.
`ess-swv-run-in-R` is currently only used for
weaving/tangling/kniting/purling, so this won't affect other aspects of
ESS.

The patch to `ess-swv-PDF` simply prevents the buffer with the TeX
output from being displayed when there are no errors compiling the TeX.
Instead, a status message is displayed in the minibuffer.

Motivation:

The usual routine to create a PDF file from a knitr Rnw document is to
run `ess-swv-knit` with `M-n r`, then `ess-swv-PDF` with `M-n P`.  It's
not hard to do `C-x 1` after these commands to close the newly displayed
window, but for anyone with multiple windows displaying, it can be
tedious to return to the previous window configuration.

TODO:

For consistency, update ess-swv-PS to mimic the behaviour of ess-swv-PDF.

### DIFF
--- a/lisp/ess-swv.el
+++ b/lisp/ess-swv.el
@@ -122,8 +122,7 @@
               (format ess-swv-processing-command cmd cmd-args)))
         (message "%s()ing %S" cmd rnw-file)
         (ess-execute Sw-cmd 'buffer nil nil)
-        (switch-to-buffer rnw-buf)
-        (ess-show-buffer (buffer-name sbuffer) nil)))))
+        (message "Finished %s()ing %S" cmd rnw-file)))))
 
 (defcustom ess-swv-processing-command ".ess_weave(%s, %s)"
   "Command used by `ess-swv-run-in-R'.
@@ -256,15 +255,16 @@ default using the first entry of `ess-swv-pdflatex-commands' and display it."
                             (file-name-nondirectory latex-filename)
                           latex-filename)))
     (if (not (= 0 pdf-status))
-        (message "** OOPS: error in '%s' (%d)!" pdflatex-cmd pdf-status)
+        (progn
+          (message "** OOPS: error in '%s' (%d)!" pdflatex-cmd pdf-status)
+          (display-buffer tex-buf))
       ;; else: pdflatex probably ok
       ;; (set-process-sentinel proc 'shell-command-sentinel)
       (if (and ess-microsoft-p (w32-shell-dos-semantics))
           (shell-command cmdstr-win)
         (message (mapconcat 'identity cmd " "))
         (apply 'start-process  (car cmd) nil cmd)))
-    (display-buffer tex-buf)
-  ))
+    (message "%s finished with status %s" pdflatex-cmd pdf-status)))
 
 
 (defun ess-insert-Sexpr ()
--- a/lisp/ess-swv.el
+++ b/lisp/ess-swv.el
@@ -88,8 +88,8 @@
 (require 'ess-r-d); for Rnw-mode
 (require 'easymenu)
 
-;; currently use exactly for "Sweave" and "Stangle"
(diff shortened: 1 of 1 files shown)


mmaechler:
The proposed changes look good to me ((but I did not even try them)).
You mention / changed 'knitting'  only, but not Sweaving. 

 Shouldn't there be analogous changes (i.e. not showing the "log" buffer by default unless in case of errors)?  Or are they not needed there?  {{sorry for being lazy and not checking myself}}

vspinu:
> The patch to ess-swv-PDF simply prevents the buffer with the TeX
output from being displayed when there are no errors compiling the TeX.
Instead, a status message is displayed in the minibuffer.

> The usual routine to create a PDF file from a knitr Rnw document is to
run ess-swv-knit with M-n r, then ess-swv-PDF with M-n P.

I am not sure about this. I often want to see tex file and then compile tex directly with auctex.

I think a more useful direction would be to have one command that would perform both kniting and pdf-latex in one step. Then hiding .tex files would make sense.

mmaechler:
Ok, so we do have to discuss this.. Maybe ask for comments on ess-help (referring readers to here).
I disagree with @vspinu here.  We should follow ESS's standard "way of thinking" namely our old motto
"The source is real" ... and in this case, the source is the `*.Rnw`. Note that in an advanced working setting (which I think still *did* work when I tried), using the corresponding CRAN package function
`patchDVI::patchSynctex('$*.synctex')`  (this is from a Makefile of mine), one can click in the PDF viewer and directly jump to the *.Rnw file (instead of just the *.tex file) and fix the bug there.
Working with *.tex -- which I tend to do in case of LaTex'ing errors is already error prone now: Often I fix the bug in the *.tex and forget to synchronize back to the underlying *.Rnw... 

All in all we should strongly strive for the user to only work with the *.Rnw  (and *.Rmd in that context) rather than with their outputs .. which are "not real" (in the sense of our motto).

vspinu:
> We should follow ESS's standard "way of thinking" namely our old motto

I agree completely with you and I often modify wrong document myself. Nevertheless, `ess-swv-knit` does knitting and hiding its output doesn't sound right. This is why I am proposing to have a one-blow command instead of this workaround.

I don't have strong opinion on this and I am not using ess-swv anymore so please feel free to merge this if you like.

mmaechler:
As I have released  ESS version `15.09-1` now, we can wait a bit longer for other comments and ideas or also for @Jehops (or you or someone else) proposing another patch which has your proposed "one-blow" command.

Jehops:
Martin Mächler <notifications@github.com> writes:

> The proposed changes look good to me ((but I did not even try them)).
> You mention / changed 'knitting'  only, but not Sweaving. 

>  Shouldn't there be analogous changes (i.e. not showing the "log" buffer by default unless in case of errors)?  Or are they not needed there?  {{sorry for being lazy and not checking myself}}

The changes affect all related functions (weaving/tangling knitting/purling) since they all call `ess-swv-run-in-R`.

Jehops:
A "one-blow" command makes sense to me as well.  I'll ask for comments on the list.  One problem I'm having (and I now see it's a problem with the patch I submitted) is determining when the kniting/weaving has finished before trying to compile the TeX output.  For that I think we need something like a synchronous version of `ess-execute`.  Am I missing such a function in ess-inf.el?

Jehops:
I've added the "one-blow" command in the second commit.

mmaechler:
Thank you @Jehops

Jehops:
Thank you @mmaechler.  I use ESS a lot, so I'm happy I can chip in a little.

PR #281, opened 2015-12-06
Specify R arguments non-interactively

lionel-:
This makes it a bit easier to call R with different arguments from elisp code. Now when `start-args` is a string or a list of strings, R will start with those arguments without prompting the user. Previously you had to let-bind `R-inferior-args` to do this.


lionel-:
I thought about making it variadic with `&rest`, but then this would preclude us to add other arguments to `(R)` in the future.

PR #282, opened 2015-12-07
Improve devtools integration

lionel-:
This improves devtools integration in several ways.

##### Additional functions:

* loading package with a prefix forces a recompile. With a double prefix, the package is unloaded.
* new checking method binded to `C-c C-t c`. With no prefix, run tests. With single prefix, run R CMD check. With double prefix, check only documentation
* new documentation method. I didn't find a good keybinding for it though.

All these commands run through `(ess-developer-send-process)`. I thus deprecated the variable `ess-developer-load-package-command` since users can use the function to send any custom commands.

##### Keeping track of packages

A new buffer local variable `ess-developer-local-package` can be set to a cons cell with package info: `(name . path)`. This way we can use devtools commands without activating developer mode. It also works in files outside a package hierarchy.

Furthermore, using a devtools command will set this local variable in the inferior buffer as well. This way the commands work when called from that buffer and it is useful to create functions using this information. For example I use it to reload R and the current package automatically (useful when debugging a crashing function).

##### Smarter package selection

This one is not related to devtools. When adding a package, the user is no longer prompted for the path when the selected package is the same as the package containing the current file.

In addition, I'd like to never ask for the path when we call a command from a package file. Would there be objections for this?


vspinu:
Hi Lionel, All these looks truly great. Unfortunately I am just starting my very long trip to Asia. I will be able to have a closer look at it in a couple of days only. Same holds for your other patch (#281). (But I can also look at them even you have merged them.)

Is the plan to keep all shortcuts in dev-map? I would like that but I am afraid we don't have that much free space there.  We will need a lot of devtools related keys. So maybe keep everything in one devtools map? Where to hook it then?

lionel-:
Don't worry Vitalie.

I will hold those PR here until the new patch release has come out. I'm not really qualified to come up with good Emacs-style bindings because as an Evil user I don't use them.

 I wish you a very nice trip!

lionel-:
@vspinu I wonder if the whole `ess-developer-packages` list is really necessary. I don't use the code injection mechanism so I'm not sure about this, but I've been using the devtools integration (which doesn't rely on `ess-developer-packages`) for a few months and it works quite well. Packages are determined with:

- whether we are in a typical package folder (R, src, man, etc). I'm currently adding a `ess-developer-check-all-dirs` variable that can be switched off when working on remotes. When `t` the whole folder hierarchy will be checked instead of just the typical folders.

- A file or dir local variable `ess-developer-local-package`.

This allows ESS to select the right package pretty much all the time.

By comparison, maintaining a list of known packages and switching manually between them seems cumbersome. Should we deprecate this?

lionel-:
@vspinu @mmaechler 

Would `C-c d` be a good prefix for `r-devtools-map`? I'm asking as an Emacs keybindings neophyte ;) Seems to be unused.

vspinu:
> @vspinu I wonder if the whole ess-developer-packages list is really necessary.

I have started working on improving the workflow some time ago ([devUI](https://github.com/emacs-ess/ESS/compare/devUI) branch). And it goes in line with your desire. But  ess-developer-packages` is still needed if you want to source into a foreign package on the fly. You will need to add a package into the development list and activate ess-developer. I often do that with core packages or packages which I don't want to clone.

> whether we are in a typical package folder (R, src, man, etc). I'm currently adding a ess-developer-check-all-dirs

If possible, we need to strive for the same behavior on remotes and locals. Why is this needed? For `ess-developer--get-package-path`? 

> Would C-c d be a good prefix for r-devtools-map? 

Not really. `C-c LETTER` bindings are reserved for users. We cannot use those.

lionel-:
> If possible, we need to strive for the same behavior on remotes and locals. Why is this needed? For ess-developer--get-package-path?

Yes, you had a note in the code that checking all folders was slow on remotes. But it'd still be nice to check all folders when working locally.

> But ess-developer-packages` is still needed if you want to source into a foreign package on the fly.

How about a `(ess-developer-select-package)` function that would wrap `(setq-local ess-developer-local-package ...)`?

vspinu:
> How about a (ess-developer-select-package) function that would wrap (setq-local ess-developer-local-package ...)?

I am not sure what's the added benefit of it? It seems less general than the current state but I a open to suggestions. If we can streamline UI and simplify the code I would be happy to have it.

Just to be clear, the current situation is like follows. There is `C-c C-t C-a runs the command ess-developer-add-package`. You add a package on the fly with `C-c C-t C-a` and toggle developer mode with `C-c C-t C-t`.  We can determine which package the function belongs to so we can have multiple packages in that list.

We can in principle merge those two steps in one. The user can be asked for package when the developer mode is turned on  on a file which could not be associated with a package.

> I don't use the code injection mechanism

My idea was to make code injection automatic for in-package code, even if the package is not listed in `ess-developer-packages`. `devtools` is good, but it's complementary to the original goal of ess-developer - to eliminate the distinction between script and in-package development.  I think currently we do quite ok on that.

vspinu:
I have looked through the code. Looks good. Could you please merge it in? 

Do you have other ideas on the short cut?  We can in principle try to fit all devtools commands in C-c C-t. It will be hard but it's not impossible.

lisp/ess-developer.el
```diff
@@ -377,21 +430,81 @@ If ALL is non-nil, deactivate in all open R buffers."
...
+         (ess-developer-send-process "devtools::load_all('%s')\n"
+                                     "Loading %s"))))
+
+(defun ess-developer-unload-package ()
+  "Interface to unload() function from devtools package."
+  (interactive)
+  (ess-developer-send-process "devtools::unload('%s')\n"
+                              "Unloading %s"))
+
+(defun ess-developer-check-package ()
```
vspinu: How about having a prompt with completion on C-u? It would be nice to have various options for check: with/out manuals, vignetes, CRAN etc.

lisp/ess-developer.el
```diff
@@ -377,21 +430,81 @@ If ALL is non-nil, deactivate in all open R buffers."
...
+         (ess-developer-send-process "devtools::test('%s')\n"
+                                     "Testing %s"))))
+
+(defun ess-developer-document-package ()
+  "Interface to document() from devtools package."
+  (interactive)
+  (ess-developer-send-process "devtools::document('%s')\n"
+                              "Documenting %s"))
+
+(defun ess-developer-install-package ()
```
vspinu: I am often running into problems when re-installing the package after the package has been already loaded. 

We need a way to have a clean re-start of R and then package install. We can also have automatic re-load of all currently loaded libraries. We can even re-load the workspace.

lionel-: This is a bug in R core, fixed in devtools by the mighty @jimhester in https://github.com/hadley/devtools/pull/1001 :)

lisp/ess-developer.el
lionel-: Hmm... `test()` (normal) and the full `check()` (C-u) are the most common operations. Maybe have the prompt on C-u C-u? And then have `check_revdep()` on another keybinding.

lionel-: Or have three functions, test, check (prompt on C-u) and check_revdep, and we bind those on a dedicated keymap `r-devtools-check`.

lionel-:
> Do you have other ideas on the short cut? We can in principle try to fit all devtools commands in C-c C-t. It will be hard but it's not impossible.

I use my own shortcuts with `evil-leader` so I'm not sure. I think it's important to have good mnemonics so fitting everything on C-c C-t doesn't seem right.

lisp/ess-developer.el
vspinu: > This is a bug in R core,

Great news.

I still think R re-start and optional re-attachment of all loaded packages is a good idea.

lionel-:
> We can in principle merge those two steps in one. The user can be asked for package when the developer mode is turned on on a file which could not be associated with a package.

Yes I think just setting the package in the file-local variable + smart package detection should be enough. Then we can simplify the UI and the code. But it's entirely your call :)

> devtools is good, but it's complementary to the original goal of ess-developer

Yup that's what I also think.

lisp/ess-developer.el
vspinu: My common operation is to check without the vignette. I poropose that we have an allist of possible check commands with the default being the first one in the list. It's one extra RET after C-u but it saves us a lot of trouble. 

A major benefit is that users can add their own check commands with whatever additional build args they need.

It might be good to extract all devtools commands into user vars for UI consistency.

lisp/ess-developer.el
lionel-: > I still think R re-start and optional re-attachment of all loaded packages is a good idea.

Right, actually I have such a function that I use all the time. It only reload the currently activated package though (`ess-developer-local-package` gets redefined in the inferior buffer each time a devtools command is invoked). 

```r
(defun haddock-inferior-ess-reload ()
  (interactive)
  (let ((package ess-developer-local-package)
        ;; No confirmation when killing the process
        kill-buffer-query-functions)
    (kill-buffer)
    (if current-prefix-arg
        (R)
      (neat-R))
    (when package
      (setq-local ess-developer-local-package package)
      (ess-developer-load-package))
    (when ess-replay-cmd
      (ess-replay))))
```

lisp/ess-developer.el
lionel-: > It might be good to extract all devtools commands into user vars for UI consistency.

On the other hand it's easy enough for a user to bind her own functions wrapping `(ess-developer-send-process)`. Is it really necessary to complicate the code/UI?

vspinu:
One option is to hook the whole devtools on C-c M-t or C-c M-d.  

Or obsolete C-c C-t C-t as we discussed above. This one as it's very fast to type.

> I use my own shortcuts with evil-leader so I'm not sure.

So you rebind keys for all modes you use?

Always wanted to ask, have you switched from VIM to emacs?

lisp/ess-developer.el
vspinu: You don't need to kill the buffer. Just call `ess-quit` and start the package with R. I am doing this manualy all the time but never got down to implementing it :(

lisp/ess-developer.el
vspinu: Valid point, but that would also involve binding to a key. Let's see what we end up with. It's more about the consistent code than UI, IMO. It's not good to leave some commands with custom vars and others without.

lionel-:
> Or obsolete C-c C-t C-t as we discussed above. This one as it's very fast to type.

This one looks like the best option for now. And we can always change it before the next release if we find something better.

> So you rebind keys for all modes you use?

Only for the commands I use often.

> have you switched from VIM to emacs?

Nope, I switched directly from BBedit to Emacs Evil :)

lisp/ess-developer.el
lionel-: How about the following consistent scheme:

* C-u → alternative command
* C-u C-u → main command with custom args

lionel-: so for checking:

* nil → `check()`
* C-u → `check(vignettes = FALSE)`
* C-u C-u → `check()` with prompt

For loading:

* nil → `load()`
* C-u → `load(recompile = TRUE)`
* C-u C-u → `load()` with prompt

vspinu: I actually meant:

    - nil → load()
    - C-u → load() with prompt

We place `alternative` command as the default in the list (press RET and get it). Otherwise the system becomes too complex. User has to remember all these C-u things and the alternative command. We would also need to add configuration both for the alternative (C-u) and the list (C-u C-u) of commands.  Not all users care about Vignetes.

For one RET speedup it's not worth it.

vspinu: One problem is that the shortcuts will be too long: `C-u C-c C-t C-t c`. I am always forgetting C-u and typing the command twice. We can have the with prompt command on the shifted last key `C-c C-t C-t C`, so that when you get to the last key you will decide if you want the default "check" or prompted "check".

lionel-:
@vspinu 

in `(ess-developer-source-current-file)`, shouldn't 
```r
source(file="%s", local=F)
```
be
```r
source(file="%s", local = TRUE)
```
?

I'm looking into deprecating `ess-developer-packages`, is it ok to have `source()` on C-u and `.essDev_source()` with no prefix? This way no need to prompt the user about anything.

lionel-:
by the way I'm getting freezes with `(ess-boolean-command "2>1")`, may be related to your last changes, can you please take a look?

lionel-:
Same thing for `(ess-developer-send-region)`, there is this `*current*` thing. I propose to send to current when:

- Called with `C-u`.
- No package can be found (in the folder hierarchy or in the local variable)

vspinu:
> source(file="%s", local=F)

I think yes. It doesn't make much sense otherwise. I wonder how it was working before.

`*current*` makes a difference only when the R is in the debugger mode. I guess you got that part.

> by the way I'm getting freezes with (ess-boolean-command "2>1"), may be related to your last changes, can you please take a look?

It should be `(ess-boolean-command "2>1\n")`.

vspinu:
> Same thing for (ess-developer-send-region), there is this *current* thing. I propose to send to current when:  Called with C-u

Not sure what you mean here. C-u is already taken in our eval commands. Now user can always toggle the developer off, but if we plan to remove C-c C-t C-t, it will not work anymore.

I often use this feature. It's not uncommon when you want to evaluate portions of the package code into current environment. Especially when in the debuger mode.

mmaechler:
Just a late general comment:  I do not use devtools, as I do differently what they provide, also as "they" promise things that are not possible {reloading a package a 2nd time without restarting R and get the *same* state as when loaded the first time, is *not* possible currently}.
Also I don't want that we insinuate (what many people do) that smartly using R as a developer means to use `devtools`.     Consequently, I'd rather **Not** have such stuff on keys that I press accidentally.

lionel-:
I'd suggest not to press keys accidentally then ;)

>  {reloading a package a 2nd time without restarting R and get the same state as when loaded the first time, is not possible currently}.

Maybe not possible, but devtools does a pretty good job. Saves tons of time.

vspinu:
The keys will be pretty long, so pressing them accidentally will rarely happen.  Devtools actually never restarts R. That was my proposal - a command to restart R and optionally re-load state and re-attach packages.

lionel-:
I'll soon upload the discussed changes. In addition, I've bound the devtools map to C-c C-w. Is that's ok?

lionel-:
At one point we should discuss which prefixes are reserved for ESS wide commands, and which prefixes can be specific to modes/languages.

lionel-:
@vspinu I can't get it to work with `ess-quit` :/

With this modification of `ess-quit-r` (new argument `no-save`):
```lisp
(defun ess-quit-r (&optional no-save)
  "Issue an exiting command to an inferior R process, and
optionally clean up.  This version is for killing *R* processes;
it asks the extra question regarding whether the workspace image
should be saved unless NO-SAVE is non-nil."
  (ess-force-buffer-current "Process to quit: " nil 'no-autostart)
  (ess-make-buffer-current)
  (let (cmd
        ;;Q     response
        (sprocess (ess-get-process ess-current-process-name)))
    (if (not sprocess) (error "No ESS process running"))
    (ess-cleanup)
    (setq cmd (format "base::q('%s')\n" (if no-save "no" "default")))
    (goto-char (marker-position (process-mark sprocess)))
    (process-send-string sprocess cmd)))
```

Then:
```lisp
(defun r-reload-inferior (&optional start-args)
  "Reload R and the currently activated developer package, if
any."
  (interactive)
  (let ((pkg-info ess-developer-local-package)
        (r-proc (ess-get-process)))
    (ess-quit-r 'no-save)
    (unless (memq (process-status r-proc) '(nil exit))
      (ess-wait-for-process r-proc))
    (R start-args)
    (when pkg-info
      (setq-local ess-developer-local-package pkg-info)
      (ess-developer-load-package))))
```

This works when stepping through it, but with an actual run the new process gets created in a new buffer with the process name incremented by 1. I think it doesn't wait for the process to complete the quitting command. Do you know why is that the case?

lionel-:
@vspinu Here is an update on the new changes.

* A package can be selected for the current buffer with `C-c C-t C-t`.

* Alternatively, it can be selected and permanently added to the file-local variable `ess-developer-local-package` with `C-c C-t t`.

* Code injection into the current environment can be activated locally with `C-c C-t C-j`.

* Code injection to a package can be activated locally with `C-c C-t j`. It uses package detection to propose a good relevant first choice in addition to all installed packages. (Btw, could we get rid of all those `attached-only` arguments?)

* When `ess-developer-code-injection-in-packages` is non-nil, code injection is automatically activated in all buffers visiting a package file, or a file with the local variable `ess-developer-local-package` set.

Let me know if it all adds up to a comfortable workflow. Should I merge at this point so it's easier to test?

vspinu:
> Here is an update on the new changes.

We had two commands before - C-c C-t C-t to toggle the developer and C-c C-t C-a to add a package. Now we have 4 commands.  First two seem to be doing what C-c C-t C-a was doing before, so we better keep it on C-c C-t C-a. Can we make the 3rd and 4th command into one toggling command? I would also put it on C-c C-t C-c as a mnemonic for "current".

I am not a big fan of C-c C-w for a few reasons. First, it would be good to have all development commands on one keymap and let the user decide if they want to move a submap to a more handy key. Secondly, C-c C-w sounds like a command related to yanking or popping and some modes a binding it to yank related stuff. And finally, we inside ESS, were trying to keep "w" keys to something web related (ess-help-web-search, on C-c C-d C-w, ess-display-help-in-browser on w in help keymap).

So if C-c C-t C-t is free now, let's hook it in there. 


>  ok, I think I got it right.

Great! Thanks. 

> could we get rid of all those attached-only arguments

@mmaechler added `attached-only` which has to do with his setup (all 4000+ packages are located on a network drive). So either we leave it as it is, or think of an automatic and fast detection system for such cases. Something like "if nr installed packages > 1000 then we look into attached only"

> At one point we should discuss which key prefixes are reserved for ESS wide commands, and which key prefixes can be specific to modes/languages.

Let's try to stick to `ess-MODE-something`. Particularly, those `r-devtools-*` should be `ess-R-devtools`. If we ever end up modularizing or re-writing ESS this will avoid name clashes.

lionel-:
> Let's try to stick to ess-MODE-something. Particularly, those r-devtools-* should be ess-R-devtools. If we ever end up modularizing or re-writing ESS this will avoid name clashes.

I was talking about keybindings. About namespaces, I would personnally prefer `sas-`, `stata-`, `r-` namespaces.

- `ess-` adds a lot of noise and bytes to our code
- `julia-` already has its own namespace
- It would make the distinction between ESS wide and language-specific functions and variables crisp clear
- This would be consistent with `cc-mode`, which has `c-mode`, `java-mode` etc and not `cc-c-mode`, `cc-java-mode`, ...

Also I would avoid caps in namespaces. I find them not aesthetics, hard to read and hard to write (the worst case being `M-x`ing a `LaTeX-` command). There are currently all sorts of capitalisation for R in ESS code, so it's not too late to change once we get to modularise ESS.

Anyway, that would be a long term change to be discussed on the mailing list when we get to ESS modernisation. We don't have to figure it out now and we can use `ess-R-devtools` for the time being. Or `ess-r-devtools` since there are already many instances of uncapitalised R and which I prefer in terms of aesthetics, number of key touches, and overall noise in source code.

lionel-:
> Can we make the 3rd and 4th command into one toggling command?

Would it feel right, given that one of them prompts the user? Alternatively, I could add *current* as a second choice in the list of packages. Or should it be first choice, with detected package as second chocie? I guess you must use this often when you debug?

For the 1st and 2nd commands (adding a package temporarily and permanently), I could merge them with a C-u switch.

I think your proposed keybindings are great, I'll upload an update.

lionel-:
> For the 1st and 2nd commands (adding a package temporarily and permanently), I could merge them with a C-u switch.

Thinking more about it, I'm not sure adding a package temporarily is even necessary. We can remove that command I think.

vspinu:
> julia- already has its own namespace

julia mode is external package. All our functions are prefixed with `ess-`. I would like to see a rewrite of r-mode and make it stand alone, such that other potential systems besides ESS could use it. This is why it's important to keep things with `ess-` to avoid future clashes.

> This would be consistent with cc-mode, which has c-mode, java-mode etc and not cc-c-mode, cc-java-mode, ...

Precisely, but unfortunately current ess-r-mode is not up to standards. It's tightly coupled with ESS and is rather messy. 

> Also I would avoid caps in namespaces. 

Sure. I am not a big fan either.

> Would it feel right, given that one of them prompts the user?

Hm. Maybe you pack things the way you see most fit and then I try them live and tell you my opinion?

> We can remove that command I think.

Yes. Let's keep things clean and with minimum commands for start.

lionel-:
> I would like to see a rewrite of r-mode and make it stand alone, such that other potential systems besides ESS could use it. This is why it's important to keep things with ess- to avoid future clashes.

Well if we claim the `r-` namespace there won't be any clashes ;)

> unfortunately current ess-r-mode is not up to standards. It's tightly coupled with ESS and is rather messy.

Hopefully we'll be able to come up with a modern design when we modularise ESS.

> Maybe you pack things the way you see most fit and then I try them live and tell you my opinion?

Just implemented a stub of the toggling version. Would be nice if we could repeat a C-c to toggle. I.e., C-c C-t C-c for `*current*`, and if we stroke one additional C-c we select a package. Is this possible?

Anyway, merging in so you can test.

vspinu:
> and if we stroke one additional C-c we select a package. Is this possible?

Yes. Such toggling is implemented in a couple of commands. C-c C-t C-t was one of them. `set-temporary-overlay-map` should be probably used for that. There is `ess--execute-electric-command` which I wrote before I discovered `set-temporary-overlay-map`. C-c C-d C-e is currently based on it. 

See also C-c C-t C-b and C-c C-t C-e for examples of how it is done directly without those overlay maps. I wrote those even before writing `ess--execute-electric-command`. In any case, it would be great to clean all of that and use an uniform style instead.

vspinu:
> Hopefully we'll be able to come up with a modern design when we modularise ESS.

This is unlikely. Everything is too much coupled together with big piles of historical material. My bet is on a rewrite from scratch taking advantages of all the new tools in emacs world which were not available 20 years ago.

lionel-:
> My bet is on a rewrite from scratch taking advantages of all the new tools

Ok, then let's stick with `ess-r-` for the legacy R mode, and maybe we'll use `r-` for the rewrite :)

lionel-:
@vspinu `set-temporary-overlay-map` is obsolete since 24.4 in favour of `set-transient-map`. What's the policy on this? Use the obsolete version or the new one with a check for Emacs version?

vspinu:
No particular policy. We want to support emacs 24. If new function does a better
job then checking version is better.

This particular case is tricky. The last argument (ON-EXIT) is a recent addition
to emacs (around emacs 24.4 I think). So using it would imply that we are not
supporting earlier emacses.

So maybe just stick with my function for now. Once we drop emacs < 24.3 we
remove it.

vspinu:
Hi Lionel, I finally got some time to play with changes.  And I am afraid I don't quite follow what you did. Could you let me know how you imagined the injection UI? It seems to be very different from what we talked. 

These are the points of my confusion:

  - What is the current workflow of the code injection (not devtools)?
  - When I open a package it's not detected automatically. I am sourcing into global env. Is that intentional?
  - You removed mode-line indicator. I think it's a bad idea. The user needs to know which package he or she develops.
  - Documentation is not very informative of the UI. I can only guess what `ess-developer-local-package` and `ess-developer-code-injection` means and what's the difference between them.  ess-developer-inject-to-package and ess-developer-select-package have almost the same doc string with no information for end user but only technical details of where the variables are saved.
 - I find the distinction between `ess-developer-inject-to-package` and `ess-developer-select-package` artificial. Why do we need two functions?

Thanks.

vspinu:
I think the notion of a development mode is still very useful. Users might want to have their custom initialization on entrance.

lisp/ess-developer.el
```diff
@@ -30,130 +30,248 @@
...
+                 (concat ", " args))))
+    (message msg name)
+    (ess-developer--update-process-local-pkg pkg-info)
+    (ess-eval-linewise (format command (concat path args)))))
+
+
+;;; Package Detection
+
+(defun ess-developer--find-package-info ()
+  "Parses DESCRIPTION file in PATH (R specific so far). PATH
```
vspinu: First line of doc strings must be self contained sentence.  See [here](http://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) for conventions. 

Please run M-x checkdoc on new modules and docstrings that you add. Our older modules are screwed, but at least we should keep good doc strings for new stuff.

lisp/ess-developer.el
```diff
@@ -30,130 +30,248 @@
...
-  "Normal hook run on entering `ess-developer' mode."
-  :group 'ess-developer
-  :type 'hook)
+project root.")
 
-(defcustom ess-developer-exit-hook nil
-  "Normal hook run on exiting `ess-developer' mode."
-  :group 'ess-developer
-  :type 'hook)
+(defvar ess-developer-check-all-dirs t
```
vspinu: Is this a user level var? Do we really need this variable?

lisp/ess-developer.el
```diff
@@ -30,130 +30,248 @@
...
 ;;; Code:
 
-;; (require 'ess-site) ;; need to assigne the keys in the map
-
 (defgroup ess-developer nil
   "ESS: developer."
   :group 'ess
   :prefix "ess-developer-")
 
-(defface ess-developer-indicator-face
```
vspinu: I am placing this stuff back. We really need a mode line indicator of the current environment.

lisp/ess-developer.el
lionel-: I added this because there was a note in the code that checking all dirs is slow on remotes. So the user has a choice of disabling folder lookup

lisp/ess-developer.el
lionel-: Thanks, I will check this.

lionel-:
> What is the current workflow of the code injection (not devtools)?

C-c C-t C-c one time to inject to current environment, then another time (I'll add the better toggling in the future) to inject to package. The first package proposed it the one detected so a simple RET is needed. Alternatively, you can activate `ess-developer-code-injection-in-packages` to have automatic injection.

The idea behind separating package and injection interface is that injection to current environment is useful outside of packages for step-debugging (IIUC).

> You removed mode-line indicator. I think it's a bad idea. The user needs to know which package he or she develops.

You're right, I can restore the mode-line to indicate what package is currently active.

> Documentation is not very informative of the UI.

Indeed I intend to review the doc and add news items etc.

> I think the notion of a development mode is still very useful. Users might want to have their custom initialization on entrance.

Would it be ok to check each opened file for a package, activate the modeline and run the custom intialisation if a package is detected? While working with devtools commands I thought automatic detection made the best workflow. If you prefer I can restore developer-mode.

Or, activate developer-mode automatically when a package is detected, and link the modeline and initialisation stuff to that mode as before. I thought that the mode lost most of its purpose when we decided to remove the package handling but I may have gone overboard with removing it :) It's still useful to run hooks and have the mode-line indeed...

lisp/ess-developer.el
vspinu: I think `ess-developer-package-dirs` is comprehensive enough. no? We can add a special setting to that list. Or if that list is nil, then we check whole hierarchy. No need for overlapping coinfgs IMO.

lionel-: It's not when you're far in the hierarchy in `include/` or `src/`. I think the `nil` idea is good.

vspinu:
> The idea behind separating package and injection interface is that injection to current environment is useful outside of packages for step-debugging (IIUC).

I think it's a complication. In each file there could be only one current environment/package active. So having one buffer local variable to indicate the environment should be enough in principle. We can use a global cache variable for package locations if you think caching is useful.

> Would it be ok to check each opened file for a package, activate the modeline and run the custom intialisation if a package is detected? 

If I understand you correctly, this is exactly what was happening before. Packages were automatically detected and developer mode was initialized in each open file. You removed that code.  I think you didn't have a chance to experience that because your `ess-developer-packages` were nil.

vspinu:
BTW, developer mode is not really a mode, it's just an entry and exit hook which users can customize. It might be worth to make it a proper minor mode.

lisp/ess-developer.el
vspinu: Yes. But the current path will contain one of those dirs in the list for sure, and you can start looking from that directory down, no need to descend from current file.

lionel-:
> It might be worth to make it a proper minor mode.

Ok, I'll restore it as a minor mode. Sorry for the confusion.

> I think it's a complication. In each file there could be only one current environment/package active. So having one buffer local variable to indicate the environment should be enough in principle.

But that means that injecting code to `*current*` would prevent us from using devtools without setting back to the package... That could lead to a lot of back-and-forth, which is why I think it's better to keep both separate.

lisp/ess-developer.el
lionel-: ha, of course :)

lionel-:
Or we just have two commands, one for injecting to current env, the other for injecting to current package? This would work with a unique package variable while avoiding the problem I just mentioned.

vspinu:
> Ok, I'll restore it as a minor mode. Sorry for the confusion.

It wasm't a minor mode before. No need to restore. Let's think what's the best and most concise way to achieve what we want.

> But that means that injecting code to *current* would prevent us from using devtools without setting back to the package..

Right. That's a problem. 

> Or we just have two commands, one for injecting to current env, the other for injecting to current package? 

I actually was thinking to extend this functionality and to allow sourcing code into any visible user defined environment. Maybe having two variables is indeed unavoidable. 

It will be hard to display both current dev package and current environment in the modeline.  

Ok. Let sleep on these things. I started going through the code and making small changes here and there. 

Are you working on the right now? Would be good not to clash.

vspinu:
BTW, with new names the mnemonics for shortcuts are lost. It feels awkward quite frankly.

lionel-:
> Are you working on the right now?

No, I have a deadline tomorrow, and will be on vacation for a week after that.

> BTW, with new names the mnemonics for shortcuts are lost

You mean for C-c C-t C-a / C-c ?

vspinu:
> No, I have a deadline tomorrow, and will be on vacation for a week after that.

Ah. Ok. Then go to work. Don't bother with this. It can wait.

> You mean for C-c C-t C-a / C-c ?

Yes. Also C-c C-t C-t xx is mind blowing. I find myself forffgetting how many times I pressed C-t and already confusing C-t and C-t C-t map. I think it's a bad key :(

lionel-:
well it was your suggestions ;) I think of C-a as "Adding" a package to the file-local vars section. But yes C-c is good for injecting to current but not for package. I had these on `C-j` initially, which I somehow find a good mnemonic for injection, but that could be just me.

> Also C-c C-t C-t xx is mind blowing

If we don't find a better key, we should at least suggest in the documentation that devtools users should bind the keymap to `C-c d` or something like that.

> Then go to work.

Aye :)

vspinu:
> well it was your suggestions ;) 

Yeh, and it sucks. My hope was that it won't surprise existing users. But now it looks to me that the functionality is so different that this argument doesn't hold anymore.

BTW, why do we need to set the package once again? I think we can do just fine without. Then have one command to set an injection environment in a broad sense (either a package or current environment or user defined environment).

lionel-:
> BTW, why do we need to set the package once again? 

It's for files outside a package. I often have a scratch file for working on packages and so it's nice to have devtools commands always work correctly while I'm visiting that file. So C-c C-t C-a adds the information to the file-local section of that file. But it should be a rarely used command.

vspinu:
For this use pattern we can do without the second command. You can select the source environment for that buffer. If environment is a package, we activate the developer mode there.  Problem solved.

We will need a real developer mode in order to automatically activate devtools and some parts of ESS functionality in C/RCpp and other relevant files.

lionel-:
> For this use pattern we can do without the second command. You can select the source environment for that buffer. If environment is a package, we activate the developer mode there. Problem solved.

Hmm I'm not sure I get it. The point is to set what package should be used for a given file or directory once and for all.

lionel-:
Btw, I have other good stuff that I will integrate in the future, like automatic detection of R process for gud debugging.

vspinu:
> Hmm I'm not sure I get it. The point is to set what package should be used for
> a given file or directory once and for all.

If you are developing from external file you are likely to want to source in the
package, no? What's the point of developing from outside? Even if you don't want
to develop, injection happens only for known objects. All others are assigned
into GlobalEnv. So it's hardly makes any difference. I think it's a small price
to pay for avoiding duplicated vars and user level commands.

> Btw, I have other good stuff that I will integrate in the future, like
> automatic detection of R process for gud debugging.

What would it bring over the tracebug? You cannot go much further existing
facilities in R, and tracebug already exploits almost all of those.

It's not really a visual debugger, but I don't know much about it. It's not well
documented.

lionel-:
>  Even if you don't want to develop, injection happens only for known objects

I didn't know that. Nice!

> So it's hardly makes any difference.

I still think that such side-effecty functions should be optional. One user var does not cost a lot. In my current workflow I rely entirely on devtools — `load_all()`, `install()`, and the occasional restart of R session. I'll try code injection the next time I work on a package but I think it should be optional.

> What would it bring over the tracebug?

I meant for debugging compiled C/C++ code ;)

vspinu:
> One user var does not cost a lot. I

Indeed. But one more confusing comand does. And I am not sure if your use case is worth it. It doesn't make much sense to use `load_all()` and `install` on a package unless you are developing that package.

One easy solution would be to prompt for package in `load_all` when you are outside of the package. It would cost you one RET per load, but would result in a cleaner code and UI.

lionel-:
> It doesn't make much sense to use load_all() and install on a package unless you are developing that package.

It makes sense that, when you trigger load_all() or any devtool command in a file belonging to a package, it uses automatically the relevant package path. I use those devtools commands like 50 times a day (if not a 100), I don't want to enter RET each time when there is no need...

But I'm not sure I understand what you're saying concretely. If I open a package file, I'm developing that package indeed.

> One easy solution would be to prompt for package in load_all when you are outside of the package. It would cost you one RET per load, but would result in a cleaner code and UI.

Specifying one buffer-local var does not make the UI dirty. Minimalism has its virtues but at the same time Emacs needs to be flexible and do the right action with minimal hassle.

vspinu:
>  I don't want to enter RET each time when there is no need...

I was not proposing that.

> But I'm not sure I understand what you're saying concretely. If I open a package file, I'm developing that package indeed.

I was reffering to your corner case example when you said you want to have devtools in a file which doesn't belong to a package. I said that's rare and there is no point to have a separate command to set some local var just for that corner case.

lionel-:
> I said that's rare and there is no point to have a separate command to set some local var just for that corner case.

In my workflow it's not a corner case. I have one such file (sometimes several) for every package I develop / contribute to.

It's not a big deal not to have the command, but it's not a big deal providing it either. Maybe just don't provide a keybinding for it?

vspinu:
After sleeping on it a bit, I think you are right. We won't be able to isolate these two things. They are quite distinct conceptually, and accessing devtools need not even be from the R sources. 

This is how I am seeing it right now. First we need a minor mode `ess-developer-mode`. We will activate this mode in all files inside an R package and will place all the functionality which is meaningful ouside R source on `ess-developer-mode-map`. We can use the prefix that you proposed - `C-c C-w`.  The injection functionality makes sense in R buffers and is tightly connected with evaluaiton mechanism, so let's keep it on C-c C-t. 

I propse to rename it to `ess-set-source-environment` and hook it on C-c C-t C-s. It's easier to type. 

The package selection command could be named `ess-set-developer-package` and maybe hooked C-c C-w C-s in `ess-devloper-mode-map`. This way `C-s` will be responsible for similar actions in these two maps.

What do you think?

lionel-:
This sounds good, especially binding the map within the minor mode. We'll have the map activated inside Rcpp files, which is a huge improvement.

Where would we bind the devtools map? `C-c C-w C-d`? Or do you want to integrate all devtools commands in `ess-developer-map`? I see two problems with the latter option: devtools command are R only; and we already have a lot of devtools commands and we're not even finished integrating them all.

Otherwise this all sounds great.

vspinu:
> Where would we bind the devtools map? C-c C-w C-d

All devtools will be on C-c C-w directly. They operate per package not per file. What do you mean by "devtools command are R only"?  If you are concerned about non R dialects in ESS, then we can call it `ess-R-developer-mode`. `ess-developer` is R only anyways.

> and we already have a lot of devtools commands and we're not even finished integrating them all.

They are a lot because you needlessly made two commands for each task. Once we have one command per task, we will be fine. We have full map available; it should be doable with good mnemonics.

I think there will be very few non devtools commands on that map.

lionel-:
> They are a lot because you needlessly made two commands for each task.

I doubled the commend on your request because you didn't like to use C-u ....

> If you are concerned about non R dialects in ESS, then we can call it ess-R-developer-mode. ess-developer is R only anyways

Or `ess-r-developer-mode` as we agreed above to use non-capitalised r prefix.

vspinu:
> didn't like to use C-u ....

I think it was a miscommunication. I proposed to put optional forms of the command on C-u. C-u should ask for alternative arguments with completion. Not like now, when only one alternative is executed.

> Or ess-r-developer-mode 

A bit of an inconvenience is that we will have ess-dev-map and `ess-r-developer-mode-map`. This is confusing. Better names for these? How about `ess-r-package-dev-map` and thus `ess-r-package-dev-mode`?

lionel-:
> I proposed to put optional forms of the command on C-u. C-u should ask for alternative arguments with completion.

No, that's how it worked before. Since the map prefix was a bit long (C-c C-t C-t), you found it inconvenient to have alternative prompts on C-u.

> How about ess-r-package-dev-map and thus ess-r-package-dev-mode?

I like these.

lionel-:
Maybe the shorter `ess-r-package-mode` is good enough?

vspinu:
`ess-r-packdev-mode`?

vspinu:
I think `ess-r-package-mode` is not suggestive enough. `package` could mean anything.

lionel-:
> ess-r-packdev-mode

I prefer your former suggestion then, `ess-r-package-dev-mode`. But will we have to prefix all functions like this? In this case we're better off with a short prefix like `ess-r-dev-`.

vspinu:
> In this case we're better off with a short prefix like ess-r-dev-.

I thought to keep ess-r-devtools prefix for functions even if they reside in package-dev map.  We already have `ess-dev-map` with heterogeneous functions in it. I think having `ess-r-dev-map` would be too confusing to users who would like to re-assign keys.

PR #295, opened 2016-02-24
Add asynchronous evaluation for inferiors

lionel-:
This adds `(ess-inferior-async fun &optional proc)` to execute `fun` once the process is idle. This works by adding a temporary output filter that checks that the last characters are equal to prompt.

There are false positives in some situations (such as invisible evaluation), so to be safe you may change the prompt before executing your R code and change it back once it's done, this will ensure everything works as it should. I would be happy if someone can suggest a better strategy to check for idleness.

I use this to execute pandoc when knitr has finished running and it works well.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -3374,6 +3374,65 @@ search path related variables."
           (ess--mark-search-list-as-changed)
           path)))))
 
+ ; Asynchronous hook
+
+(defvar ess-inferior-async-active-filters nil)
+
+(defsubst ess-inferior-async-generate-id ()
+  (intern (concat "ess-async-filter-" (number-to-string (abs (random))))))
+
+(defun ess-inferior-async (fun &optional proc)
+  "Execute FUN once PROC is ready by creating a one-off comint
+filter that will cleans after itself. FUN can be a symbol
+representing a function or a lambda."
+  (let* ((proc (or proc (ess-get-process ess-local-process-name)))
+         (proc-name (with-current-buffer (process-buffer proc)
+                      ess-local-process-name))
+         (id (ess-inferior-async-generate-id))
+         (hooked (ess-inferior-async-generate-filter id proc-name)))
+    (setq ess-inferior-async-active-filters
+          (push (list id fun) ess-inferior-async-active-filters))
+    (add-hook 'comint-output-filter-functions hooked)))
+
+(defun ess-inferior-async-generate-filter (id proc-name)
+  `(lambda (string)
+     (condition-case err
+         (progn
+           (update-ess-process-name-list)
+           (cond
+            ;; Process cannot be found, so clean up
+            ((not (assoc-string ,proc-name ess-process-name-list))
+             (ess-inferior-async-cleanup (quote ,id)))
+            ;; Comint process is an inferior ESS, run the filter
+            ((and (eq major-mode 'inferior-ess-mode)
+                  (string= ess-local-process-name ,proc-name))
+             (let ((ready (ess-inferior-ready-p string)))
+               (when ready
+                 (unwind-protect
+                     (ess-inferior-async-run (quote ,id))
+                   (ess-inferior-async-cleanup (quote ,id))))))))
+       (error
+        (message (concat "Error during async evaluation: " (cadr err)))
+        (ess-inferior-async-cleanup (quote ,id))))
+     string))
+
+(defun ess-inferior-async-run (id)
+  (let ((fun (cadr (assoc id ess-inferior-async-active-filters))))
+    (funcall fun)))
+
+;; False positives when invisible commands are issued.
+(defun ess-inferior-ready-p (string)
+  (let* ((prompt-length (length inferior-ess-primary-prompt))
+         (substring (when (>= (length string) prompt-length)
+                      (substring string (- prompt-length)))))
+    (string= substring inferior-ess-primary-prompt)))
+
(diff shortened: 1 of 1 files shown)


lisp/ess-inf.el
```diff
@@ -3374,6 +3374,65 @@ search path related variables."
...
+        (message (concat "Error during async evaluation: " (cadr err)))
+        (ess-inferior-async-cleanup (quote ,id))))
+     string))
+
+(defun ess-inferior-async-run (id)
+  (let ((fun (cadr (assoc id ess-inferior-async-active-filters))))
+    (funcall fun)))
+
+;; False positives when invisible commands are issued.
+(defun ess-inferior-ready-p (string)
```
lionel-: We should also use this in `(inferior-ess-set-status)` since comparing a substring should be faster than matching a regexp over the output.

vspinu:
Have you checked `ess-async-command` and `ess-async-command-delayed`?  I implemented those years ago, but they didn't work in foolproof fashion for  time consuming tasks. So I didn't use them for anything, but unless something broke in between they should be doing exactly what you need. 

The standard way to check for availability of the process is by `(process-get proc 'busy)` or `(ess-process-get 'busy) `.

I think hunting prompt in comint hook is a bad idea. If you need something do it in our filters directly. It's too much stuff going on and it's already quite layered. Tracking prompts in different places will only complicate the system. See `inferior-ess-tracebug-output-filter` which is our default filter.

This prompt traking bussiness will most likely fail if an assync command is fired when the process is already busy. Your idea of changing process prompt might be a way to go. We can have special prompt like [id]>> from which we pick up the callback id.

lionel-:
I use the prompt-change strategy for my pandoc hook and I didn't have any problem as of yet.

> Have you checked ess-async-command and ess-async-command-delayed?

Oh I didn't know about those. The `callback` and `interrupt-callback` are what I'm looking for right?


> The standard way to check for availability of the process is by (process-get proc 'busy) or (ess-process-get 'busy).

But how can we use this without blocking?

vspinu:
> Oh I didn't know about those. The callback and interrupt-callback are what I'm looking for right?

Yes, indeed. Those were originally intended for subtle stuff to execute behind user's back. Like for example update `default-directory` of the inferior to current working directory in R.  

> But how can we use this without blocking?

What do you mean? Those are queries of process variables. They don't block. `busy property is set and reset each time you see a prompt (inferior-ess-set-status) or execute a string (ess-send-string). That tracking system is as good as you can go without RPC protocols.

lionel-:
> What do you mean?

I'm not sure how to use these attributes to execute lisp code only when the inferior is ready, in a way that doesn't block like `ess-wait-for-process`. But I'll look into `ess-command-async`, thanks for the pointer.

vspinu:
> 'm not sure how to use these attributes to execute lisp code only when the inferior is ready, 

That you won't be able to know for sure. That's the biggest unsolved problem. Most likely if 'busy is nil, the process is ready, but that need not be under some conditions.

The best thing that I could think of so far is to have a short ping to the process (like (ess-send-string "\n")). If after that one the process is not busy, then we are sure it's idle. This can screw the user in some very limited corner cases, but I think it's worth the trouble. If we have such ping implemented, then we can use it in completion, eldoc and all other process comunication without stalling the process like we do right now.

lionel-:
How about generating a unique id for each callback. Then:

- We add `options(prompt = id)` at the end of the R forms.
- When we catch the id at the end of a comint output, we restore the prompt and execute the callback.

This seems robust to me but I may be missing something.

vspinu:
Yes. This is precisely what I had in mind as well but how do you reset the prompt at the end?
Wouldn't it be simpler to print some special string like it's currently done in injection code? It should be very short so that emacs process to unlikely to split it.

lionel-:
ok let's just send a simple string containing the ID then, catch it, filter it out of the output, and execute the callback.

vspinu:
The current implementation of async command work reasonably well. One inconvenience  is that they are commands, and in order to print the output to the user you would need to call `ess-eval-linewise` in that functions. That's the easy part. The more tricky part is with the `string` argument to the callback. Right now the string is only the last piece that the process feeds and is most commonly just ">". So it's useless for any output processing or error sniffing. 

I needed to figure this part for polymode ESS weavers and exporters. Have a look at how the command and callbacks are implemented in [polymode](https://github.com/vspinu/polymode/blob/master/modes/poly-R.el#L402-L432) right now.

The `pm--ess-run-command` is essentially `ess-async-command` with last line replaced by `ess-eval-linewise`. The tricky part is in `pm--ess-callback`. You basically have to hunt output `string` in 3 different places. 

So fixing current async system is pretty easy. First allow for `visual` evaluation. Secnd, when callback is running accumulate process output into a string and provide it to the callback function at the end. The whole thing could be enhanced by changing R prompt to include an ID as we discussed earlier. 
  
I would go with this. Dealing with comint hooks is ridden with perils and you won't be able to implement async commands because those are not even getting to comint filters. So let's stick with our filters, it's a small step to enhance what has been already done there.

lionel-:
yup I'll look into your approach next time I'm generating documents with pandoc. Thanks.

lionel-:
I might just try to reuse polymode infrastructure

vspinu:
what do you need besides of what is already there?  There is [pandoc exporter](https://github.com/vspinu/polymode/blob/master/polymode-export.el#L341) in polymode.

PR #298, opened 2016-03-05
Add getwd and setwd commands for Julia.

tpapp:
Trivial change, uses `cd()` and `pwd()` commands of Julia.

Also aligned `ess-julia-customize-alist` for better readability, put that in a separate commit.


vspinu:
Thank you!

PR #304, opened 2016-03-13
Add autoloads for R-mode and opening *.R files

### LINKED ISSUE #42: Automatically open .R files in R-mode
Currently, you have to insert the following into `.emacs` in order for `.R` files to open in R-mode:

    (autoload 'R-mode "ess-site.el" "" t)
    (add-to-list 'auto-mode-alist '("\\.R\\'" . R-mode))

Could we please register .R file extensions with R-mode as part of ess installation?

davidshepherd7:
With this you should just be able to install ess via `package.el` and it all "just work" when you open an R file.

Hopefully fixes #42 

I tested it by building the ess package with melpa and loading it, and it worked nicely.


vspinu:
Thanks!

PR #309, opened 2016-04-02
ESS clean up + ess-r-package (WIP)

lionel-:
@mmaechler @vspinu

ok this turned out to be a bit more massive than planned.

* This includes the new ess-r-package-mode

* In addition, this goes a long way of cleaning up ESS and making the core independent from the dialect stuff. I created lisp/old/ to put stuff that isn't relevant anymore. All utils are moved to ess-utils.el.

* This also extends the testing framework and now it's possible to easily write tests that interact with the R inferior. I've updated the Makefile in the tests directory. Just go there and type `make` to run all the tests, including indentation.

* The evaluation commands have been refactored so that no context-specific code is left in ess-inf.el. This mostly concerns DDE, and also SAS and Stata to a lesser extent. @rmheiberger @rsparapa, if I'm not mistaken you've been involved with those dialects. Could you test the branch please?

Martin, if no objection to making ESSR target R 2.0 or higher, I'll turn it to a namespaced package and upload it to this branch (this is mostly ready).


lisp/ess-custom.el
```diff
@@ -2108,27 +2113,26 @@ for help files.  The default value is nil for other systems."
 
 ;;;;; names for communication using MS-Windows 9x/NT ddeclient mechanism
 
-(defcustom inferior-ess-ddeclient nil
+(defcustom inferior-ess-ddeclient "Initial"
```
vspinu: I think we can drop support for DDE. I doubt anyone uses that anymore. No-one tests is anyways. Let's disucss this on ess-devel.


Why "Initial". Looks like a wart.

lisp/ess-custom.el
```diff
@@ -2321,41 +2325,88 @@ and `ess-tracebug'.")
...
-(defvar ess-load-command "source(\"%s\")\n"
+(defvar ess-load-command "source('%s')\n"
   "Dialect specific format-string for building the ess command to load a file.
 
 This format string should use %s to substitute a file name and should
 result in an ESS expression that will command the inferior ESS to load
 that file.")
-(define-obsolete-variable-alias 'inferior-ess-load-command 'ess-load-command "ESS v13.09")
-
-(defvar ess-load-visibly-command nil
```
vspinu: There was a good reason for all these commands. Removing these will surely break something.

lisp/ess-site.el
```diff
@@ -56,20 +56,18 @@
```
vspinu: A lot of ess-site.el were deleted. Where did it go?  I think people might get upset by this. ess-site.el was intended as an example of configuration file.

lisp/ess-sp4-d.el
```diff
@@ -34,11 +34,10 @@
```
vspinu: all *sp*.el files exept the most recent sp6 should go to /old.  I think sp6w is probably unnecessary as well. But who knows :(

lisp/old/ess-old-setup.el
```diff
@@ -0,0 +1,344 @@
+;;; ess-old-setup.el --- Old setup suggestions from ess-site.ely
+
```
vspinu: Ok. I see. ess-site was moved here. So maybe call it ess-old-site.el then?

vspinu:
This is all great. The only problem is that there are too many unrelated changes. It's difficult to get around . developer2, test infrastructure and ESS would have been better kept separated. If possible it would be good to cherry pick code re-organization out into a separate branch. Then merge test and developer2. Then, once we have tested developer2 go ahead into code re-organization. Is it already late for this?

There are many more old files files out there - *xlt*.el *vst*.el, *send*.el almost all *sp*.el,    r-gui.el, *omg*, ess-font-lock, *eldoc.el, *debug.el, *comp.el, *compat.el, *dde.el. These should be moved to old/ eventually.

lionel-:
> The only problem is that there are too many unrelated changes.

I know, I've been cleaning up as I went :/ I should have given links of the most relevant commits:


##### Config and utils clean-up

- ess-site.el https://github.com/emacs-ess/ESS/pull/309/commits/cde69977353ec74d18d3ff8a2db4d9b2ef6bb0a6
- ess-utils.el https://github.com/emacs-ess/ESS/pull/309/commits/ebcdfc185c91c99ee371d01f000acf6040fe209a
- ess.el https://github.com/emacs-ess/ESS/pull/309/commits/d040133302cdc244ce301326a3cfd30cafefb79e


##### Refactoring of evaluation functions

https://github.com/emacs-ess/ESS/pull/309/commits/d8f24ecd56c2a89ef51615ad14b17b85104f0c2b


##### New testing macros

https://github.com/emacs-ess/ESS/pull/309/commits/603f35f77e2039f5fea8c52a643347b70f77c4be


>  If possible it would be good to cherry pick code re-organization out into a separate branch.

It's possible to move the clean-up of ess.el, ess-site.el and ess-utils.el in another branch. It's mostly harmless to merge it in though, it shouldn't affect the end user. Do you prefer a separate branch so it shows up as a different merge in the git history?

The thing is that the most important reorganisation (the evaluation code) is tied to the new evaluation functions for the developer mode.


> There are many more old files files out there - xlt.el vst.el, send.el almost all sp.el, r-gui.el, omg, ess-font-lock, *eldoc.el, *debug.el, *comp.el, *compat.el, *dde.el. These should be moved to old/ eventually.

Yes, I think we should focus on finishing the cleanup for the next release. This would involve:

- Proper derived modes.

- Complete independence of ESS core from the derived modes ESS core would consist of ess-mode.el, ess-inf.el and ess-utils.el. Maybe we should split the latter in several files, like ess-utils.el, ess-syntax.el, ess-integration.el.

- Aiming for ELPA inclusion. I did not look into this yet but it may involve quite a bit of work if their guidelines are stringent (just updating the docstrings to Emacs standards would involve some work).

- I'd really like to take this opportunity to change the namespace of derived modes. For R we could have something like `r-` or `rstats-`.


#### About the package mode:

We now have two minor modes, `ess-r-special-evaluation-mode` and `ess-r-package-mode`. The former makes it easier to have a custom mode-line to show the namespace in which we are sending code. The modeline looks like this:

- Only namespaced eval in `foo` → `[src:foo]`
- Both package and namespaced-eval in `foo` → `[pkg:src:foo]`
- Package active for `bar` and namespaced-eval in `foo` → `[src:foo] [pkg:bar]`

I think `eval` would be better than `src` but it's probably best to keep those as short as possible.

Since you want to turn on namespaced-eval in packages by default, I think the behaviour should be as close as possible to regular evaluation. I made some work towards this:

- Now the only difference in terms of UI is that the evaluation messages are prefixed with `[foo]` when evaluating in a namespace.

- I removed the summary with the number of functions sourced and exported when sourcing a buffer, but on second thought it's not bad or confusing to have this. Maybe make it an option.

- I think sourcing only objects already existing in a namespace is a bit confusing. This should be an option like `ess-r-namespaced-load-only-existing` (not yet implemented).

lionel-:
>  I think we should focus on finishing the cleanup for the next release. 

I mean the release in 6 month ;)

lisp/ess-custom.el
lionel-: > Why "Initial". Looks like a wart.

yes, someone initialised that variable to "Initial" in ess-site.el... I should have changed the code to deal with nil rather than "Initial" but I think I was starting to give up at that point ;) I'll change it to nil.

> I think we can drop support for DDE. I doubt anyone uses that anymore.

It's needed for S-plus on Windows. I'm not against dropping support, but let's see if @rmheiberger manages to run the latest ESS with these environments.

> No-one tests is anyways.

Yes, we really need basic tests for all the inferiors officially supported, and some people on the team in charge of running them before each release. I think we should drop support for all the systems we are not able to test.

lisp/ess-custom.el
lionel-: There was not much point in multiplying the number of variables for each different parameter. There are now two possibilities: using a simple `ess-load/eval-command`, or writing a formatting function and binding it to `ess-format-load/eval-command-function` in the customise-alist. That function should be in charge of parameterising the command. The function has priority over the variable when both are defined.

lisp/ess-sp4-d.el
lionel-: I read on the mailing list that people still use old versions of S-plus because you have to pay to buy new ones. I have no idea if someone is still stuck with S-plus and uses ESS though. We're almost in 2020, at some point S-plus has got to die :)

lisp/old/ess-old-setup.el
lionel-: I think setup is more suggestive.

lisp/ess-custom.el
rmheiberger: i will look at this later in the week.  my prior is that dde should be retained as is, with no changes, so that S-Plus gui will continue to work on windows.
similarly ess-sp6w.el should be retained.

when i invented the term "Initial" i had areason.  please don't change it until i recall the situation.

mmaechler:
We ***REALLY*** need to be sure that DDE is not needed anymore.  
I had thought it still *was* crucial for Windows... but as I only rarely use Windows I'm not sure..
Rich @rmheiberger - can you help us here?

lionel-:
No worries Martin, I put a lot of effort into making sure it still works as before, all the while removing the dependence of ESS core on DDE.

There could be bugs though because I'm not in the position to test this.

lisp/ess-custom.el
rmheiberger: I am back home.

M-x S-initiate from ESS-5.12 works correctly on Windows 10 with S-Plus 8.2
GUI.
M-x S-initiate from ESS-15.03-1 doesn't work on Windows 10 with S-Plus 8.2
GUI.


The major difference between the two versions of ess-sp6w-d.el is the name
changes of variables to drop the 6.  My plan is to revert the older
8-Nov-2010 version of the file into the current ESS and then make it work.

Once it works again, I recommend that it be frozen.  No new wonderful
style features should be incorporated into ess-sp6w-d.el

I barely use S-Plus anymore, and I rarely use Windows except for
testing compatability--needed because much of my university is windows
based.
I personally switched to Macintosh four years ago.  As one of you
commented, S-Plus itself is old---but it still works.  Therefore ESS should
continue to still work with it.

The major response that I think ESS should have to TIBCO is to
automatically detect when TERR is installed and generate the function
M-x TERR.  This would be done at the same time that the various versions
of R are detected and the M-x R-* functions are generated.

The MS DDE technology is the communications technology used by M-x S  on
Windows.  It is a one-way
communication.  C-c C-n and related keystrokes send lines from the
*ESS* buffers to the independent S-Plus GUI process.  Nothing comes
back.

M-x Sqpe on Windows works in the ESS-standard manner of running a
command-line
interpreter in an *S* inferior buffer.  Sqpe on Windows is the S-Plus
equivalent
of Rterm on Windows.

lionel-: Thanks Richard. Could you please make those changes in the developer2 branch rather than on master, so that we don't have to do the work twice.

Is ess-rgui.el still relevant at all for using ESS on Windows? If it's possible to interact with the R process the normal way, I think we should drop support for RGUI interaction, especially considering the limitations you just exposed.

lisp/ess-custom.el
vspinu: > or writing a formatting function and binding it to ess-format-load/eval-command-function in the customise-alist. 

What is the point of this? Do you have a particular use case when those must be functions?

ESS always dealt with two types of evaluation - visual and non-visual. I think this is a well defined and non-controversial scheme. It might be unclean down there at constants level, but removing the distinction completely doesn't solve the problem. Quite the oposite, having only one evaluation format or function will likely to confuse things.

lisp/ess-custom.el
vspinu: > My plan is to revert the older 8-Nov-2010 version of the file into the current ESS and then make it work. 

That's too old. The 6-cleanup happened quite some years ago. If no-one complained for missing r-gui support since then, it probably means that no-one is using it. 

I don't see any point in keeping DDE support for R, R-s gui is rudimentary. Keeping it for S-plus doesn't make much sense either. There are fewer and fewer users of S-plus and those are unlikely to be ESS users anyways. ESS users are unlikely to use GUI either. So keeping it alive is a wasted effort IMO. 

That being said, if S-PLUS gui part could be isolated of the rest and sealed as you propose it would be good. We can put it in old/ box and forget about it.

lisp/old/ess-old-setup.el
vspinu: > I think setup is more suggestive.

Maybe but it's not conventional. ESS users are used to -site. Emacs and other *nix packages are using -site for site-wise configuration. Emacs packages are using `-config.el` for examples of configuration.

Let's just keep it ess-site-old.el. Such a name would quickly pop up in google search is old users would search for ess-site.

lisp/ess-custom.el
lionel-: > ESS always dealt with two types of evaluation - visual and non-visual.

Nothing has changed on that point. `visibly` gets treated at two levels: in the evaluation functions themselves (interaction with inferior), and within the inferior. We are only talking about the latter.

The R dialect had special command variables with different combinations of `visibly` and `noecho`. I needed to add `package` and `verbose` parameters for namespaced evaluation. Are we to multiply exponentially the number of evaluation commands? Both for eval and load? We need a function here.

The signature for the eval formatting function is

```lisp
(defun ess-format-eval-command (string &optional visibly output file &rest args))
```

and for load it's

```lisp
(defun ess-format-load-command (file &optional visibly output &rest args))
```

The `&rest args` is meant for dialect specific parameters. All other parameters are ESS-wide. I think that is a reasonable scheme.

Maybe we need to move `output` to `&rest args` since it's currently R specific though.

lisp/old/ess-old-setup.el
lionel-: Alright, I'll change it back.

lisp/ess-custom.el
lionel-: > if S-PLUS gui part could be isolated

Already done in this branch :)
We just need to check I didn't break anything.

vspinu:
>>   I think we should focus on finishing the cleanup for the next release.
> I mean the release in 6 month ;)

Ok. So we agree that the cleanup that you have started is in a wrong place. Let's put all non-developer2 changes on a separate cleanup branch. Tests could stay on developer2 as they are harmless. 

After the release we continue working on cleanup branch for a month or two and merge once we think it's ready. 

> Yes, I think we should focus on finishing the cleanup for the next release. This would involve:

I agree with all your proposals. Let's do all that on a `cleanup` branch. 

> Since you want to turn on namespaced-eval in packages by default, I think the behaviour should be as close as possible to regular evaluation. I made some work towards this:

Thanks. It's a right direction indeed.

>  removed the summary with the number of functions sourced and exported when sourcing a buffer, but on second thought it's not bad or confusing to have this.

I think it's a good default to have summary. Some things might not be source properly for whatever reason (R changes, devtools interference etc) and user must have a clear feedback that ESS is sending modified functions/methods into right environment. Let's keep this non-optional unless you have a strong reason to have an option for this part.

> I had thought it still was crucial for Window

It's only for communication with running R-gui. Why would someone ever want that?

I think that the approach that Lionel has taken is the right one. Isolate from the core, make a separate package and a DDE activation command, then move to old/. Interested users can get it directly from there, or we can autoload the package after users interactively activate DDE.

lionel-:
> Let's put all non-developer2 changes on a separate cleanup branch.

Ok. @rsparapa @rmheiberger, please don't work on developer2 until I have separated the branches, as the history will be rewritten.

lionel-:
> If we make a bunch of changes on master, then they will be orphaned.

This could lead to some merge/rebase conflicts to be resolved manually, but that's not a big deal. If you don't modify the evaluation functions in ess-inf.el and ess-r-d.el, it should be fine.

lisp/ess-custom.el
vspinu: I think i see the point now. So it's only an internal change. Fine with me then. It feels a bit awkward to have a generater for formater. 

Would it be tricky to use `format-spec` with a range of pre-specified formaters here? For example %f is subsituted for file, %v for TRUE if evaluation must be done visibly etc.  

If we end up reducing clutter, why not merge these two functions into one then? You already have `file` argument for `-eval-command`. We can make `string` optional; when nil. Means load `file`.

lisp/ess-custom.el
vspinu: > We just need to check I didn't break anything.

Aha great. @rmheiberger could you please make sure that you test/work on this branch or wait till it's merged? Simple reversion to 2010 will surely not work anymore.

rmheiberger: Lionel,

considering the set of emails following the one I am responding to, I think
I
will stay with my original plan and make dde for Sgui work correctly on my
local copy.

I am naive on git so I won't attempt to upload until later.
Please send me the line I need to download a clean copy of whatever branch
you want me to start with.  Once it works locally, then I will ask for the
next set of instructions to upload and merge it.

Rich

lionel-: let's put the cleanup part in another branch and merge developer2 into master, then we can all work on it from there. Any objection @vspinu ?

lisp/ess-custom.el
lionel-: > Would it be tricky to use format-spec with a range of pre-specified formaters here?

Yes it would work but it would be less flexible I think.

> If we end up reducing clutter, why not merge these two functions into one then? You already have file argument for -eval-command. We can make string optional; when nil. Means load file.

That could be a good idea, but would it make things less clear? How would we call this function?

lisp/ess-custom.el
vspinu: >> On Mon, Apr 04 2016 14:45, Lionel Henry wrote:

> let's put the cleanup part in another branch and merge developer2 into master,
> then we can all work on it from there. Any objection @vspinu ?

No. Please go for it.

lisp/ess-custom.el
vspinu: >> On Mon, Apr 04 2016 14:51, Lionel Henry wrote:

> Yes it would work but it would be less flexible I think.

Probably so, but it's easier for the user to modify. There might be that the
current format commands are already modified by some users.

Quite frankly I mind the most the excessive layers of indirection. Function ->
format string -> low level function (.ess.eval, .ess.source etc).

> That could be a good idea, but would it make things less clear? How would we
> call this function?

We use terms source, load and eval somewhat interchangeably. How about
*-eval-formatter?  The long name ess-format-eval-command is not very appealing
to me because 'format' reads as verb in there.


  Vitalie

lionel-: > Function -> format string -> low level function (.ess.eval, .ess.source etc).

No, the function returns a string ready for evaluation. The returned string does not need to be formatted.

> The long name ess-format-eval-command is not very appealing to me because 'format' reads as verb in there.

That's intended, it does format a function to send to the inferior based on a bunch of parameters.

> There might be that the current format commands are already modified by some users.

That could be a concern. Does anyone do that though? They can still write their own function.

> We use terms source, load and eval somewhat interchangeably.

I think the nomenclature is that load and eval are both about a form of sourcing. Load is about a file, eval about a command.

Where things get blurry is that we load files through eval for tramp performance... I was confused for a long time by this.

vspinu: >> On Mon, Apr 04 2016 16:09, Lionel Henry wrote:

>> The long name ess-format-eval-command is not very appealing to me because 'format' reads as verb in there.

> That's intended, it does format a function to send to the inferior based on a
> bunch of parameters.

I see. How about more meaningful ess-build-eval-command command then. In pure
functional tradition it would be sufficient to have ess-eval-command, but it's
too confusing.

>> We use terms source, load and eval somewhat interchangeably.
> Load is about a file, eval about a command.

Internally it comes down to the same thing - sourcing files.

I think it would be cleaner to deal with formatting issues in one "format"
command. Let's see what are other IDE's are calling it.

> Where things get blurry is that we load files through eval for tramp
> performance... I was confused for a long time by this.

I don't remember the details but this is happening only when the file is not on
a remote. It should not happen when file and process are on the same machine.

  Vitalie

lionel-: > How about more meaningful ess-build-eval-command command then.

That's a good name.

lionel-:
@rsparapa @rmheiberger 

The codebase for the next release is now on master.

rsparapa:
@rsparapa<https://github.com/rsparapa> @rmheiberger<https://github.com/rmheiberger>

The codebase for the next release is now on master.

I can confirm that the when-let bug is now gone and running SAS in batch and
interactively works.

mmaechler:
>>>>> Rodney Sparapani <notifications@github.com>
>>>>>     on Tue, 5 Apr 2016 06:59:31 -0700 writes:

    > @rsparapa<https://github.com/rsparapa> @rmheiberger<https://github.com/rmheiberger>

    > The codebase for the next release is now on master.

Indeed.... probably a good idea, so we really all (I hope!) work
with it.

However, for me  "ess-developer" is very broken:
All the keys I knew by heart no longer work (probably on purpose
    by someone ... I am still not convinced why old timers like me
    *must* re-train their spinal cord to new keys just because you
    find it cleaner)

But not even the Emacs ESS menu works to turn on developer
mode...

Please help!

Martin

PR #312, opened 2016-04-05
ESS clean up

lionel-:
The cleanup stuff is on the branch ESS/cleanup. Opening this PR for further discussions.

TODO proposal:
##### ess-mode
- [ ] Aiming for ELPA inclusion. I did not look into this yet but it may involve quite a bit of work if their guidelines are stringent (just updating the docstrings to Emacs standards would involve some work).
- [ ] Replace customize-alist by mode local bindings from CEDET's mode-local.el. This way changing a setting will work in all opened buffers. That will make ESS much easier to configure for the end user.
- [ ] Complete independence of ESS core from the derived modes. ESS core would consist of ess-mode.el, ess-inf.el and ess-utils.el. Maybe we should split the latter in several files, like ess-utils.el, ess-syntax.el, ess-integration.el.
- [ ] Maybe rename inferior-ess-mode to inf-ess-mode. This would be shorter and would be consistent with the Clojure inferior.
##### Derived modes
- [ ] Derived modes properly derived from ess-mode and inferior-ess-mode.
- [ ] Move defcustoms from ess-custom.el to their respective files.
- [ ] Take this opportunity to change the namespace of derived modes.
  
  For R we could have something like `r-` or `rstats-`. The S language could be `slang-mode` and `rstats-mode` and `splus-mode` would derive from it. Unless we move splus to lisp/old, in which case we can integrate all S stuff into the R mode.
  
  Stop using the confusing -d -l convention. Rename `ess-r-d.el` to `rstats-mode.el`; merge `ess-sas-d.el` and `ess-sas-l.el` into `sas-mode.el`, etc


lionel-:
@vspinu now rebased on master.

What's the best way to approach this? Revert the generics commits on master, rebase cleanup on master a second time, and then unrevert the commits? So that we keep a common history.

vspinu:
Ok. Let's not complicate matters and simply leave it like this. The only new feature I want to add is to change vignette UI a bit. This should not create conflicts.

PR #324, opened 2016-04-24
Support filling Roxygen blocks that contain backtick characters

nfdisco:
I've run the tests and it didn't seem to break anything.

### DIFF
--- a/lisp/ess-roxy.el
+++ b/lisp/ess-roxy.el
@@ -67,6 +67,7 @@
 (eval-when-compile
   (require 'cl))
 (autoload 'Rd-preview-help "ess-rd" "[autoload]" t)
+(require 'essddr "ess-rd.el")
 
 ;; ------------------
 (defvar ess-roxy-mode-map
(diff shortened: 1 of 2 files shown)


lisp/ess-roxy.el
```diff
@@ -731,19 +732,22 @@ list of strings."
...
-         (temp-table (make-syntax-table S-syntax-table))
          (paragraph-start (concat "\\(" ess-roxy-re "\\(" paragraph-start
-                                  "\\|[ \t]*@" "\\)" "\\)\\|\\(" paragraph-start "\\)")))
-     ;; Prevent the roxy prefix to be interpreted as comment or string
-     ;; starter
-     (modify-syntax-entry ?# "w" temp-table)
-     (modify-syntax-entry ?' "w" temp-table)
+                                  "\\|[ \t]*@" "\\)" "\\)\\|\\(" paragraph-start "\\)"))
+         (temp-table (if ,examples
+			 (make-syntax-table S-syntax-table)
```
lionel-: Could you use spaces instead of tabs please.

test/literate/roxy.R
```diff
@@ -74,3 +74,24 @@ NULL
 
 ##'
 ##' ¶
+
+
+### 3 ----------------------------------------------------------------
```
lionel-: This test case could be a bit more minimal, i.e. just

```r
##' ¶Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor `incididunt' ut labore et ``dolore'' magna aliqua.
```

Without any other field so it's clearer we are testing for quotes (and we probably should come up with a way of labelling the test cases).

lionel-:
Otherwise looks good!

nfdisco:
Done!

Apparently the test needs to be terminated with a NULL otherwise it fails. In a buffer it works with or without the NULL.

lionel-:
Thanks!

Issue #233, opened 2015-09-10
R support for lambda.r function declarations


vixr:
R functions can be defined in an alternative way using lambda.r

foo(a, b) %as% {
...
}

Sometimes also with a %when% clause.

There are several advantages to lambda.r including type signatures and pattern matching. It would be useful for ess to recognize these patterns as R functions.

eddelbuettel:
:-1: 

Not a fan of pipes and even less so of other random extensions. The ESS developers work _really hard_ to make the existing language and Emacs cooperate peacefully and productively.  I see the above as a distraction.  But then I don't contribute code here so my vote doesn't really count ...

vixr:
I'm curious what the issue with pipes is, they seem like a flexible way of expressing otherwise messy data transformations.

At any rate, I don't see how pipes are relevant - lambda.r is meant to support some semblance of type checking, pattern matching, and other functional idioms. 

I'll admit that the lambda.r package may not that mainstream, so not supporting language extensions is a legitimate critique. On the lambda.r tokens such as %as%, %::%, %when% seem fairly unique.

eddelbuettel:
Fair point re tokens.  My view on pipes is somewhat personal, but their admiration is not quite as widespread as some corners of the the interwebs may suggest.

mmaechler:
I also strongly prefer functional syntax.   And our indentation defaults should primarily do the correct things for functional programming style.
Yes, pipes can be nice for "displaying" a data analysis.  But when you develop the analysis, you shold work with intermediate results, plot them, str() them, etc.... so the pipe is not even natural there.
I see this as a fashion... Of course it will remain.. but I don't believe it will become very wide spread in the long run.   Data analysis is very interactive, very much about looking at intermediate results, and not about one direct long (piped) process from beginning to end.

vspinu:
I have looked at lambda.r and it's interesting idea but it will unlikely ever become mainstream. The complication with yet another layer of types developed mostly for the pattern matching will just scary most of users away (as if we don't already have enough class systems in R).  I see `lambda.r` as an experimental/didactical package.

That being said, we should defenitely support block constructs like `(...) %blah% { ...}`. It's also the case for `foreach` package which has `(...) %dopar$ { ..}` constructs among similar others.

lionel-:
> Data analysis is very interactive, very much about looking at intermediate results, and not about one direct long (piped) process from beginning to end.

[...]

PR #329, opened 2016-05-03
Lexer and parser functions (WIP)

lionel-:
Putting the syntax stuff I've been working on here until the next release is out. Could be a while before I finish this.

- Proper lexer
- Pratt parsing, forward and backward

This will be instrumental in providing syntactically valid sexp motions and sophisticated block and defun navigation.

In addition, some improvements to fontlock:

- Backquoted names have now their own face, `ess-backquoted-face`. They don't look like strings by default.
- There is a new `ess-operator-face` from which the existing `ess-%op%-face` now derives.
- Strings and backquoted names are now consistently recognised as function calls or function definitions.

Would a font face for function parameters be useful? I.e. `param` in `funcall(param = arg)`. I think it probably wouldn't be very helpful, so not worth the trouble. We could also distinguish between `=` as variable assignment and `=` as parameter assignment.

fontlock


lionel-:
Hi @vspinu,

This branch treats all pair-delimited symbols (`%op%`, `` `name` ``) as strings. This works fairly well for R-mode buffers, but breaks badly in the inferior because it's not rare to get percentage reports in outputs. This causes fontlocks to detect operators where there is none.

I had similar problem for years with strings. Fontlock regularly gets confused with the string boundaries while printing long streams of character vectors. It probably receives output by chunk and there is no guarantee that the opening string delimiter is part of that chunk. So I think it's time to fix that as well.

There are several things we could do.

- In any case, I think we need to treat `%` as `"."` syntax class in inferior buffers.

- As for `'` and `"`, we could either treat them as `"."`, or continue to have them as string delimiters but define an extend-region function to make sure the fontified we have all relevant delimiters in the fontified chunk. I'm not sure if that's even possible though. And that could slow down the display of large outputs.

- In addition, I'd like all pair characters to behave normally for *input*. So we need to have two different fontification schemes for output and input.

So my questions: do you know of a reliable way to execute some lisp code before output starts pouring, and again when we get back to a prompt? That code would switch syntax tables so we get proper fontification.

Do you think it's reasonable to stop treating strings as strings in output chunks?

vspinu:
> This branch treats all pair-delimited symbols (%op%, `name`) as strings. 

Why do you do this btw? In what sense they are treated as strings? In the OP you said that backquoted names don't look like strings anymore.

> Fontlock regularly gets confused with the string boundaries while printing long streams of character vectors.

I think this could in principle be fixed but it will probably impact output speed too much. Currently there are intentional hacks not to extend region too much.  I would not bother with the inferior output too much. People are used to scrweups on long output and the correct way is not to print that long ouptut anyways. 

> So my questions: do you know of a reliable way to execute some lisp code before output starts pouring, and again when we get back to a prompt? That code would switch syntax tables so we get proper fontification.

This sounds like a huge complication to me. I think it's just better to disable `%` or replace it with a regexp fontification in the inferior. But again, what was wrong with treating `%` as punctuation and regexping for `%op%` as we did before?

lionel-:
> Why do you do this btw?

I've found that's the most robust and fastest way to handle this problem. I.e. nested stuff like `%"'hop'"%` works out automatically and reliably. You can forward-sexp in front of it etc.

> In what sense they are treated as strings?

Only in the syntactic sense of being text enclosed by paired delimiters. They don't look like strings.


> This sounds like a huge complication to me.

It would be pretty straightforward if only we could execute code before and after sending output. I think that would be useful in general to do this, so I'm asking if that's possible.

Note that we can just use a syntax table with a regular class for `%` in inferior buffers and have output display just as before, while enjoying the robustness of correctly paired symbols in the script buffers. I'm just trying to make things better and I'd be quite happy if there were some hooks I could use to that end :)

vspinu:
> Only in the syntactic sense of being text enclosed by paired delimiters. 

I guess this means that `parse-partial-sexp` and `font-lock-fontify-syntactically-region` treat them as strings.

> It would be pretty straightforward if only we could execute code before and after sending output. I think that would be useful in general to do this, so I'm asking if that's possible.

What do you have in mind more concretely? Do you plan to put syntax tables on different regions in inferior? What happens if previous output, or previous input or boundary between regions got edited?

I have no idea how heavy those per region syntax tables are, or how they could impact the rendering speed, but we should be extra careful not to clutter the inferior because speed is an important factor there.

> Note that we can just use a syntax table with a regular class for % in inferior buffers and have output display just as before

I would stick with this for now. Let's see what is the benefit if the new system in script buffers and if it's worth the fire we will port it to inferior as well.

lionel-:
> Do you plan to put syntax tables on different regions in inferior?

Nope, just to switch the active syntax table in the buffer, so there shouldn't be any speed or memory issue.

I think we should disable fontification of all unreliable elements in output, at least by default. You're right that users know that output fontification is janky, but that's not a good thing. I'm aiming for complete predictability of fontification in ESS R.

But now that you mention editing, you're right that there could be complications, i.e. as soon as you start modifying the output buffer all those percentages will be treated as operator delimiters again.

I see in the documentation that we can add a `syntax-table` property to text. But that means those properties would be carried in when you copy the output to a regular R buffer, that'd be bad.

I also see that you can run hooks when point leaves or enters a region with the text properties `point-entered` and `point-left`. I think having those on the comint prompt could work. That'd allow us to switch between the output syntax-table and the regular syntax-table each time we leave or get back to the prompt. I'll try that and see if that works reliably. Hmm... but again that could lead to some weird behaviour when old inputs are edited... Maybe we'd need to remove those point-motion properties with another property recording modification hook, which seems like a lot of headaches.

You're probably right it's not worth the effort ;)

lionel-:
>  but again that could lead to some weird behaviour when old inputs are edited

Maybe this ties up into the issue of removing all motion constraining properties in old prompts.

lionel-:
The more I think about it, the more I feel like like the input region after the prompt  should be enclosed in a singleton overlay. This way we can have a hard prompt boundary with regular fontification of input, while keeping complete output editability and output-specific fontification,

lionel-:
Other things I'd like to stop fontifying in outputs: keywords like `in`, `if` etc. I find them distracting in reports such as from `devtools::check()`, or in `warning()` or `message()` text.

At the same time we'd like to have evaluation output (of functions for instance) fontified properly with all keywords and pair-delimited expressions etc. MPI could be useful there.

vspinu:
BTW, we have our own font-lock-fontify-region function. It might be possible to use it to fontify imput region differently than the output. The problem is to identify that we are in the input block and I don't see how this could be done reliably.

lionel-:
other instances of conditional fontification:

- usage and examples fields in the inbuilt help.
- examples field in roxygen comments

vspinu:
Hi Lionel, what is the status of this? I would like to attempt derived modes rewrite. Would be nice if this could be merged first. Then I would like to switch onto ess-cleanup branch.

lionel-:
I agree this should not linger for too long. I may spend a few days on it next week, I'm just afraid it will be a time hole as I can't really afford it.

On the other hand this can probably just be merged in as is, since the unstable part (the parser) is just unused code right now. So if you like we can merge it right now (but let me check I don't have local commits to push first).

vspinu:
> So if you like we can merge it right now (but let me check I don't have local commits to push first).

Please don't work on it before you finish with the thesis. Such things bring extra stress and unhappiness. So if you could just spend 5-10 minutes and check if evaluation works as expected and merge it in, that would be great.

lionel-:
here you go @vspinu

lisp/ess-r-d.el
```diff
@@ -190,16 +190,57 @@
...
+
+;; Prevent string delimiting characters from messing up output in the
+;; inferior buffer
+(setq inferior-ess-r-syntax-table (copy-syntax-table ess-r-syntax-table))
+(modify-syntax-entry ?\' "." inferior-ess-r-syntax-table)
+(modify-syntax-entry ?\" "." inferior-ess-r-syntax-table)
+(modify-syntax-entry ?` "." inferior-ess-r-syntax-table)
+(modify-syntax-entry ?% "." inferior-ess-r-syntax-table)
+
+(defun ess-r-font-lock-syntactic-face-function (state)
```
vspinu: @lionel-, I am extracting indentation and font-lock pieces into a stand alone r-mode, and I wonder if this piece is strictly necessary and how it actually works. 

First of all, can we simply include backquote into our function regexp and remove the `font-lock-function-name-face` from here? Secondly, is does this actually work as expected? Both backquote and `%` are punctuation characters in ess-r-mode and if  I understand correctly this function will never be called with `%op%` and backquoted names.

vspinu: Ok. For first question, I realized that you wanted to be able to accommodate `"foo" <- function(...)` which is a valid R code. Then it should be here indeed.

vspinu: Ah. Also nevermind the second part. I was looking at wrong place for the syntax definition. Both backquote and `%` are string delimiters in r-mode but not in the inferior.

lionel-: > I am extracting indentation and font-lock pieces into a stand alone r-mode,

Great!

I don't remember exactly as I wrote this two months ago. I think this is a corollary of treating %op% and backquoted names as syntactic strings. In general I don't like dealing with paired identifiers with regexes and I think this function is more maintainable and extensible. We now have a more accurate and more general fontification of function calls and defuns (see screenshot attached in the OP).

lisp/ess-r-d.el
```diff
@@ -190,16 +190,57 @@
...
+  (let ((string-end (save-excursion
+                      (and (nth 3 state)
+                           (ess-goto-char (nth 8 state))
+                           (ess-forward-sexp)
```
vspinu: Why is this `ess-forward-sexp` needed here?

lisp/ess-r-d.el
```diff
@@ -190,16 +190,57 @@
...
+                           (point)))))
+    (when (eq (nth 3 state) ?`)
+      (put-text-property (nth 8 state) string-end 'ess-r-backquoted t))
+    (cond
+     ((eq (nth 3 state) ?%)
+      'ess-%op%-face)
+     ((save-excursion
+        (and (ess-goto-char string-end)
+             (ess-looking-at "<-")
```
lionel-: I would agree but the problem is to handle these cases:

```r
`fun` <- # comment
  function() {}
```

lisp/ess-r-d.el
lionel-: If I'm not mistaken the forward-sexp here triggers a full rescan of the buffer by syntax.c functions so it would be better to avoid it. If the closing delimiter is saved in the syntax-ppss cache it would be better to use it. I don't think it is?

Another possiblity is to look for the corresponding character manually but then you need to deal with escaped characters etc, so it's easier to forward-sexp.

lisp/ess-r-d.el
vspinu: >  would agree but the problem is to handle these cases:

Ok. I see. Let's try to add such explanation in the code as we go. It's often hard to see the motivation behind those (many) small functions that you have in `r-syntax.el`.

lisp/ess-r-d.el
lionel-: yes so if you want to find the corresponding closing delim manually you need to handle escaped ones: `"\""`

vspinu: Hm. I would expect parse-partial-sexp to handle that automaticaly and return correct region automatically. I will have a look at the later stage. For now leaving everything as it is.

lisp/ess-r-d.el
lionel-: Agreed. There's a bug by the way, it should be `(ess-looking-at "function\\b" t)))` to handle the comment-in-the-way case. The second parameter allows to ignore newlines and comments in the lookup.

lionel-: Just pushed a fix on master.

vspinu: >  to handle the comment-in-the-way case. 

I am not sure this is worth the flame. Font lock keywords for function definitions is basically `"\\(\\sw+\\)[ \t]*\\(<-\\)[ \t\n]*function\\b"` and making it work with comments is non-trivial. So for feature consistency I would stick to simple regexp here as well.

lionel-: that's a fair point.

But if we ever want to support custom defuns (see issue #233 about the `%as%` operator for example) we'll need an elisp matcher. That's not very complicated to implement and then we can handle comments and provide a hook for custom defuns as well.

PR #330, opened 2016-05-05
Differentiate checking from testing

fmichonneau:
Minor modification, use message "Checking" instead of "Testing" when using devtools::check().

Thanks for adding this devtools integration!


lionel-:
Thanks!

PR #361, opened 2016-08-25
Fix typo in ess-describe-at-point-method docstring

dellison:
Change "Whehter" to "Whether" in the docstring for ess-describe-at-point-method.


lionel-:
Thanks!

dellison:
My pleasure! :)

PR #373, opened 2016-09-27
Passing data via MPI

lionel-:
@vspinu could you check the approach here?

This adds an MPI handler to pass data from R, and `ess-data-command` to retrieve it.

I use these facilities to prompt for a devtools command with completion. The completion set is dynamically populated from all devtools functions taking `pkg` as argument.

### DIFF
--- a/etc/ESSR/R/mpi.R
+++ b/etc/ESSR/R/mpi.R
@@ -17,4 +17,7 @@
     .ess_mpi_send("eval", expr, callback)
 }
 
+.ess_mpi_error <- function(msg, ...) {
+    .ess_mpi_send("error", sprintf(msg, ...))
+}
 
--- a/etc/ESSR/R/mpi.R
+++ b/etc/ESSR/R/mpi.R
@@ -21,3 +21,11 @@
     .ess_mpi_send("error", sprintf(msg, ...))
 }
 
+.ess_mpi_data <- function(data) {
+    if (!is.character(data)) {
+        .ess_mpi_error("MPI expected character vector")
+    } else {
+        data <- paste(data, collapse = "")
+        .ess_mpi_send("data", data)
+    }
+}
--- a/etc/ESSR/R/pkg.R
+++ b/etc/ESSR/R/pkg.R
@@ -1,0 +1,25 @@
+
+.ess_keep <- function(.x, .f, ...) {
+  is_true <- vapply(.x, .f, logical(1), ...)
+  .x[is_true]
+}
+
+.ess_devtools_functions <- function() {
+  if (!requireNamespace("devtools")) {
+    .ess_mpi_error("devtools is not installed")
+  }
+  devtools_env <- asNamespace("devtools")
+  exports <- getNamespaceExports("devtools")
+  funs_exported <- as.list(devtools_env)[exports]
+
+  is_first_arg <- function(f, arg) {
+    args <- names(formals(f))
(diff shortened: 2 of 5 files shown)


vspinu:
I am not sure I understand the benefit of this. The main point of MPI is to work asynchronously. What your code seem to be doing is replacing synchronous `ess-get-words-from-vector` with a much more complicated code.  Due to your non-local exit I very much doubt that it works correctly for multiple data responses (I guess this was your original motivation to use MPI for this).

Every entry in `ess-mpi-alist` is a callback. So if you want to use mpi you would need to include the whole `ess-r-devtools-ask` into that `data` callback. But I don't see much point in it. Cannot you just use `ess-get-words-from-vector` in your `ess-r-devtools-ask`? 

Also`data` handler is too generic. What does it do? Transfer data? All mpi handlers transfer some kind of data.

lionel-:
It just seemed a much cleaner approach to use your MPI delimiters than parsing a regexp for the printout of a character vector. The MPI handler works with functions that return data invisibly or that have a print method for example. Furthermore I think it should be possible to make it serialise R data to lisp data in a more general way in the future. E.g. serialising lists of lists of character vectors.

It does works with multiple data transfer, the non-local exit is contained within the handler loop.

I agree that it can only work synchronously but I'm not sure this means we can't use asynchronous facilities to implement this (by blocking).

lionel-:
> with a much more complicated code

this is hurtful, I thought it was rather simple and elegant :)

vspinu:
> It just seemed a much cleaner approach to use your MPI delimiters than parsing a regexp for the printout of a character vector. 

It's exactly same thing. Those delimiters are also searched by a regexp. 

This MPI system is simplified and add hock to allow some basic callbacks and I think it should be better used for that purpose only. 

FWIW, I have a working implementation of a generic mpi system that would allow arbitrary transport protocols and encoders. Haven't made it public as yet (this start of the academic year turned to be very busy). So in this regard I don't really care much about this simple ess MPI. So if you really think this data transfer is that useful, feel free to get it in. 

> this is hurtful, I thought it was rather simple and elegant :)

Yeh. Sorry about that. I just think this synchronous alteration obscures the purpose of MPI. IMO MPI should never be synchronous.

lionel-:
> It's exactly same thing. Those delimiters are also searched by a regexp.

It's not the same thing because it is structured. There is no robust way to parse lists, numbers, strings and symbols by parsing R output, but it is possible with proper handlers.

> FWIW, I have a working implementation of a generic mpi system that would allow arbitrary transport protocols and encoders. Haven't made it public as yet (this start of the academic year turned to be very busy). So in this regard I don't really care much about this simple ess MPI. So if you really think this data transfer is that useful, feel free to get it in.

Exciting news! Well, if you have something better in the works, I can just use `ess-get-words-from-vector` for the time being.

vspinu:
> It's not the same thing because it is structured. 

True. You can as well build your elisp data structure in R, print and read-back with `read`. This is currently done in `ess-function-arguments`. 

In my experience MPIs that enforce structured data protocols are often overengineered and lack flexibility. For emacs, in  99.99% of the cases, dict and lisp structures will be enough, an dict is a plist which is also list.  If a binary data transfer is needed, you can always pass it as a string and decode on the other end with the handler.

> Well, if you have something better in the works,

I am still struggling with a range of design choices. Will need to put all that in a white-paper form and then expose to relevant people. Will ping you up once it's done.

PR #390, opened 2016-12-25
Use company-complete in ess-indent-or-complete if available and enabled

izahn:
ESS has mechanisms for using tab to complete, but these mechanisms don't seem to have been updated to use company for completions if available. This uses company completions in `ess-indent-or-complete` if available and enabled via the `ess-use-company` variable.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -22,24 +22,6 @@ from capturing the cursor.
 @item @ESS{[R]}: New command @code{ess-r-package-use-dir}. It sets the
 working directory of the current process to the current package directory.
 
-@item @ESS{[R]}: Launching R from a package file now recognises the package
-root as the likely startup directory for R. If
-@code{ess-ask-for-ess-directory} is non-nil, ESS prompts with the
-package root as default selection. Otherwise the working directory is
-automatically set to the package root.
-
-An exception is made when you are opening a new process from the
-@code{tests} folder. Since @code{R CMD check} sets the working directory
-to that folder when executing test files, this folder is preferred over
-the package root.
-
-The mechanism for selecting a directory is a bit more general if you are
-using Emacs 25.1 or newer. On those recent Emacsen R packages are now
-recognised as Emacs projects and ESS queries Emacs projects to get the
-default startup directory. This means that any git or subversion
-directory will be recognised as project root even if it is not an R
-package (and in fact regardless of language or dialect).
-
 @item @ESS{[R]} Lookup for references in inferior buffers has been
 improved. New variable @code{ess-r-package-source-roots} contains
 package sub-directories which are searched recursively during the file
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -3098,6 +3098,9 @@ Used to store the values for passing on to newly created buffers.")
 
 (make-variable-buffer-local 'ess-listing-minor-mode)
 
+(defvar ess--enable-experimental-projects nil
+  "Enable experimental project support in ESS")
+
 (provide 'ess-custom)
 
  ; Local variables section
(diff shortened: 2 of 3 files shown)


lionel-:
I don't know a lot about completion in Emacs but I think you're supposed to use the CAPF backend to use company with `completion-at-point`. That's a better way than adding special branching for company in all Emacs packages.

izahn:
Yeah, I may not have considered all the complexities here. Basically I want to `tab` to indent or complete using company. https://github.com/company-mode/company-mode/issues/94#issuecomment-40884387 shows how to do it generally, except that it doesn't work with ESS because ESS does it's own `indent-or-complete`.

On the other hand, if a user says they want tab to complete (by setting `ess-tab-complete-in-script`), why not give them the best completion available? Is there any advantage to providing `capf` instead of `company` in this case? Company has `capf` backend, so if nothing matches first you'll end up getting `capf` completion anyway. At least I think that's the case...

lionel-:
Maybe `ess-indent-or-complete` should go away and be replaced with `indent-for-tab-command `? Does the former have any useful feature compared to the latter?

izahn:
Comparing https://github.com/izahn/ESS/blob/0d27fecc9936d21cd2e984de9c48f9b7f7c9f07f/lisp/ess-mode.el#L909 with https://github.com/emacs-mirror/emacs/blob/master/lisp/indent.el#L105 suggests that the main difference is that `ess-indent-or-complete` can be configured to bail out of completion and indent instead based on the text following point. Not sure if that is useful enough to justify keeping the `ess-indent-or-complete`.

vspinu:
This patch cannot actually be. company is not part of emacs proper.

izahn:
@vspinu can you clarify what this PR does that is not allowed? Why does it matter that company is not part of emacs?

vspinu:
Because ESS doesn't require company and some people might not even have it installed. Some folks might even prefer the default completion mechanism without the popup.

izahn:
Right, the idea here is to use `company` only if available (`featurep 'company`) and requested by the user (`ess-use-company`). It should work exactly as before if company is not available, or if the user opts out via `ess-use-company`. So why can't this be?

lionel-:
In R terms, you're offering a `Suggests:` feature, not an `Imports:`. It could make sense but I think in this case the better way is to use Emacs builtin command. A lot of stuff was added to ESS over the years, at times when Emacs was not as featureful.

vspinu:
Ah, that could work indeed.  I think this is an uniform gain but `ess-tab-complete-in-script` is nil by default, so my guess is that not many users will benefit from this. 

My personal setup is to make `tab` complete within  `company-active-map`:

```el
(define-key company-active-map [tab] 'company-complete-selection)
(define-key company-active-map (kbd "TAB") 'company-complete-selection)
```

This way I can indent with TAB in all modes and complete with TAB once company is up.

AFAICS the only difference between the above and proposed change to ESS is the small delay till company pops up, but that's a negligible difference for a generic solution.

izahn:
My personal set up is to disable auto-starting the completion and start it manually. This works great, except that I can't find a key-binding that makes me happy. I really want to initiate completion with tab, the way it works in many other environments.

I've got this working everywhere in emacs that I care about using https://github.com/company-mode/company-mode/issues/94#issuecomment-40884387

Now I just need to figure out how to make ESS work the same. Maybe this PR is not the best way, though I think it has advantages (e.g., consistency; if ESS is using company for automatic completion why should we use something else if completion is initiated manually instead of automatically?).

izahn:
Any other thoughts on this? I explored some other options without much success. @lionel- what did you mean by 
> the better way is to use Emacs builtin command

?
Do you mean replacing `ess-indent-or-complete`  with `indent-for-tab-command`? I'm certainly willing to look into that if you think it's a good idea.

lionel-:
> Do you mean replacing ess-indent-or-complete with indent-for-tab-command? I'm certainly willing to look into that if you think it's a good idea.

Yes for duplicated features I think we need to remove old ESS commands in favour of new Emacs ones.

PR #391, opened 2017-01-01
Add functions for pipes

jabranham:
I oftentimes run into the situation where I have a pipe:

```R
library(tidyverse)
mtcars %>%
  group_by(cyl) %>% 
  summarize(mean = mean(mpg),
|           n = n()) %>%
    filter(cyl > 4)
```

where `|` represents point and I only want to evaluate the pipe through that line. C-RET only evals that line and C-c C-c evals the whole paragraph. This is a function I came up with to help out with that situation. Hopefully someone else finds it useful. 

I also bound it to C-|

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1960,6 +1960,33 @@ for `ess-eval-region'."
         (forward-line 1)
       (ess-next-code-line 1))))
 
+(defun ess-beginning-of-pipe-or-end-of-line ()
+  "Find point position of end of line or beginning of pipe %>%"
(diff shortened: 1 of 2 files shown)


lionel-:
I think your patch is not general and not robust enough. For generality, it should work with any operators, e.g. it'd also be useful for ggplot2 pipelines. For robustness, it should use syntactic logic, e.g. something like `ess-climb-continuations`. However that function assumes point is in front of an operator so you need some logic for moving point at the right place first. I've been meaning to solve that problem reliably but haven't found the time yet.

PR #399, opened 2017-02-11
Provide feature ess-rd in library ess-rd.el

tarsius:
Previously ess-rd.el only provided the mismatched essddr.  To play it
safe this only adds the ess-rd feature, without removing the essddr
feature.


lionel-:
hmm, seems to come from a file renaming from way back. Thanks!

lionel-:
and thanks for your work on magit. It's a godsend.

tarsius:
You are welcome!

PR #400, opened 2017-02-11
Drop old compatibility

lionel-:
This is from Stefan Monnier


lionel-:
Just went over it quickly, this should not affect users with modern Emacses.

@vspinu after merging this, we should probably merge the commits in #312 to prevent branches from diverging too much.

vspinu:
@lionel- thanks for doing this. I am getting gradually disconnected from the ESS internals :(

vspinu:
I have looked briefly into the changes. All are straightforward indeed; please merge if it works.  Thanks again! 

As to #312, would be nice to have that merged as well, but I am afraid I won't be able to help much these days.

PR #404, opened 2017-02-19
Cleanup ess-site

lionel-:
This is a rebase of the commits in #312. This is mostly a reorganisation of ess-site:

- old stuff has been moved to a `./old/` directory.
- relevant stuff has been moved to ess-utils.el.

A concrete effect is that it should now be possible to load dialects independently (if not, that is a bug). E.g. you can load the R mode with `(require 'ess-r-d)` without bringing in unwanted modules (SAS, julia, etc).

@mmaechler @vspinu we should merge this as soon as possible to avoid branches from diverging and the painful merge resolutions it entails.


lionel-:
now that people can include module directly, I think we should rename files to `ess-r-mode.el` etc, at least for the modules that we know work independently. The common stuff would go to a `-lang` file, e.g. we'd have `ess-s-mode.el` and `ess-s-lang.el` (though the S stuff is increasingly irrelevant).

mmaechler:
"uncapitalization": Is this really needed because of emacs coding standards?

I really would very very much prefer `foo-R-bar` to `foo-r-bar` .  The language is "R", not "r".

lionel-:
We've already been converging toward uncapitalised R for a while. The language is "R" but capitalised letters make the code ugly (though that's of course a matter of opinion, but it's highly unconventional in Emacs code) and hard to read (noisy type color). It makes sense for things related to the R process though, so we call `R` and not `r`.

vspinu:
Yeh. This is one of those issues that don't quite have a good solution. AFAIC we should stick to uniform naming throughout, either r or R.  I don't quite see `ess-R-***` as being ugly but maybe that's because I am so used to R being R. To me `ess-r-***` is rather strange, but not that a big deal anyways.

lionel-:
>  To me ess-r-*** is rather strange,

To me it looks like an elisp namespace :)

Imagine if we had `ESS-R-foo-bar` everywhere...

lionel-:
Could we merge this? About the capitalisation issue, most of the code uses `ess-r-` already, so the last commit just makes `ess-r-d.el` (notice the small r btw ;)) consistent for a few remaining `ess-R-` variables. If we should ever switch to the latter style, this would need to be discussed and implemented in a separate PR.

lionel-:
One advantage of uniformising variable names in that commit is to enable packages like `nameless` to work consistently with the ESS codebase: https://github.com/Malabarba/Nameless

vspinu:
I haven't had a look but something tells me that I won't be able to give it a careful review anyways. Does it work smoothly for you? Do you have stata on your machine? If not could you make a simlink named stata to R and see if basic stata interaction works?

lionel-:
Interesting. Can R serves as a surrogate inferior for all languages? that could be useful for ensuring a minimum amount of unit testing on other environments than R.

I do have a version of Stata lying around on my computer, though I haven't run it for like 4 years. I'll look into this.

lionel-:
@vspinu @mmaechler Everything seems to work well. `ess-sta-d` can now be required independently as well. `ess-mode` now loads `ess-inf` and all ESS core files, so dialect files just need to require it to get most of what they need.

I think we should start renaming files that we know load independently to `ess-r-mode`, `ess-stata-mode`, etc. What do you think?

lionel-:
Maybe we should even rename them to `r-mode.el`, `r-syntax.el`, `stata-mode.el` etc, so it's easier to spot the files revelant to each dialect. That'd also be consistent with the command names for loading the modes (e.g. `stata-mode`).

vspinu:
> I think we should start renaming files that we know load independently to ess-r-mode, ess-stata-mode, etc. What do you think?

Yes. This sounds reasonable. 

> Maybe we should even rename them to r-mode.el, r-syntax.el, stata-mode.el etc

Let's not mix with that. There will be a standalone r-mode.el some day wich will lead to confusion.

lionel-:
On the contrary it's quite straightforward with git. Commit history is completely transparent to file renaming. Maybe you're thinking about SVN repos?

vspinu:
Yes. With git it should work as long as git detects that files were renamed. That should be explicitly checked. I don't have much experience with that but it happened to me that it simply showed a file was deleted and a new one was created.

lionel-:
yes the practical problem is that git doesn't always detect renaming if you make other changes in the new file. So to be safe the commit should not have any other changes.

By the way I'm not sure anymore that a new `r-mode` is the way forward. I think we should evolve within the ESS framework towards real derived modes. Then we can make the switch to `r-`, `stata-` etc namespaces, with files containing compatibility aliases to old variables and public functions. In short I think incremental changes is the way forward rather than rewriting everything. Though one advantage to rewriting is that we sort out the copyright issue and qualify for ELPA inclusion. But I think it should be possible to track down areas written by unreachable contributors and rewrite or remove those, it's probably a quite small portion of the code.

But we can discuss all this later on the mailing list and rename to `ess-r-mode.el` for now if you prefer (since we can always have a compatibility `(provide 'ess-r-mode)` later in the `r-mode.el` file). Does that sound good?

vspinu:
> But we can discuss all this later on the mailing list and rename to ess-r-mode.el for now if you prefer (since we can always have a compatibility (provide 'ess-r-mode) later in the r-mode.el file). Does that sound good?

No, it does not. That creates confusion and potentially file clashes. I don't see any problem with `ess-xyz-` prefix for files. If modes are named all with ess-xyzmode, then let's stick with the same file names.

lionel-:
> No, it does not.

I suggested renaming with the prefix, noting that we can always switch to unprefixed later if the dialects become proper modes. Does that sound good? ;)

vspinu:
Arh, sorry, my bad. Sure, we are on the same page then. If Martin has no further objections other than tracking historical changes, then let's do that then.

mmaechler:
Yes,  easily tracking history has been my concern.  It may be that in all the "bad" cases I did both rename and change before committing and so thought this was similarly problematic as with svn.

lionel-:
@mmaechler I renamed R and Stata files and checked that the history wasn't altered. I think the main danger is when you move code from one file into another. For example, if we moved the code from `ess-stata-lang.el` (previously `ess-sta-l.el`) to `ess-stata-mode.el`, we'd lose history for that moved code. I'd love to hear about a workaround for this issue if anyone knows about one.

FYI I've been slowly rewriting the parser code to something much simpler. Once I've finished that, we'll have a proper way of analysing surrounding R code with cursor at arbitrary positions (inside tokens etc). Then I'll tackle all the block/defun related issues.

PR #420, opened 2017-04-30
Latex completion from julia REPL

antoine-levitt:
This is the shortest patch ever, but it works for me and I don't see any issue, so here goes. To test, run a julia REPL, type \alpha and press TAB.


vspinu:
Thanks!

PR #421, opened 2017-05-01
julia: set EDITOR environment variable to emacsclient

antoine-levitt:
Trivial one-liner, but it makes `@edit sin(pi)` do the job (without, the EDITOR variable is unset and it calls nano, which messes with the terminal)


vspinu:
This is not the place to change environment variables. It will overwrite any user specified one. 

The right place is `.bash_profile` or `(setenv ...)` in user's `.emacs`.

antoine-levitt:
That's a fair point, but since the terminal is very basic, many editors will fail (at least the terminal ones), in such a fashion that the only way out is to kill the buffer (potentially losing data). How about a compromise? If EDITOR is unset and the server is running, then set EDITOR to emacsclient.

vspinu:
I have a custom setup for my editor and don't want some 3rd parties to set it up. 

A compromise would be to set it when the environment variable is empty, but then you have to make sure that it works equally well on all OSes and both in X and terminal.

PR #423, opened 2017-05-15
remove unmatched parenthesis from julia logo

antoine-levitt:
Julia's logo has an unmatched parenthesis, which wrecks havoc with modes like autopair. This removes it.


lionel-:
Thanks!

PR #425, opened 2017-05-16
fix latexsub completion

antoine-levitt:
As it turns out, https://github.com/emacs-ess/ESS/pull/420 was wrong and broke completion. completion-at-point-functions has a weird interface! This should hopefully fix it.


antoine-levitt:
Hi, could this be merged? The current version is badly broken (by my fault in https://github.com/emacs-ess/ESS/pull/420)

lionel-:
Thanks!

Issue #441, opened 2017-07-18
ESS Julia eldoc feature not working


my-little-repository:
I just downloaded julia-mode and ESS (ess-version: 16.10-1 [elpa: 20170710.118]) from MELPA and neither completion nor eldoc work. 

When `ess-eldoc-show-on-symbol` is true and I  type the word `map` at the *julia* repl, eldoc reports the following error. 

     eldoc error: (wrong-type-argument sequencep map) 

I browsed ESS source and I can't make sense of the code defining the function `ess-julia-eldoc-function` in the  file `ess-julia.el`, see https://github.com/emacs-ess/ESS/blob/master/lisp/ess-julia.el#L259. The local variable funname is first set at the value given by `(ess-symbol-at-point)` which returns a symbol. Then the length of the symbol is computed. Symbols do not have length, hence the error.

This is on GNU Emacs 25.1.1 of 2017-04-23.

cszang:
Same problem here. I tried with both ESS 16.10-1 from MELPA and current ESS git HEAD (064174931f0cbf91ad24fda883909a06eee10f6e), also with emacs -Q and fresh installs of ESS and company. 

Confirmed for both Emacs 25.2.1 and 25.1.1. ESS works just perfect with R with both company and auto-complete.

vspinu:
Folks, we are tired of tracking julia's changes. Mostly because none of us is actively using julia. Please have a look at [ESS module](https://github.com/emacs-ess/ESS/blob/master/etc/ess-julia.jl#L1) see what is broken and change whatever you see fit (preferably backward compatible). We would be more than happy to accept any PRs on this. ESS module is close to trivial, anyone with even basic julia knowledge should be able to figure it out.

cszang:
Fair enough, thanks.

vspinu:
Fixed by @nverno in #442 .

PR #442, opened 2017-07-19
replace ASCIIString with AbstractString in julia module

nverno:
Replaces ASCIIString with AbstractString in the julia module.
ASCIIString is gone from v0.6, so the completion module stopped
working.  Eldoc still seems not to work. -- #441

### DIFF
--- a/etc/ess-julia.jl
+++ b/etc/ess-julia.jl
@@ -18,39 +18,34 @@ function fun_args(m::Method)
     if !isempty(tv)
         Base.show_delim_array(io, tv, '{', ',', '}', false)
     end
-    li = m.func.code
-    e = Base.uncompressed_ast(li)
-    argnames = e.args[1]
     print(io, "(")
-    print_joined(io, [escape_string(isempty(d[2]) ? d[1] : d[1]*"::"*d[2]) for d in decls], ",", ",")    
+    join(io, [escape_string(isempty(d[2]) ? d[1] : d[1]*"::"*d[2]) for d in decls], ",", ",")    
     print(io, ")")
 end 
 
-VERSION >= v"0.4-" && (Base.function_module(f::Function)=f.env.module)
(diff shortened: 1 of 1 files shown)


vspinu:
Would you be so kind to look at eldoc as well? I guess they changed the representation of function objects so there is no `env` slot there:

```julia
julia> ESS.fun_args(cat)
ERROR: type #cat has no field env
 in fun_args(::Function) at ./none:33
```

Should be easy to fix.  There might even be an exposed API by now for args and help string.

nverno:
I've barely used Julia myself, but I'll search a bit and see if I can figure it out.  I think the != comparison should be against String, not AbstractString?

nverno:
@vspinu Would it be possible to provide a partial example of what the output should look like from your example above? Should it be a list of the form,

    ESS.fun_args(cat)
    > (list nil nil '((\"catdim::Integer\") (\"dims::Integer\", \"X::Union{BitArray, Bool}...\") ...)

 I just started reading through the Julia abstract method table documentation.  I will try to make an attempt to parse it, but since my knowledge of Julia is very basic, it may be ugly/take me a while.

It would be great if you could confirm I'm on the right track before I head down the wrong road.  I am redefining `function_module` as 

    VERSION >= v"0.4-" && (Base.function_module(f::Function)=typeof(f).name.module)

and the method table and `defs` seem to be,

    mt = typeof(cat).name.mt
    typeof(cat).name.mt.defs

However, the dispatch to `fun_args(m::Method)` seems to need some modification.  If this seems right so far I will dig a little deeper.
Thanks

nverno:
Ok, the eldoc seems to work on v0.5 and v0.6.  I didn't test for earlier versions.
I didn't change any elisp, so there is still the problem when `ess-eldoc-show-on-symbol` is
 non-nil since it returns a symbol instead of a string.

vspinu:
Thanks nverno. You changes indeed work for me. Thanks a bunch! I have fixed the string issue on the elisp side.

PR #455, opened 2017-09-23
More robust reload command

lionel-:
This makes the reload command work during a debugging session or when R is prompting for input (e.g. after a crash). There's a new function `ess-debug-alive-p()` to check whether we are debugging or if recover is prompting for a debugging frame.

Includes a refactoring of debug commands to simplify control flow, reduce verbosity, and force the process. I think ideally all interactive commands should force the process if one is not attached. What do you think?


lionel-:
Now extracted the interrupt logic into `ess-interrupt()`. It is called in `ess-quit()` in addition to `inferior-ess-reload()`.

vspinu:
Looks good. Thanks @lionel-!

PR #456, opened 2017-09-23
Implement ess-r-package-use-dir()

lionel-:
For switching process directory to the current package directory.

This factors `ess-use-dir()` out of `ess-use-this-dir()`. Compared to the latter which sets the wd to the buffer directory, the former takes a directory as parameter.

I'm not sure what keybinding to give to `ess-r-package-use-dir()`. I think it should have one as it is quite useful for package developers.


lionel-:
> I'm not sure what keybinding

`<M-S-return>` seems to make sense.

vspinu:
How about switching backwards on C-u prefix? 

For  FWIW my personal use case is to start a new R session inside project dir. Feels a bit cleaner this way.

lionel-:
> for non-package R scripts, what would the typical use case be

It would issue an error indicating that the file is not part of a package.

> How about switching backwards on C-u prefix?

Good idea.

> For FWIW my personal use case is to start a new R session inside project dir

I often start sessions from files inside `/R` or `/src`. I think it'd make sense to automatically switch to the package directory when a session is started. And also when it's reloaded? I'll add this to the PR.

vspinu:
>  I think it'd make sense to automatically switch to the package directory when a session is started. 

We can ask for package root directory at startup. This might be enough for all needs actually. You can always `ess-set-working-directory` one dir down when desired.

lionel-:
Launching or reloading R from a package file now sets the working directory to the package directory if a package is detected. This way there shouldn't be any need to call `ess-r-package-use-dir()` so I didn't expose it in the UI (keybinding or C-u prefix for the use dir command).

doc/newfeat.texi
```diff
@@ -15,6 +15,13 @@ process is attached, ESS now switches automatically to one (prompting
...
 will now work during a debug session or when R is prompting for input
 (for instance after a crash).
+
+@item @ESS{[R]}: New command @code{ess-r-package-use-dir}. It sets the
+working directory of the current process to the current package directory.
+
+@item @ESS{[R]}: Launching or reloading R from a package file now sets
+the working directory to the package path. You can now launch R from
+files within @code{R/} or @code{src/} without worrying about the wd.
 @end itemize
```
vspinu: Actually what I had in mind is something much less intrusive. I thought that proposing package dir on `M-x R` by default should be enough. 

I am pretty sure this new behavior will be confusing to the users. People might want to start in a sub-dir for whatever reason and messing up with their wd in the background seems like a bad design to me.

lionel-: It's not in background, there's a visible `setwd()` that appears in the terminal. I think we talked past each other, I thought you suggested that behaviour.

I don't think `M-x R` should be prompting anything to the user, that'd be annoying. We could make it an option but I think it should be enabled by default as this looks like good design to me that the working directory is set to the project root by default.

lionel-: I think purposefully setting the wd inside `R/` or `src/` or other subfolders is a very rare use case. I spend my time developing packages and I never needed that. It's always possible to switch to a subfolder after having started R.

lionel-: Sorry, I had completely forgot about `ess-ask-for-ess-directory` because I disabled it years ago. How about this:

- if non-nil, we ask with the package directory as default
- if nil, we switch to the package directory directly (but verbosely)

vspinu: Sounds like a good tradeoff. Also, could you then add a line or two of clarifications to `ess-ask-for-ess-directory`? I am keeping that option on as I find it occasionally useful to be worth one extra RET.

mmaechler:
Did not follow this issue closely, but I think I'm very close @vspinu and his typical behavior here.  I think I  rarely would appreciate that ESS changes the wd in the background without user interaction.

Only loosely related:  When using an R script from inside <pkg>/tests/*.R   ESS thinks this code should run inside the pkg namespace which is wrong:  It's true for all  <pkg>/R/*.R files, but wrong for all <pkg>/tests/*.R files.

lionel-:
I think we all agree. It's all about the value of `ess-ask-for-ess-directory`, I'll make it right soon.

Relatedly, Emacs core is gaining project capabilities so eventually we'll want to make the implicit project directory more general then just packages for the case of `ess-ask-for-ess-directory` set to nil.

lionel-:
@mmaechler 

> Only loosely related: When using an R script from inside /tests/.R ESS thinks this code should run inside the pkg namespace which is wrong

Is this about `ess-r-package-auto-set-evaluation-env`? I disabled it so I'm not seeing this effect. We could check for the current directory in `ess-r-package-set-namespaced-evaluation()`.

vspinu:
Adding to @mmaechler's  point. There are indeed two issues with tests, one is that code in tests should not be executed within a package, and another is that the code in tests should not be evaluated within package directory. Tests are evaluated with wd set to their own directory. Even testthat tests are executed in the `./tests/testhat/` directory. 

Another use case that I have is `./revdeps/` with `revdepcheck` package. I have an R script in that directory which assume that I am running within that directory. So, there are indeed use-cases when wd should not be automatically set to the package directory.

PR #457, opened 2017-10-07
Improve handling of starting directory

lionel-:
- Integrates `ess-r-package-mode` with Emacs 25.1 projects.
- The default prompt directory is now the current project directory. This includes R packages and version controlled projects (git, svn, ...).
- Prompt is made a bit nicer
- If `ess-ask-for-ess-prompt` is `nil`, switch to the default directory (so the current project if there is one)

The project stuff is optional and old Emacsen will just use current previous methods for choosing the default directory. I think eventually we should deprecate `ess-directory-function` in favour of Emacs projects `project-find-functions`.

Includes a bit of code cleanup.


lisp/ess-inf.el
```diff
@@ -149,33 +146,33 @@ Alternatively, it can appear in its own frame if
              (eq major-mode 'inferior-ess-mode))
         (setq startdir  (if ess-ask-for-ess-directory
                             (ess-get-directory defdir temp-dialect procname)
-                          defdir)
-              buf       (current-buffer)
-              ;; don't change existing buffer name in this case; It is very
-              ;; commong to restart the process in the same buffer.
-              buf-name-str (buffer-name)
-              method    1))
+                          defdir))
```
vspinu: Why over the sudden you decided for this? I aggregate setqs whenever I see the chance ;) Find it neater and more readable. But well, it's tiny.

lisp/ess-inf.el
```diff
@@ -581,21 +574,30 @@ This marks the process with a message, at a particular time point."
...
+;; Deprecate `ess-directory-function' in favour of `project-find-functions'?
+(defun inferior-ess--default-directory ()
+  (let ((dir (or (and ess-directory-function
+                      (funcall ess-directory-function))
+                 ess-default-directory
+                 default-directory)))
+    (directory-file-name dir)))
+
+(defun inferior-ess--get-directory (procname dialect)
+  "This returns the directory of the current project"
```
vspinu: Nitpicking, but let's do new docs according to emacs standards (imperative statement + properly document all params).

The name is not particularly suggestive. What directory does this return more concretely?  Shouldn't this just be part of `ess-prompt-for-directory`? It's a bit strange that a helper `--` function calls main utility function and not the other way around.

lisp/ess-r-package.el
```diff
@@ -87,6 +87,11 @@ whether the current file is part of a package, or the value of
       (with-ess-process-buffer t
         ess-r-package-info)))
 
+(defun ess-r-package-project (dir)
```
vspinu: Why do we need so many functions that do similar stuff `ess-r-package-project`, `ess-r-package--package-dir-info`, `ess-r-package-info`, `ess-r-package--find-package-path`. It's confusing. Cannot we reduce them to 1 max 2 functions?

BTW, shouldn't it be `ess-r-package-project-root` or `...-dir`?

vspinu:
Looks fine to me, but I would really want to see streamlining of ess-r-package API. It's quite hairy at the moment with all those "path/dir/root/info" functions.

lisp/ess-inf.el
lionel-: I thought aggregated setq were against coding standards but after a bit of googling it seems I was mistaken and there are two standard visions of two groups of people arguing over this ;)

Personally I prefer disaggregated because it attracts attention over small infelicities (mutation), are easier to refactor, and I personally find it more readable. I can switch it back to previous form if you prefer.

lisp/ess-inf.el
lionel-: >  let's do new docs according to emacs standards

Agreed. That was a leftover from the time where this was a `(def-generic ess-current-project ()`. I'll just remove the docstring.

> The name is not particularly suggestive.

Agreed, it should probably be `inferior-ess--get-startup-directory`. I don't like that function either and thought about merging it in `ess-prompt-for-directory` but it's a bit annoying because of that dialect specific handling. Also this checks for the value `ess-ask-for-ess-directory` to avoid multiple check in `inferior-ess-mode`.

> It's a bit strange that a helper -- function calls main utility function and not the other way around.

I think that's alright if it reduces repetition in another main function.

lisp/ess-r-package.el
lionel-: There's a function that does the searching + caching, one that returns the package info in ESS format, one that returns the package info in Emacs project format, ... I'm not sure we can streamline it further but a better naming scheme would make things clearer. I'll try to simplify.

> shouldn't it be ess-r-package-project-root

I don't think so because it doesn't return a project root, it returns a cons cell that includes the project name.

lisp/ess-inf.el
vspinu: Nevermind. It's not important. Not what we have some high standards in this project for coding style:). But in my code I always aggregate as I find all those side-effectful repeated `setq` noisy and annoying.

lisp/ess-r-package.el
vspinu: >  I'll try to simplify.

Thanks!

lionel-:
@vspinu Can you please take another look? I also tweaked the names of the source dirs functions as suggested in https://github.com/emacs-ess/ESS/commit/d559b3d0e7db97a74823c20fda4095d1a14d9209.

doc/newfeat.texi
```diff
@@ -28,11 +22,35 @@ from capturing the cursor.
...
-@item @ESS{[R]}: Launching or reloading R from a package file now sets
-the working directory to the package path. You can now launch R from
-files within @code{R/} or @code{src/} without worrying about the wd.
+@item @ESS{[R]}: Launching R from a package file now recognises the package
+root as the likely startup directory for R. If
+@code{ess-ask-for-ess-directory} is non-nil, ESS prompts with the
+package root as default selection. Otherwise the working directory is
+automatically set to the package root.
+
+The mechanism is a bit more general if you are using Emacs 25.1 or
```
vspinu: Do we really need this behavior? If I start R in a subdirectory of a git project I probably want to start in that sub-directory not in the root. 

This is an unlikely scenario though and I can leave with it if it indeed simplifies and modernizes the codebase. 

What about "test" sub-directory?

doc/newfeat.texi
```diff
@@ -28,11 +22,35 @@ from capturing the cursor.
...
+older. On those recent Emacsen R packages are now recognised as Emacs
+projects and ESS queries Emacs projects to get the default startup
+directory. This means that any git or subversion directory will be
+recognised as project root even if it is not an R package.
+
+In light of these developments, @code{ess-r-package-get-info} is
```
vspinu: Need not be in the newfeat, I think. `ess-r-package-get-info` is kinda an internal function, so I think we can just silently deprecate it. Discussion of this feature got a bit too technical.

lisp/ess-inf.el
```diff
@@ -116,25 +116,17 @@ Alternatively, it can appear in its own frame if
...
            (temp-lang temp-ess-lang)
-           (procname (let ((ntry 0) ;; find the next non-existent process N (*R:N*)
-                           (done nil))
-                       (while (not done)
-                         (setq ntry (1+ ntry)
-                               done (not
-                                     (get-process (ess-proc-name
-                                                   ntry
-                                                   temp-dialect)))))
+           ;; Find the next non-existent process N (*R:N*)
```
vspinu: Much nicer. I am now amazed that I could write that beast :/.

doc/newfeat.texi
```diff
@@ -28,11 +22,35 @@ from capturing the cursor.
...
+deprecated in favour of @code{ess-r-package-project}. This function
+returns an Emacs project instance, e.g. a cons cell of the project name
+as symbol and the package directory as string. In addition
+@code{ess-r-package-name} returns the current package name as a string.
+
+In the future (long term), we will thus deprecate
```
vspinu: Idem.

doc/newfeat.texi
lionel-: I think ESS badly needs a more integrated notion of project and this is the start of it. For data analysis your scripts should work from the root of the project not in subdirectories.

As for the test directory, maybe we can define a notion of subprojects within R packages? Could you describe the use cases for this please? I'm not sure when this would be useful, I never switch to the `test` folder in my workflow. In fact switching to that folder would be disruptive in my case. If we add this feature it should be optional.

> This is an unlikely scenario though

It happens all to the time in my workflow because I often switch to a test file with projectile and then open R. Or I switch to a file in `/R` or `/src`. Or `/.metadata` (which is a nice folder to put your package notes and scratch files because it is automatically ignored by `R CMD build` and you can add that folder in your global `.gitignore` too). I almost never open a file at the root and then open R.

vspinu: > the use cases for this please? 

The use case is as discussed earlier. Tests are evaluated (by R CMD check) in the test sub-directory not at the root. So if I start a fresh R session in test sub-dir, it probably shouldn't be at the root. For the eval part, tests should not be evaluate within the package.

> In fact switching to that folder would be disruptive in my case.

When my test fails, I jump to that particular test and try to figure it out locally. I find it hard to imagine a different workflow. 

> This is an unlikely scenario though

By this I meant that you open R script in a non-R-package.  My earlier concern were exclusively about such case.

lionel-: >  Tests are evaluated (by R CMD check) in the test sub-directory not at the root.

You're right, it might be a good idea to add a notion of subproject later on then. Maybe a better integration of R CMD check would be better, i.e. ideally it would run in a fresh process. Though that poses the question of simulating an installed package which is easiest with `load_all()` but this would impose devtools on users. I personally (almost) never execute these scripts manually because I use testthat and `devtools::test()`.

> When my test fails, I jump to that particular test and try to figure it out locally. I find it hard to imagine a different workflow.

To debug the test you typically don't need R to be in a particular folder, unless the test uses a local resource. I guess for the case of local resources (e.g. a csv file) it does make sense to be in a non-root folder. However it seems it would be hard for ESS to predict which folder because that is dependent on development tools and test scripts.

On the other hand being in a non-root folder makes it difficult to use tools provided by packages like `usethis` or `revdepcheck`.

> By this I meant that you open R script in a non-R-package.

Ok. I think the notion of projects should be helpful to data analysis users. It seems to be good practice for scripts to execute relative to the project root.

vspinu: > I guess for the case of local resources

Precisely.

> However it seems it would be hard for ESS to predict which folder because that is dependent on development tools and test scripts.

Let's just make an exception for "test" sub-directory and be done with it. I guess, we don't want to get too complicated on this. It's not an essential functionality anyways.

> It seems to be good practice for scripts to execute relative to the project root.

Good point. Let's do a bit of nudging then.

vspinu:
> Maybe it's best not to introduce another variable if no one is going to change it?

Sorry. I missed this post. Have you already introduced the variable? If not, let's not do it. I thought a bit more about it and I think it's best not to do any exceptions at all, even for the test sub-dir. Consistent behavior is a good thing.

PR #458, opened 2017-10-10
Clean up ESS noweb font lock mode

lionel-:
From Stefan Monnier.


lionel-:
@vspinu Should we merge this?

Issue #430, opened 2017-06-15
Problems with traceback when sourcing files


ganesh-krishnan:
When I source file using C-c C-l (which calls `.ess.source()`), the traceback output stops at the `.ess.source()` level.

**Reproducible example**
Create a file test.R with:

```
abc <- 1
def
```

Now source with C-c C-l. Traceback output is shown below:

```
6: stop(msg, call. = FALSE)
5: value[[3L]](cond)
4: tryCatchOne(expr, names, parentenv, handlers[[1L]])
3: tryCatchList(expr, classes, parentenv, handlers)
2: tryCatch(ss(file, echo = visibly, local = local, print.eval = output, 
       max.deparse.length = max.deparse.length, keep.source = keep.source), 
       error = function(x) {
           if (fake.source) {
               .ess.file.remove(file)
           }
           msg <- .ess.strip.error(x$message, file)
           stop(msg, call. = FALSE)
       })
1: .ess.source("test.R", visibly = FALSE, output = TRUE)
```

I'm on R version 3.3.2 with ESS version 16.10-1. 

Is there something I'm missing or is this a bug?

mmaechler:
Yes, this is a bug..... and indeed a regression.   The fact that R's
traceback() stops with .ess_source(...)  is unfortunate... but maybe not
easy to workaround because we (ESS) uses  tryCatch() inside .ess_source()
so "we" catch the error... maybe we "treat it" differently (i.e. _should_ be smartert in treating it)

OTOH, note that if you use a bit a larger file,
then  ESS   C-c `  (or via menu: [ESS] -> [Tracebug] -> [Show traceback])
gives a nice  *ess-traceback* buffer:

    -----------------------------------------------------------------------------
    + Error in eval(ei, envir) (from source-err.R#5) : object 'def' not found


    5: eval(ei, envir)
    4: eval(ei, envir)
    3: withVisible(eval(ei, envir))
    2: ss(file, echo = visibly, local = local, print.eval = output, 
	   max.deparse.length = max.deparse.length, keep.source = TRUE)
    1: .ess.source("/u/maechler/emacs/ess/etc/source-err.R", visibly = FALSE, 
	   output = TRUE)
    -----------------------------------------------------------------------------


in which the first line contains  file + line-number
which is even clickable, so you jump to the place of where the error
happened which is nice.

ess-core:
I don't know other fix for this than to remove that tryCatch. This issue is
rather annoying especially in the context of package development.

[...]

PR #459, opened 2017-10-15
Fix traceback not showing properly

ganesh-krishnan:
Tracebacks are not being shown properly because the tryCatch handler does not receive the information. This PR fixes that by using `on.exit` as suggested by @ess-core  in this [comment](https://github.com/emacs-ess/ESS/issues/430#issuecomment-308995905).

### DIFF
--- a/etc/ESSR/R/.basic.R
+++ b/etc/ESSR/R/.basic.R
@@ -96,30 +96,23 @@
                         message.prefix = "")
 {
     if (is.null(local)) {
-        local <- if (.ess.Rversion > '2.13') parent.frame() else FALSE
+        local <- if (.ess.Rversion > "2.13")
+            parent.frame()
+        else FALSE
     }
+    ss <- if (.ess.Rversion >= "2.8")
+        base::source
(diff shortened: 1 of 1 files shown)


vspinu:
Thank you!

lionel-:
Since this PR the following test fails: https://github.com/emacs-ess/ESS/commit/4260522d137690c4dada2ae84344e0f312858c99

This test was to make sure that error messages do not mention the temporary source files. Since the error are no longer stripped evaluating a region containing `foo bar` now results in this kind of error message:

```r
Error in ss(file, echo = visibly, local = local, print.eval = output,  : 
  /var/folders/t_/5cfrj9810ydf1wbk0r8xg7h00000gq/T/namespace.R!74335qWV:1:5: unexpected symbol
1: foo bar
        ^
```

Overall it seems a lesser evil compared to the annoyance of bad stacktraces though.

vspinu:
We should remove the test then. I don't see a major issue with temporary files and it's hard to hide those anyways. As long as the navigation to those errors works as expected it should be fine.

PR #460, opened 2017-10-29
Fix verbose startup

lionel-:
This makes ESS set up the R state and options silently at startup. The goals are:

1. To make important changes to the state more obvious.
2. To make the R startup look clean when using ESS.

About 1, the startup directory is verbosely selected when `ess-ask-for-ess-directory` is `nil`. Maybe it'd even be a good idea to always show the `setwd` verbosely even when the user made the selection? It's the most important information about the process state at startup.

About 2, a tricky part was to avoid sending a visible newline at startup. I had to reintroduce `inferior-ess--wait-for-prompt`. If `no-wait` is set, it times out after half a second.


lisp/ess-inf.el
```diff
@@ -705,6 +710,21 @@ process happens interactively (when possible)."
               (ess-get-process ess-current-process-name))
           (error "Process %s is not running" name))))))
 
+(defun inferior-ess--wait-for-prompt (seconds)
```
vspinu: Having two extra functions to save one extra ">" is an overkill. We will end up with  two idle mechanisms one with :busy process property and one which looks for prompts in the inferior. I don't think it's worth it.

lionel-: I know I'm desperately perfectionistic, but I think it's worth it. The double `> >` looks very sloppy to me. It's barely 12 lines of internal code ESS can survive it ;).

mmaechler:
I think it's a good idea to make the   setwd()   explicit in the transcript.

lisp/ess-inf.el
vspinu: Well, I think in this case you are perfectionist about very unimportant things. You have loads of "> >  >" generated during the evaluation, why bother about extra >? How about having a perfectly clean code base instead :)  

> It's barely 12 lines of internal code ESS can survive it ;).

Well, the main thing is not the code, but the naming which apparently indicates the same intent as built-in :busy functionality.  If you really want to clean startup screen, how about the simple search-replace in the buffer? We already clean something there, it could be made more generic by fixing the double prompt plus all the potential irregularities in an uniform fashion.

lionel-: > Well, I think in this case you are perfectionist about very unimportant things. 

Of course otherwise this wouldn't be perfectionism.

I'll try the search-replace. Perhaps we could merge this as is for now with a FIXME comment as Martin wanted to release soon?

> You have loads of "> > >" generated during the evaluation

Oh those are a major bother as well, they produce unaligned outputs :(
This would be a very good reason to revisit the FIXME mentioned above.

lionel-:
@mmaechler @vspinu Do we want this in the next release or do you prefer to wait? The goal is to clean up the verbosity of process startup so that it looks less messy and so that only important information shows up to the user, e.g. the startup folder. This is especially important for users that have `ess-ask-for-ess-directory` set to nil, it makes the process startup folder more tangible:

startup

vspinu:
I think changes are quite innocuous but given that there is wait-for-prompt there I would leave it for after the release. We don't want hanging inferior buffers at startup. I also would like to see naming of "idle-prompt-p "  changed into something more informative. Maybe "comint-ends-with-prompt-p" or something?

jabranham:
This works well for me, except that M-x R leaves the point behind the prompt. I submitted a PR to @lionel- 's branch that solves that by calling `(comint-goto-process-mark)` at the end.

lionel-:
@vspinu I just realised that since we are going for explicit setwd's in the process buffer, we can just use the `setwd()` to catch the prompt.

I also removed the `ess-directory` variable in favour of the process-local variable `default-directory` (with a helper `inferior-ess-default-directory` to retrieve it). This should make things simpler.

Is this now good to merge? (I'll add a couple of tests for visible working directory switching)

vspinu:
Looks good to me. It might create conflicts with @jabranham's PR.

PR #469, opened 2017-12-30
Update essd-els.el: remote-inferior-ess-pager

ShuguangSun:
remote-inferior-ess-pager: If we work on Windows and SSH to a unix server, the pager should not be console (as default for Windows).

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -836,8 +836,8 @@ Returns the name of the selected process."
   (setq ess-dialect (or ess-dialect
                         (ess-completing-read
                          "Set `ess-dialect'"
-                         (delete-dups (list "R" "S+" S+-dialect-name
-                                            "stata" STA-dialect-name
+                         (delete-dups (list "R" "S+" (if (boundp 'S+-dialect-name) S+-dialect-name "S+")
+                                            "stata" (if (boundp 'STA-dialect-name) STA-dialect-name "stata")
                                             "julia" "SAS" "XLS"  "ViSta")))))
 
   (let* ((pname-list (delq nil ;; keep only those mathing dialect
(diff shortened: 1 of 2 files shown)


lionel-:
I think the variable should be called `inferior-ess-remote-pager`, be `nil` by default, and then test with:

```
(or inferior-ess-remote-pager inferior-ess-pager)
```

ShuguangSun:
Thanks. Updated accordingly.

lionel-:
@ShuguangSun Thanks! This should make your patch backward compatible with user configurations.

@vspinu @mmaechler does this contribution sound reasonable? I never used ESS with remotes so I can't tell.

vspinu:
I think this is reasonable, but generally better approach would be to query R process for the OS once it's started. Maybe we can have the pager set to"auto" by default.

mmaechler:
Good idea, @vspinu  (considering "auto" and have ESS ask R about it).   Note that this  is quite delicate.
We had a big problem in the previous setup ca one year ago,  when our OS (Fedora Linux) changed it's behavior with respect to the definition of the PAGER environment variable as that does influence R's own pager setup and behavior.   In other words: This is delicate because R has a flexible setup that can be tweaked in several ways (environment variable,  options(), .. and is all OS/Platform dependent.

vspinu:
> This is delicate because R has a flexible setup that can be tweaked in several ways (environment variable, options()

I am not quite following. Why shall we care about R's configuration with respect to the PAGER settings? All *nix distribution have `cat` and we need to use it for proper pagination of help files, right?

ShuguangSun:
Sorry to mixed to pull requests up.
The new one is for ess-inf.el and make it more robust.

I customized not to load stata and under the default setting, if I run ess-roxygen-preview-text, it will fail because it can't find STA-dialect-name.

vspinu:
I think this is better than nothing. ess-els is semi-obsolete anyway, so not much work will be expected here anyways. `STA-dialect-name` treatment is also good. The minor issue is custom mismatch, but I will fix that myself. Thanks @ShuguangSun!

PR #470, opened 2018-01-10
Only activate package mode in editing buffers

### LINKED ISSUE #466: *shell* buffer in ESS "package mode"
Not a killer bug; but still awkward and a bit embarrassing:

After having looked a bit at CRAN package `diceR` (a test of which was a bit fishy and failing in R-devel only).... The modeline in *shell* mode is showing as if it was an R buffer in "package-mode", see screen shot [at SfS ftp](ftp://stat.ethz.ch/U/maechler/ESS/screenshot_Emacs_shell_pkg-mode_2017-11-25.png)... Ok, ftp fails; using [Dropbox](https://www.dropbox.com/s/3pqcg6rx0vm6kmk/screenshot_Emacs_shell_pkg-mode_2017-11-25.png?dl=0)

lionel-:
Closes #466 

With this patch `ess-r-package-mode` only auto-activates in editing buffers, i.e. ones whose major mode inherits from `text-mode` or `prog-mode`. Note that ESS modes now inherit from `prog-mode` since https://github.com/emacs-ess/ESS/commit/b439eb34ddaf01d336bd70c482436f665ea7cefe.

I agree with @mmaechler that this makes more sense. WDYT @vspinu?

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -6,6 +6,10 @@ Changes and New Features in development version:
 
 @item ESS modes now inherit from @code{prog-mode}.
 
+@item @ESS{[R]}: The package development minor mode now only activates
+within editing buffers, i.e. ones that inherit from @code{prog-mode} or
+@code{text-mode}.
+
 @end itemize
 
 
--- a/lisp/ess-r-package.el
+++ b/lisp/ess-r-package.el
@@ -426,9 +426,7 @@ disable the mode line entirely."
   "Activate developer if current file is part of a package."
   (when (and ess-r-package-auto-activate
              (derived-mode-p 'text-mode 'prog-mode)
-             (not (memq major-mode '(minibuffer-inactive-mode fundamental-mode)))
-             (or (buffer-file-name)
-                 default-directory))
+             (or (buffer-file-name) default-directory))
     ;; FIXME Emacs 25.1: Use `when-let'
     (let ((pkg-info (ess-r-package-project)))
       (when pkg-info
--- a/lisp/ess-r-package.el
+++ b/lisp/ess-r-package.el
@@ -425,7 +425,7 @@ disable the mode line entirely."
 (defun ess-r-package-auto-activate ()
   "Activate developer if current file is part of a package."
   (when (and ess-r-package-auto-activate
-             (or (provided-mode-derived-p major-mode 'text-mode 'prog-mode))
+             (derived-mode-p 'text-mode 'prog-mode)
              (not (memq major-mode '(minibuffer-inactive-mode fundamental-mode)))
              (or (buffer-file-name)
                  default-directory))
--- a/lisp/ess-r-package.el
+++ b/lisp/ess-r-package.el
@@ -425,6 +425,7 @@ disable the mode line entirely."
 (defun ess-r-package-auto-activate ()
   "Activate developer if current file is part of a package."
   (when (and ess-r-package-auto-activate
+             (or (provided-mode-derived-p major-mode 'text-mode 'prog-mode))
              (not (memq major-mode '(minibuffer-inactive-mode fundamental-mode)))
              (or (buffer-file-name)
                  default-directory))
--- a/test/ess-r-tests.el
+++ b/test/ess-r-tests.el
@@ -51,7 +51,7 @@
 
 ;;; ess-r-package-mode
 
-(ert-deftest ess-r-package-auto-activate ()
+(ert-deftest ess-r-package-auto-activation ()
   (with-temp-buffer
     (text-mode)
     (hack-local-variables)
@@ -59,6 +59,12 @@
   (with-r-file "dummy-pkg/R/test.R"
     (hack-local-variables)
     (should ess-r-package-mode)))
+
+(ert-deftest ess-r-package-no-auto-activation ()
+  (with-r-file "dummy-pkg/R/test.R"
+    (eshell)
+    (should (not ess-r-package-mode))
+    (kill-buffer)))
 
 
 ;;; Namespaced evaluation


mmaechler:
Thank you, @lionel- .  This would be much better. 
 (May still be quite intrusive for my taste, unless this only concerns editing buffers related to a file where that file is clearly inside a package)

lionel-:
Minor modes are supposed to be a minimum intrusive ;)

You can disable auto-activation by setting `ess-r-package-auto-activate` to `nil`.

lionel-:
>  unless this only concerns editing buffers related to a file where that file is clearly inside a package

oops I missed that part, yes the package mode only auto-activates within editing buffers that are part of an R package.

vspinu:
Is this hardcoded now? There are a lot of relevant buffers even if they are not designed for editing (ediff, magit, ibuffer, ESS-help buffers, shell for me). So do you propose to explicitly hunt all those buffers now? I am sorry, but this is a wrong direction.

I am running actions like install package (C-c C-w i) from anywhere and I don't want to constantly switch to .R buffer to do so. Not having it in all relevant buffers is a huge inconvenience, a much bigger inconvenience than having it active in arguably(!) irrelevant shell buffers. 

I insist that we should have an exclusion variable (`ess-R-exclude-autoactivation-modes`) instead of an inclusion one. And, ideally, it should be empty by default.

lionel-:
How about a variable `ess-r-package-auto-activate-modes` in which you can supply a list of symbols of major mode parents. I think the default should be `prog-mode` and `text-mode` and then you can set it to `fundamental-mode` in your user config.

vspinu:
This could be, but then I would like to see fundamental-mode by default.  I fail to see why having it in text-mode is more vailuble than having it in shell or ediff buffers. Quite the opposite actually; both shell and ediff are more related to prog-modes than text-mode.

lionel-:
Because shell, eshell, dired etc are not tied to a package buffer, they are general purpose Emacs tools that you use and reuse and possibly wander far from the original package folders. It just seems cleaner to me that only editing buffers enable the package mode so I would prefer that to be the default. I like that we make it possible to use it for any mode though with that variable though.

vspinu:
> possibly wander far from the original package folders

Who does that? You normally re-start the buffer or open another shell. It's much faster than the cd command. Most of the time shell stays in the original dir.

What about ediff, magit (with dozens of sub-modes), dired, compilation, grep, igrep, occur (just to name a few)? All these modes are tightly connected to the current project. Shall then @mmaechler (or someone else  person who doesn't like shell integration) list all those modes explicitly?

lionel-:
FWIW I use the package mode all the time and I don't think I ever called a package command from elsewhere than an editing buffer. It doesn't feel natural to me. I think it's because I only consider editing buffers to be linked to the R process and not other processes. I _think_ this fits how most users intuitively think about process interaction in ESS.

Overall I don't feel about this matter strongly enough to discuss it all day though ;)

vspinu:
I do it from all buffers and it was quite irritating in the beginning when it wasn't activated in cpp or magit buffers. 

> Overall I don't feel about this matter strongly enough to discuss it all day though ;)

Same here. It's a tinny issue.

mmaechler:
@vspinu:  I understand that for you, your proposal is ideal.  This will also be the case for people who only do R programming / R package development in that instance of emacs.
Many emacs users are different -- particularly those whose use emacs as it was intended by some: Emacs is their complete interface to almost all of the OS.  They do many other things in that same emacs which are related to their day job,  but not to R programming.    For those people your proposed default is very aggressive, They would perceive it as   "ESS is taking over Emacs".   For this reason, I very strongly agree with @lionel- and disagree with your preference of default.

If ever ESS should become more part of an extended "standard" emacs setup, ESS has to behave itself..

vspinu:
My main issue is that that if we are already polluting foreighn modes (text-modes, prog-modes), it's not such a big deal to "pollute" all of them.

But ok. If you feel that having an extra minor mode activated is such a big deal. Let's move with the @lionel-'s proposal. But we must state the `fundamental-mode` trick in the documentation.

lionel-:
Martin this wasn't finished. When the person sending a PR has commit rights you normally them merge ;)

mmaechler:
@lionel- I'm sorry... misunderstood your intent / idea.  I really did not want force anything!

vspinu:
Setting this variable to `fundamental-mode` doesn't work. `(derived-mode-p 'fundamental-mode)` returns nil everywhere for me.

lionel-:
Oh hmm... looks like `prog-mode` inherits from `nil` rather than `fundamental-mode`. I'll change it so that setting the variable to `nil` means enabling the package mode everywhere.

lionel-:
ok we don't need a new variable actually. I just made `ess-r-package-auto-activate` take lists of modes. If `t` the package auto-activates unconditionally, if `nil` it doesn't auto-activate.

https://github.com/emacs-ess/ESS/commit/ec2c2f5649d01071ce9a7079072ef415f9426149

vspinu:
I have been using this with the default settings since the change and I now insist that this was a bad idea. Three patterns keep re-occurring for me:

- I jump to shell and do `./cleanup` (script which is part of many R packages) and I want to re-install the package immediately from the shell. Shell buffer is part of package development workflow for many people, no doubts about that.
- I go to Makevars, change something, then C-c C-w, but it doesn't work because Makevars is in fundamental mode by default.
- I commit something with magit and before pushing to the upstream I want to run CMD check (C-c C-w C-c), which doesn't work. 

We are really creating more problems than solving here. I propose to change the default and make exclusion variable instead of the inclusions. It does make much more sense this way. If ess-r-package-mode interferes in some other mode then the user could simply deactivate it in that mode. Listing all modes where to activate the package is way more tedious operation.

lionel-:
I don't care much about this issue so I'll let Martin weigh in, but I don't have any of these problems even though I use eshell, magit and the package mode. Also `Makevars` should open in `makefile-mode` if you configure it properly.

lionel-:
It feels like Emacs should have a global convention for setting up inclusion and exclusion lists of modes for global minor modes (this would be useful for the package mode even if it's not a global mode per se).

lionel-:
> Also Makevars should open in makefile-mode if you configure it properly.

I do miss the package mode in `DESCRIPTION` and `NAMESPACE` which both open in fundamental mode.

vspinu:
> Also Makevars should open in makefile-mode if you configure it properly.

I never bothered configuring it and I bet most of the casual Makevars users don't do that either.

> I do miss the package mode in DESCRIPTION and NAMESPACE which both open in fundamental mode.

Indeed, I missed the elephant the in the room. 

In any case, I started feeling strongly on this. So if either of you are not glued to it, let's please change. I really think we have an unfortunate design here.

>  I'll let Martin weigh in,

Martin, do you use any of the ess-r-package functionality routinely? If not, I hope that from the above discussion you could relate to the issues that this change creates. An extra "pkg" in the mode-line is really not a good reason for the inconveniences that this "solution" creates.

vspinu:
More I think about it more I realize that we haven't defined our problem rigorously. Our ess-r-package functionality is for "R packages" not for ess-mode or inferior-ess-mode or any other mode. It's designed for broader contexts than emacs modes are intended for. An emacs mode is just one of the possible definitions of the current "context", but it's not the only one, nor it's the coarsest or most refined one.

mmaechler:
Even though I'm vacationing w slow/bad internet connection.  I do protest.
The shell buffer e.g., I use for many things  during the time that emacs is open.
Remember:  To the aficionado, Emacs is the operating system.   R package development is just one of the many things I use my open emacs for,   and that is true for  "true" or old school emacs users.
ESS is *not* to take over emacs  ... definitely not by default.
And I would *shell* never want to be part of package mode, even when I am mainly working with one package -- which is about one third of my emacs time, I'd guess.   Also often there are package dependencies I have to debug... and I want  *shell* behave independently.  Also I often have several *shell* buffers in the same emacs, and at most *one* of them should be tied to a given package in package mode.

mmaechler:
I definitely have  'Makevars' in makefile-mode  and have had so for 10-20 years...  others like me would be frustrated if ESS interfered with that.

vspinu:
Martin, we need to make tradeoffs based on common sense and use patterns. Nothing is perfect, but we should strive to minimize the damage and maximize the benefits. Current situation interferes with at least 5 concrete use patterns, but it still unclear to me how package mode interferes with your patterns. You never gave a concrete example and your argument is rather generic. No-one is taking over your emacs or interfering with *all* your shell buffers, or changing your makefile-mode. We just adding one minor-mode to the hundreds of minor modes out there.

Let's have a sober look at it. We want the package mode to be active within a context of the package, right? Given current emacs implementation, the best way we can isolate this context is by looking at `default-directory` and check if that directory is part of the package.   The functionality of `ess-r-package` has little to do with text-modes, prog-modes, ess-mode or even inferior-ess-mode (except that it uses it as a communication mechanism).  Thus, restricting it to some specific modes doesn't make much sense. I hope we can all agree on that.  Add to this the frustrating UI which interferes with the common use patterns and it should be obvious that we have a sub-optimal design at the moment. 

Finally, if we don't have a solid agreement on this, we should stick to the backward compatible default which was in place since 2016 (2 releases). The fact that it took you, Martin, 1.5 years to notice the issue suggests that it's not such a big problem after all ;)

mmaechler:
Note: the issue is not that  package-developer mode is auto-activated (which happened many moons ago); that has been a phantastic positive move.

The issue is that since very recently that mode interfers too much with  emacs' `default-directory`, in a way where `M-RET` hangs for me, and if it does not hang takes clearly much too long.  I would like the interference with my `*shell*` to be changed.  If I do  `cd`, `pushd` or `popd` in my *shell*, `default-directory` must "follow" as it has been doing in Emacs  shell-mode forever.
I do `cd` (in *shell*) and then  find-file (`C-x C-f`) and of course want to be able get the correct directory and not a very different.   Similar, if after the `cd` (or `pushd` or `popd`), I remain in the shell and want [Tab] completion on file names -- *relative* to my current path -- "current" of course with respect to *shell* , and *not* to ess-r-package variables.

lionel-:
> The issue is that since very recently that mode interfers too much with emacs' default-directory

I don't remember a change of this nature. Of course if the package mode is the culprit that is clearly a bug. It would help to pinpoint a commit where this started to happen.

vspinu:
I also don't see the connection between ess-r-package mode and shell buffer lookup. Is this remote? 

How to reproduce exactly? Going to shell and M-RET works as expected for me. 

How sure you are it's because of the package mode? Does `(setq ess-r-package-auto-activate nil)` fix the issue?

vspinu:
This issue started taking a toll on me. So I have just made a range of improvements to the r-package-mode which, I believe, should make everyone happy.

First, there is now an automatic detection of the changes of the `default-directory` in shell and eshell buffers.  Each time you do `cd` or `shell-resync-dirs`  package mode will automatically activate or deactivate depending on the new value of `default-directory`. This change is lightweight and won't tax the performance.

Second I have added an exclusion variable `ess-r-package-exclude-modes` which is more meaningful in the current context than the introduced by this PR "inclusion" semantics of the `ess-r-package-auto-activate`. Plus, the exclusion variable preserves backward compatibility.

Finally, ess-package-mode is now unconditionally activated in DESCRIPTION and NAMESPACE. I expect that most users will have both of these in fundamental mode.

lionel-:
These changes are fine with me but generally I think we should discuss UI changes in PRs.

lionel-:
The changes broke a test @vspinu, could you please take a look?

PR #472, opened 2018-01-11
Make `Makeconf` portable.

lq:
`which` does not have `-s` option on Linux.

### DIFF
--- a/Makeconf
+++ b/Makeconf
@@ -87,7 +87,7 @@ UNINSTALL = rm -f
 DOCDIR=$(PREFIX)/share/doc/ess
 
 # N.B. $(shell) is a GNU-ism: we need a workaround to be UNIX make friendly
-DOWNLOAD=$(shell which -s wget && echo 'wget -O -' || which curl)
+DOWNLOAD=$(shell command -v wget && echo 'wget -O -' || command -v curl)
 ##DOWNLOAD = wget -O -
 ##DOWNLOAD = curl
 
--- a/Makeconf
+++ b/Makeconf
@@ -65,7 +65,6 @@ EMACSBATCH = $(EMACS) -batch -no-site-file -no-init-file
 ##                 1st) source-file & 2nd) target-directory
 ## UNINSTALL       deletes all arguments
 ## DOCDIR          Destination of other doc files
-## DOWNLOAD        Download internet file to stdout
 
 SHELL = /bin/sh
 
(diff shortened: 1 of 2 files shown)


vspinu:
But `command` is not available either.  Or it's some GNU Make thingy? 

@rsparapa you added that line recently. Could you please have a look? I think we can safely remove `-s` option.

rsparapa:
Yeah, I don't know what command is either.  Here's what I have now tested on Linux and macOS which should do it.  DOWNLOAD=$(shell which wget > /dev/null && echo 'wget -O -' || which curl)
However, the purpose of DOWNLOAD is to get julia-mode.el.  But this target is only grabbed once; after that you have it and it doesn't get it again.  So, I think the julia-mode.el target of lisp/Makefile should be performed every time not just once.

lq:
That must be a bug of `make` on your system. `command` is a POSIX standard utility (http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html):

> The command -v and -V options were added to satisfy requirements from users that are currently accomplished by three different historical utilities: type in the System V shell, whence in the KornShell, and which in the C shell. Since there is no historical agreement on how and what to accomplish here, the POSIX command utility was enhanced and the historical utilities were left unmodified. The C shell which merely conducts a path search. The KornShell whence is more elaborate-in addition to the categories required by POSIX, it also reports on tracked aliases, exported aliases, and undefined functions.

lq:
On my Debian box `command` is not found too. It is implemented as a shell built-in function there, but `make` failed to handle it. However, `DOWNLOAD != command -v wget && echo ' -O -' || command -v curl` is OK. Conversely, on my mac, `$(shell ...)` works but `!=` does not. For now, it may be better just dropping the `-s` option.

lq:
@mmaechler That's irrelevant. Now it is downloaded only once. Better make julia-emacs a submodule?

vspinu:
That was the original idea, but it's hard (impossible?) given that julia-mode.el is part of the huge julia repo and we don't want to pull in the whole repo.

rsparapa:
Martin: Just droping the invalid '-s' as was proposed as well, seems feasible.
Rodney: No, just dropping -s is not correct.  You need do the following:
DOWNLOAD=$(shell which wget > /dev/null && echo 'wget -O -' || which curl)

lq:
@vspinu `julia-mode.el` resides in https://github.com/JuliaEditorSupport/julia-emacs/, which essentially only contain the lisp file.

vspinu:
This wasn't the case before. It should be submodule then, of course.

rsparapa:
Change made.
DOWNLOAD=$(shell which wget > /dev/null && echo 'wget -O -' || which curl)

PR #473, opened 2018-01-15
Use run-mode-hooks for major mode setups

YiLiu6240:
This patch replaces `run-hooks` with `run-mode-hooks` in major mode setups,
in accordance with the Emacs' major mode conventions
(https://www.gnu.org/software/emacs/manual/html_node/elisp/Major-Mode-Conventions.html#Major-Mode-Conventions).

### DIFF
--- a/lisp/ess-bugs-d.el
+++ b/lisp/ess-bugs-d.el
@@ -254,7 +254,7 @@
   (make-local-variable 'font-lock-defaults)
   (setq font-lock-defaults '(ess-bugs-font-lock-keywords nil t))
   (setq ess-language "S") ; mimic S for ess-smart-underscore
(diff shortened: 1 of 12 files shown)


lionel-:
Thanks for this! Could you please double check you didn't switch to run-mode-hook in minor modes? I'm not familiar with all these modes but `ess-noweb-font-lock-mode-hook` looks like a minor mode.

YiLiu6240:
Thanks for pointing this out. I reverted the changes in `ess-noweb-font-lock-mode` and `ess-noweb-mode`.

mmaechler:
Thank you, @YiLiu6240 , for the corrections.  This now looks good to me... i.e. all the remaining changed ones are major modes.

Hence I dare to merge..

PR #475, opened 2018-01-21
Add an Xref backend for R mode

### LINKED ISSUE #307: jump to function definition?
Hello, I wonder if there is a way to jump to and edit the function definition? very similar to `ess-dump-object-into-edit-buffer` but the editing needs to be applied to the original source file.

Thank you very much.

Y

atheriel:
These changes implement the `xref` API for R mode, providing a "jump to definition" implementation for functions. R stores the source code location of functions that are read into the interpreter (via `source` or similar), so I use that when it's available. Otherwise, the function definition is dumped into a buffer with the special major mode `ess-r-source-mode`, which is a read-only mode with the same font lock as the other R modes. It also decorates the header with the namespace and byte compilation info, if that's appropriate.

I've set the hooks to enable the backend by default, so `M-.` and `M-,` should work out of the box in Emacs 25 and later.

I've read the discussion of this topic in #307, and while this implementation is not exactly what was envisioned there I believe it's broadly consistent, and would close #307.

The one outstanding issue I'm having is that `(set-syntax-table ess-r-syntax-table)` does not seem to be working correctly. I'd welcome any hypothesis as to why.

Any further comments or questions are welcome.

### DIFF
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -98,23 +98,38 @@ because they were sent to the interpreter directly, or because
 they are loaded via byte-compiled packages."
   (when (ess-boolean-command (format "!base::is.null(utils::getSrcref(%s))\n"
                                      (ess-r-xref--get-symbol symbol)))
-    (let ((fname (string-trim
+    (let* ((fname (string-trim
+                   (ess-string-command
+                    (format "base::cat(utils::getSrcFilename(%s, full.names = TRUE)[1], \"\\n\")\n"
+                            (ess-r-xref--get-symbol symbol)))))
+           ;; Check if the file srcref is cached by ESS.
+           (srcref (gethash fname ess--srcrefs))
+           (ess-buff (when srcref (ess--dbg-find-buffer (car srcref))))
+           ;; R seems to keep track of lines even for non-existent files, too.
+           (line (string-to-number
                   (ess-string-command
-                   (format "base::cat(utils::getSrcFilename(%s, full.names = TRUE)[1], \"\\n\")\n"
-                           (ess-r-xref--get-symbol symbol)))))
-          (line (string-to-number
-                 (ess-string-command
-                  (format "base::cat(utils::getSrcLocation(%s, which = \"line\")[1], \"\\n\")\n"
-                          (ess-r-xref--get-symbol symbol))))))
-      ;; For now, require that the file exists.
-      ;; TODO: Handle srcfiles cached by ESS.
-      ;; TODO: Handle temp install directories.
-      (when (ess-boolean-command (format "base::file.exists(\"%s\")\n" fname))
+                   (format "base::cat(utils::getSrcLocation(%s, which = \"line\")[1], \"\\n\")\n"
+                           (ess-r-xref--get-symbol symbol))))))
+      (cond
+       ;; When the function was evaluated linewise into the interpreter, the
+       ;; filename will be empty. We can't do anything in this case.
+       ((string-equal fname "") nil)
+       ;; If the file exists, jump into it.
+       ((ess-boolean-command (format "base::file.exists(\"%s\")\n" fname))
         (list (xref-make (if (listp symbol)
                              (concat (car symbol) "::" (cadr symbol))
                            symbol)
                          (xref-make-file-location
-                          (expand-file-name fname) line 0)))))))
+                          (expand-file-name fname) line 0))))
+       ;; For functions eval'd into ESS, we can use the cached file position.
+       (ess-buff (list (xref-make
+                        (if (listp symbol)
+                            (concat (car symbol) "::" (cadr symbol))
+                          symbol)
+                        (xref-make-buffer-location ess-buff (caddr srcref)))))
+       ;; TODO: Handle other non-existing srcfiles, such as temp install
+       ;; directories.
+       (t nil))))) ;; We didn't find it.
 
 
 (provide 'ess-r-xref)
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -94,13 +94,66 @@ they are loaded via byte-compiled packages."
   (let ((cmd (format "cat(\"%s <- \"); print.function(%s)\n"
                      fn fn))
         (buff (get-buffer-create
-               (format "*definition[R]:%s*" fn))))
+               (format "*definition[R]:%s*" fn)))
+        (proc ess-local-process-name))
     (with-current-buffer (ess-command cmd buff)
-      ;; TODO: This should probably have its own bespoke mode, just using the
-      ;; syntax of `r-mode', so as to avoid pullin in all of ESS's hooks.
-      (r-mode)
-      (setq buffer-read-only t))
+      (ess-r-source-mode)
+      (setq-local ess-local-process-name proc))
     (list (xref-make (format "%s" fn) (xref-make-buffer-location buff 0)))))
+
+
+;;; R Function Source Viewer Mode
+
+(defun ess-r-xref--parse-byte-compiled-p (buff)
+  "Determine whether the R function printed in BUFF is byte-compiled.
+
+This also erases the metadata once the compilation status has
+been determined."
+  (with-current-buffer buff
+    (goto-char (point-max))
+    (when (re-search-backward "^<bytecode" nil 'noerror)
+      ;; Delete the metadata, if it's there.
+      (beginning-of-line)
+      (kill-whole-line)
+      t)))
+
+(defun ess-r-xref--parse-environment (buff)
+  "Determine the environment of the R function printed in BUFF.
+
+This also erases the metadata once the environment has been
+determined."
+  (with-current-buffer buff
+    (goto-char (point-max))
+    (let* ((found (re-search-backward "^<environment: \\(.*\\)>$" nil 'noerror))
+           (env (or (when found (match-string-no-properties 1)) ".GlobalEnv")))
+      ;; Delete the metadata, if it's there.
+      (when found (beginning-of-line) (kill-whole-line))
+      env)))
+
+;; Defined in ess-r-mode.
+(defvar ess-r-syntax-table)
+
+(defvar-local ess-r-source-byte-compiled-p nil
+  "Indicates whether the R function on display is byte-compiled.")
+
+(defvar-local ess-r-source-environment nil
+  "The environment the R function on display belongs to.")
+
+(define-derived-mode ess-r-source-mode special-mode "R Source"
+  "Major mode for navigating the source of local/byte-compiled R functions."
+  :syntax-table nil
+  (setq buffer-read-only nil)
+  ;; Use R's syntax table.
+  (set-syntax-table ess-r-syntax-table)
+  ;; Move metadata to the header line.
+  (let ((compiled (ess-r-xref--parse-byte-compiled-p (current-buffer)))
+        (env (ess-r-xref--parse-environment (current-buffer))))
+    (setq-local ess-r-source-byte-compiled-p compiled)
+    (setq-local ess-r-source-environment env)
+    (setq header-line-format
+          (concat "Environment: " env (when compiled " (byte-compiled)"))))
+  (add-hook 'xref-backend-functions #'ess-r-xref-backend nil 'local)
+  (setq buffer-read-only t))
 
 (provide 'ess-r-xref)
 
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -32,6 +32,9 @@
 (require 'xref)
 (require 'ess)
 
+;; Silence the byte compiler.
+(declare-function inferior-ess-r-force "ess-r-mode.el")
+
 
 ;;; Xref API
 
@@ -47,35 +50,24 @@
     (if (cdr parts) parts (car parts))))
 
 (cl-defmethod xref-backend-definitions ((_backend (eql ess-r)) symbol)
-  (ess-r-xref--check-for-process
-   ;; Ignore non-functions.
-   (when (ess-r-xref--fn-exists-p symbol)
-     (or (ess-r-xref--srcfile-xref symbol)
-         (ess-r-xref--body-xref symbol)))))
+  (inferior-ess-r-force)
+  ;; Ignore non-functions.
+  (when (ess-r-xref--fn-exists-p symbol)
+    (or (ess-r-xref--srcfile-xref symbol)
+        (ess-r-xref--body-xref symbol))))
 
 (cl-defmethod xref-backend-apropos ((_backend (eql ess-r)))
   ;; Not yet supported.
   nil)
 
 (cl-defmethod xref-backend-identifier-completion-table ((_backend (eql ess-r)))
-  (ess-r-xref--check-for-process
-   (let ((raw (ess-get-object-list ess-local-process-name)))
-     ;; This is not technically correct, since some of these objects are
-     ;; non-functions. But it *is* approximately correct, and expensive to
-     ;; filter with existing functions. A better strategy may be desirable in
-     ;; the future.
-     raw)))
-
-(defmacro ess-r-xref--check-for-process (&rest body)
-  "Wraps statements in BODY with `ess-make-buffer-current'."
-  `(if (not (ess-make-buffer-current))
-       (prog1 nil
-         ;; Give a message on the second invocation, as with
-         ;; `ess-complete-object-name'.
-         (when (string-match "xref" (symbol-name last-command))
-           (message "No ESS process associated with current buffer. Definitions \
-are not available.")))
-     ,@body))
+  (inferior-ess-r-force)
+  (let ((raw (ess-get-object-list ess-local-process-name)))
+    ;; This is not technically correct, since some of these objects are
+    ;; non-functions. But it *is* approximately correct, and expensive to
+    ;; filter with existing functions. A better strategy may be desirable in
+    ;; the future.
+    raw))
 
 
 ;;; Source File Locations
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -47,8 +47,10 @@
   (when (ess-r-xref--exists symbol)
     (if (ess-r-xref--has-srcfile symbol)
         (ess-r-xref--find-srcfile symbol)
-      ;; TODO: Implement non-srcfile references.
-      nil)))
+      (if (ess-r-xref--is-function symbol)
+          (ess-r-xref--find-body symbol)
+        ;; Ignore non-functions for now.
+        nil))))
 
 
 ;;; Source File Locations
@@ -79,6 +81,26 @@ they are loaded via byte-compiled packages."
                       (expand-file-name (string-trim file))
                       (string-to-number line)
                       0)))))
+
+
+;;; Local/Byte-compiled Locations
+
+(defun ess-r-xref--is-function (symbol)
+  "Check whether SYMBOL is an R symbol."
+  (ess-boolean-command (format "is.function(%s)\n" symbol)))
+
+(defun ess-r-xref--find-body (fn)
+  "Creates an xref to a buffer containing the body of the R function FN."
+  (let ((cmd (format "cat(\"%s <- \"); print.function(%s)\n"
+                     fn fn))
+        (buff (get-buffer-create
+               (format "*definition[R]:%s*" fn))))
+    (with-current-buffer (ess-command cmd buff)
+      ;; TODO: This should probably have its own bespoke mode, just using the
+      ;; syntax of `r-mode', so as to avoid pullin in all of ESS's hooks.
+      (r-mode)
+      (setq buffer-read-only t))
+    (list (xref-make (format "%s" fn) (xref-make-buffer-location buff 0)))))
 
 (provide 'ess-r-xref)
 
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -1,0 +1,85 @@
+;;; ess-r-xref.el --- An xref backend for R. -*- lexical-binding: t -*-
+;;
+;; Author: Aaron Jacobs
+;; Created: 21 January 2018
+;; Maintainer: ESS-core <ESS-core@r-project.org>
+;;
+;; Keywords: languages, statistics, xref
+;; Package-Requires: ((emacs "25"))
+;;
+;; This file is part of ESS.
+;;
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 2, or (at your option)
+;; any later version.
+;;
+;; This file is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; A copy of the GNU General Public License is available at
+;; http://www.r-project.org/Licenses/
+
+;;; Commentary:
+
+;; This file contains an xref backend for `R-mode'.
+
+;;; Code:
+
+(require 'subr-x)
+(require 'xref)
+(require 'ess)
+
+
+;;; Xref API
+
+(defun ess-r-xref-backend ()
+  "An `xref-backend-functions' implementation for `R-mode'."
+  'ess-r)
+
+(cl-defmethod xref-backend-identifier-at-point ((_backend (eql ess-r)))
+  (ess-symbol-at-point))
+
+(cl-defmethod xref-backend-definitions ((_backend (eql ess-r)) symbol)
+  (inferior-ess-r-force)
+  (when (ess-r-xref--exists symbol)
+    (if (ess-r-xref--has-srcfile symbol)
+        (ess-r-xref--find-srcfile symbol)
+      ;; TODO: Implement non-srcfile references.
+      nil)))
+
+
+;;; Source File Locations
+
+(defun ess-r-xref--exists (symbol)
+  "Check whether SYMBOL is a known R symbol."
+  (ess-boolean-command (format "exists(\"%s\")\n" symbol)))
+
+(defun ess-r-xref--has-srcfile (symbol)
+  "Checks that the R symbol SYMBOL has a valid source file reference.
+
+Most functions will not have source file references, either
+because they were sent to the interpreter directly, or because
+they are loaded via byte-compiled packages."
+  (ess-boolean-command
+   (format
+    "!is.null(utils::getSrcref(%s)) && file.exists(utils::getSrcFilename(%s))\n"
+    symbol symbol)))
+
+(defun ess-r-xref--find-srcfile (symbol)
+  "Creates an xref for the source file reference of R symbol SYMBOL."
+  (let ((file (ess-string-command
+               (format "cat(utils::getSrcFilename(%s)[1], \"\\n\")\n" symbol)))
+        (line (ess-string-command
+               (format "cat(utils::getSrcLocation(%s, which = \"line\")[1], \"\\n\")\n" symbol))))
+    (list (xref-make (format "%s" symbol)
+                     (xref-make-file-location
+                      (expand-file-name (string-trim file))
+                      (string-to-number line)
+                      0)))))
+
+(provide 'ess-r-xref)
+
+;;; ess-r-xref.el ends here
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -43,14 +43,37 @@
   (ess-symbol-at-point))
 
 (cl-defmethod xref-backend-definitions ((_backend (eql ess-r)) symbol)
-  (inferior-ess-r-force)
-  (when (ess-r-xref--exists symbol)
-    (if (ess-r-xref--has-srcfile symbol)
-        (ess-r-xref--find-srcfile symbol)
-      (if (ess-r-xref--is-function symbol)
-          (ess-r-xref--find-body symbol)
-        ;; Ignore non-functions for now.
-        nil))))
+  (ess-r-xref--check-for-process
+   (when (ess-r-xref--exists symbol)
+     (if (ess-r-xref--has-srcfile symbol)
+         (ess-r-xref--find-srcfile symbol)
+       ;; Ignore non-functions for now.
+       (when (ess-r-xref--is-function symbol)
+         (ess-r-xref--find-body symbol))))))
+
+(cl-defmethod xref-backend-apropos ((_backend (eql ess-r)))
+  ;; Not yet supported.
+  nil)
+
+(cl-defmethod xref-backend-identifier-completion-table ((_backend (eql ess-r)))
+  (ess-r-xref--check-for-process
+   (let ((raw (ess-get-object-list ess-local-process-name)))
+     ;; This is not technically correct, since some of these objects are
+     ;; non-functions. But it *is* approximately correct, and expensive to
+     ;; filter with existing functions. A better strategy may be desirable in
+     ;; the future.
+     raw)))
+
+(defmacro ess-r-xref--check-for-process (&rest body)
+  "Wraps statements in BODY with `ess-make-buffer-current'."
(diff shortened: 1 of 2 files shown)


lionel-:
woot awesome thanks! I'll try to take a look later this week

vspinu:
Great initiative. Thanks! 

I cannot try this. Getting:

```el
Debugger entered--Lisp error: (cl-no-applicable-method xref-backend-identifier-completion-table ess-r)
  signal(cl-no-applicable-method (xref-backend-identifier-completion-table ess-r))
  cl-no-applicable-method(#s(cl--generic :name xref-backend-identifier-completion-table :dispatches ((0 #s(cl--generic-generalizer :name cl--generic-eql-generalizer :priority 100 :tagcode-function #f(compiled-function (name &rest _) #<bytecode 0x25addf>) :specializers-function #f(compiled-function (tag &rest _) #<bytecode 0x25adff>)) #s(cl--generic-generalizer :name cl--generic-t-generalizer :priority 0 :tagcode-function #f(compiled-function (name &rest _) #<bytecode 0x259e2b>) :specializers-function #f(compiled-function (tag &rest _) #<bytecode 0x259e3f>)))) :method-table (#s(cl--generic-method :specializers ((eql etags)) :qualifiers nil :uses-cnm nil :function #f(compiled-function (backend) #<bytecode 0x12bff3d>)) #s(cl--generic-method :specializers ((eql elisp)) :qualifiers nil :uses-cnm nil :function #f(compiled-function (backend) #<bytecode 0x4b4f4d>))) :options nil) ess-r)
  apply(cl-no-applicable-method #s(cl--generic :name xref-backend-identifier-completion-table :dispatches ((0 #s(cl--generic-generalizer :name cl--generic-eql-generalizer :priority 100 :tagcode-function #f(compiled-function (name &rest _) #<bytecode 0x25addf>) :specializers-function #f(compiled-function (tag &rest _) #<bytecode 0x25adff>)) #s(cl--generic-generalizer :name cl--generic-t-generalizer :priority 0 :tagcode-function #f(compiled-function (name &rest _) #<bytecode 0x259e2b>) :specializers-function #f(compiled-function (tag &rest _) #<bytecode 0x259e3f>)))) :method-table (#s(cl--generic-method :specializers ((eql etags)) :qualifiers nil :uses-cnm nil :function #f(compiled-function (backend) #<bytecode 0x12bff3d>)) #s(cl--generic-method :specializers ((eql elisp)) :qualifiers nil :uses-cnm nil :function #f(compiled-function (backend) #<bytecode 0x4b4f4d>))) :options nil) ess-r)
  #f(compiled-function (&rest args) #<bytecode 0x17fbbe5>)(ess-r)
  apply(#f(compiled-function (&rest args) #<bytecode 0x17fbbe5>) ess-r nil)
  xref-backend-identifier-completion-table(ess-r)
  xref--read-identifier("Find definitions of: ")
```

Both with emacs 25 and 26.

lisp/ess-r-xref.el
```diff
@@ -0,0 +1,163 @@
...
+;; Defined in ess-r-mode.
+(defvar ess-r-syntax-table)
+
+(defvar-local ess-r-source-byte-compiled-p nil
+  "Indicates whether the R function on display is byte-compiled.")
+
+(defvar-local ess-r-source-environment nil
+  "The environment the R function on display belongs to.")
+
+(define-derived-mode ess-r-source-mode special-mode "R Source"
```
vspinu: Why is this mode needed? (I cannot test the code at the moment and I don't have a clear picture of what's going on.)

lisp/ess-r-xref.el
```diff
@@ -0,0 +1,163 @@
...

+;;; Source File Locations
+
+(defun ess-r-xref--exists (symbol)
+  "Check whether SYMBOL is a known R symbol."
+  (ess-boolean-command (format "exists(\"%s\")\n" symbol)))
+
+(defun ess-r-xref--has-srcfile (symbol)
+  "Checks that the R symbol SYMBOL has a valid source file reference.
+
```
vspinu: A small stylistic comment. 

I am personally not a big fan of the recent trend in elisp comunity of splitting inline docs into single-sentence paragraphs. Especially if it's about internal functions. Code files are for code and as a developer you read the code 100 times but the docs only once or twice. Such docs take loads of space and their usability is questionable. Code should speak for itself IMO.

In this case you could safely keep only the last paragraph with no loss. Everyone can see that you are checking for srcref by glancing at the code.

lisp/ess-r-xref.el
atheriel: Although it shares syntax with R buffers, it's not an editable buffer (since it represents the internal state of the R interpreter), and so it doesn't need any of the usual ESS apparatus (evaluation, completion, etc). I thought making a simple special mode was a good way to express this.

lisp/ess-r-xref.el
atheriel: Ok.

vspinu:
I don't know the internals of xref (it has been a while on my to do list :/) All I did was to call M-. on a function name from a library.

lisp/ess-r-xref.el
atheriel: Only if that package has been `source()`'d into a running interpreter. Packages that are byte-compiled or evaluated linewise into the interpreter do not have source files known to the R process, so the best we can do is dump out what R knows (i.e. print the function). If the function is not known to a running R process at all, the jump will simply fail, or fall back onto etags (or whatever else you have set in your `xref-backend-functions` list).

lisp/ess-r-xref.el
atheriel: Removed the extra paragraph.

atheriel:
@vspinu I've implemented `xref-backend-identifier-completion-table`. See if it works for you now.

lisp/ess-r-xref.el
mmaechler: @atheriel : Note that byte-compiled code will be come prevalent in R in the future.  *But* that byte-compiled code will still have source references (and has now).  So I hope you assume source references but you do*not* assume non-bytecompiled code.

vspinu:
Yes. It works! I think we can pick it up from here. There are a bunch of improvements that I would like to see before we can close #307:

 -  Make namespaced references like base::data.frame work
 -  Make xref work on code evaluated by ESS eval commands. We do track locations internally with all those `file.R!8kMTUf#2` (depending on the value of `ess-inject-source`). Thus, for most ess eval commands we can recover the files. I will fix this once this is merged. 
-  Make xref work on personal packages installed through CMD install and devtools (devtools at the least as I am pretty confident it tracks the src locations automatically). 
- I don't think R-source mode is useful. We should reuse our `ess-dump-object-into-edit-buffer` (C-c C-e C-d) and make it smarter by allowing namespaced evaluation within that buffer. That is, if I xref on `base::data.frame` -> modify -> eval then next time I xref on this symbol it pops up the current version of the code. This is useful for quick tweaking of the code which doesn't belong to you. 
- Maybe translate src references directly into packages located in `ess-tracebug-search-path`, but I am not sure this is a good idea as it might lead to confusioins with different versions of the package (e.g. if a package is installed from CRAN and my local package is from github).

AFAIC we can merge this right away and start tweaking on it within ESS. This patch consists of a separate ess-r-xref.el file and doesn't modify anything essential of the existing code. So it's safe in that regard.  

@atheriel would you like to add anything else for now?

lionel-:
> AFAIC we can merge this right away

I would like to take a look at it before, if only to discuss where to go from there.

vspinu:
Sure. No hurry.

lisp/ess-r-xref.el
atheriel: @mmaechler Byte-compiled functions do not appear to have the `srcref` attribute, at least in my own testing -- is there somewhere else the source location is stored?

lionel-: Only locally defined functions will have srcrefs. But my guess is that you don't make any distinction between bytecoded and non-bytecoded functions and just look at the srcref attribute? It would be hard to make the distinction anyway, they have the same type.

atheriel:
@vspinu Yes, I would like to get namespaced references working before this is merged -- it was a clear omission for a common use case. I have an idea about how I'd like to do it, as well.

I hadn't realized that the "file.R!8kMTUf#2" files had a meaning, they just looked to me like an idiosyncrasy of R while I was debugging -- and it's why there is a check that `file.exists()` in `ess-r-xref--has-srcref`. It's great to know that we can recover the location of evaluated functions this way, although I can't seem to figure out how by looking at the existing evaluation code -- am I missing something? Should I try to implement `srcref` manually in `.ess.source`?

Similarly I was unaware that **devtools** tracked source location, but I agree this would be another nice feature to have. It should work already, since these functions have a `srcref` attribute. Is anyone experiencing otherwise?

On the subject of a specialized R-source-mode, maybe I'm misunderstanding your desired workflow. One can't actually modify `base::data.frame`, so it doesn't make much sense to me for it to open in an editable buffer (as with `ess-dump-object-into-edit-buffer`, which to me seems much more suited to modifying values and data, not functions). Usually when I'm looking for a function definition in R it is because I want to know how a function in a package works, not because I want to modify it. So having an edit-focused buffer is undesirable.

lisp/ess-r-xref.el
atheriel: The distinction is made by parsing the output of `print.function()`. If there's a bytecode string at the bottom, it's considered byte-compiled, and otherwise it is not. I don't actually do anything with this information, though, other than put it in the header.

mmaechler:
It seems you are not aware of  `R CMD INSTALL --keep-source  <pkg>`   I would strongly hope `devtools` use these instead of reinventing the wheel.

lionel-:
> It seems you are not aware of R CMD INSTALL --with-keep.source <pkg> I would strongly hope devtools use these instead of reinventing the wheel.

I was not, thanks! You can supply INSTALL arguments with devtools but I think sources are not kept by default. devtools workflow is not based on installation though.

vspinu:
Devtools does that by default. It picks `getOption("keep.source.pkgs")` which is TRUE by default.

lionel-:
> It picks getOption("keep.source.pkgs") which is TRUE by default.

Hmm it's FALSE in my session...

atheriel:
I'm not sure it's that easy. `getOption("keep.source.pkgs")` is `FALSE` by default in my instance of R (3.4), but even if I change this and reinstall packages, their `srcref` attributes are not useful -- they point to temporary files removed after installation.

atheriel:
And on Windows, where binary packages are the default, even setting `keep.source.pkgs = TRUE` does not result in functions with `srcref` attributes.

vspinu:
>  can't seem to figure out how by looking at the existing evaluation code

The file associations are stored in the hash-map `ess--srcrefs` but I think all what you need here is to use `ess--dbg-goto-ref` function. Just supply the files which you pick from R's srcref and it should do the work. It also should do the right thing if file names are incomplete (uses `ess--dbg-find-buffer` internally).   

All of these are internals which I forgot the details of by now and those might need a bit of tweaking. But, by all means, if you are interested in digging and fixing this part yourself be my guest ;)

> One can't actually modify base::data.frame, 

You can. That's exactly the point of the namespaced evaluation in ESS which is activated inside packages. When you evaluate code inside an R package it is sourced directly into correct environments (package and namespace).  You can do the same with the edit buffer and the original intention was to tweak the edit buffer returned by `ess-dump-object-into-edit-buffer` to allow for that. We just didn't have long enough hands to do that just yet.

> when I'm looking for a function definition in R it is because I want to know how a function in a package works,

For that you just do `C-c C-e C-d` and you see the function definition. xref is not needed here at all. This is precisely why I am against xref dumping source files if the src hasn't been found. Such duplication of the existing functionality is confusing. I think in this case with xref we either should throw "no source location found" or delegate to ess-dump-object-into-edit-buffer.

> I'm not sure it's that easy. getOption("keep.source.pkgs") is FALSE

Argh. I am setting it myself in .Rprofile. Just tried with R -q but -q is the emacs thingy :/ Well, then we have to explicitly pass it withing R-package mode.  

> their srcref attributes are not useful -- they point to temporary files removed after installation.

They are useful, you just have to redirect temporary files into original directories. Ideally this mapping should be preserved across sessions somehow, but even if we search for packages in `ess-tracebug-search-path` and `ess-r-package-library-path` (@lionel- what is the intention of this variable?) it would be good enough. 

Also maybe there is a way to instruct devtools (or CMD install?) not to create temporaries. We need to dig into that.

atheriel:
>They are useful, you just have to redirect temporary files into original directories.

That's a good point. We could have some heuristics to look for temporary installation-like directories and replace them.

vspinu:
Yes. Heuristics and maybe exact mapping. When you call `ess-r-devtools-install-package` (and maybe ess-r-devtools-install-github?) the mapping could be saved. The trick is to save it across emacs sessions. We will have to manage our own .ess file in user-emacs-directory with such information.

Yet another improvement is to make this stuff work smoothly on remotes, but let's figure the local stuff first.

vspinu:
A lot of such heuristics are already in place in `ess--dbg-find-buffer`.

lionel-:
First of all thanks a lot for this contribution! It was clearly written with great care.

I'll start with a few comments:

- About `ess-r-xref--check-for-process`, it could probably be replaced with a call to `inferior-ess-r-force`. The current trend in ESS is to just start an inferior if there is no live one, switch to a single living inferior right away, or prompt the user when multiple live inferiors coexist. Also why did you make it a macro instead of a function? It doesn't install temporary side effects for `body` so could just be a function to call.

- When you call an R function make sure it is namespaced, even if it's a base function like `cat()`. Commands are run in the current process and we don't know what state it might be in. `cat` could be masked with another function.

- I agree with Vitalie that it's better to avoid docstrings for documenting internal functions. I prefer reserving docstrings to user documentation and using `;;` comments for notes aimed at the developer. Of course the function name and argument names should ideally speak for themselves.



Now onto the big picture. There are three main cases:


### No srcrefs

In this case I think we should delegate to alternative backends. This should be configurable with a list of backends. E.g. an "anywhere" xref backend, the etags backend, or an "edit in buffer" fallback (with `C-c C-e d` rather than the new `ess-r-source-mode`).

The Emacs authors mentioned on emacs-devel that a mode backend could fallback to the other backends if needed. Given the current EIEIO design I'm not sure what's the best way to accomplish this, maybe simply calling the backend methods directly. We can worry about that later.


### We have srcrefs

Let's use those to navigate to a relevant source file. We can distinguish two subcases:

- The srcref file exists. It means the package was directly installed from the sources, perhaps with `C-c C-w C-i`, or load-all'd with `C-c C-w C-l`. In this case we just use it as xref location.

- The srcref file does not exist, most probably because the package was installed from a temporary directory. For instance because it was a remote dependency downloaded on the fly or a package installed via `devtools::install_github()`.


#### Dealing with missing install directories with a source cache?

I see discussion about mapping from the temporary directory back to the source dirs as in tracebug temporary files. However I don't think we have this kind of situation here. Either we've installed from a permanent location and the srcrefs are correct, or we've installed from a temporary location and this location was lost.

What we can do however is to download the missing source tree manually and maintain a cache of package sources. `utils::packageDescription()` can tell us whether we have a package installed from CRAN or from github. If the latter, devtools keeps metadata on the relevant git SHA1 we should download. We'd download the source tree in our cache location once and then forward srcref-based xref lookups there.

What to do for packages with no srcrefs? E.g. packages installed with binaries on Windows and macOS. A simple solution would be to ask the user if they want to reinstall the package from source (the exact same version) from the ESS package cache location. We can pick up the relevant version to download from `utils::packageDescription()` (checking whether we indeed have a CRAN package and which version).

One issue with the source cache approach is invalidation. What if the user edits the files there? I don't think we need to detect that but we can provide a way for the user to reset a package source instead.



#### Preserving the srcrefs on installation

We do need `keep.source.pkgs` to be `TRUE` in order to get useful srcrefs. Perhaps it would make sense to set this option when the process is launched instead of requiring the user to set it? We need a way to disable this meddling though. Perhaps an `ess-r-managed-options` variable?

On a side note, I was thinking of managing `options("width")` automatically with `window-size-change-functions`. Users could remove `'width` from `ess-r-managed-options` if they don't want this.


### No function definition found

This case will arise when looking at functions imported in the namespace of a package. We could analyse the `NAMESPACE` file to figure out where we should look for a definition. We can worry about this later because `C-c C-w C-l` (e.g. `devtools::load_all()`) provides a work around.


### To sum up

Overall your patch is a great start, we just need to make sure `C-c C-w C-i` installs packages with their srcrefs. Then maybe the source cache approach would be a nice way of dealing with srcrefs pointing to temporary dirs. Then we can worry about falling back to other backends like etags or an edit buffer.

vspinu:
> What we can do however is to download the missing source tree manually and maintain a cache of package sources

This sounds a bit too complicated for the POC solution. I would rather bring this to a minimal functioning state first. Then if people (maybe you) have extra time to spare, they can push it forward for a full proof solution like the "managed cache".

> we should download.

How would you download? Using R or emacs? How would this work on remotes? Will you download to remote or local location? 

In case of github source we might not even download at all. The simplest solution is to have a backend that `browse-url`s into the web browser.

> We do need keep.source.pkgs to be TRUE in order to get useful srcrefs.

For C-c C-w i we should set it directly   n `ess-r-devtools-install-package`? 

> Perhaps an ess-r-managed-options variable?

Would be nice to have indeed. We hardwire options currently which is not so nice.

> Users could remove 'width from ess-r-managed-options if they don't want this.

Maybe have it as an alist of some default values e.g. ((width . auto) (pager . auto)). Some options might be manageable in several ways. User could then customize to a specific value which we inline with options(...) at the beginning of the session.


Back to xref. I would propose to start with several simple backends for now. Hold them all in `ess-r-xref-backends` variable and install as local `xref-backend-functions` within R files. For example:


 - ess-r-xref-srcref-backend (handles existent locations and tmp locations generated by ESS eval commands)
 
 - ess-r-xref-user-dir-backend (maps all temporary srcrefs to ess-r-package-library-path). This one is trivial; just map a tmp file to this path and see if the file exists. It would work nicely with remotes as long as users have a local copy of the package. Occasional mismatches are negligible and should be managed by the user (at least for now). 

 - ess-r-xref-github-browser-url-backend (works for github packages). Even for packages from CRAN we can check if URL contains link to github and use that.

 - ess-r-xref-dump-into-edit-buffer-backend


We can add more sophisticated backends later. For now, it's important to keep it simple and quickly start trying. There will surely be plenty of issues to solve. How those backends interact? How do user config affects the search (I set xref-prompt-for-identifier to t)? How remotes work? etc.

atheriel:
I've just modified to code to work with namespaced functions, and cleaned up a little. I'm not going to have the bandwidth to respond to all of your comments in one go, so I'll just give a partial answer/comment:

> About ess-r-xref--check-for-process, it could probably be replaced with a call to inferior-ess-r-force. The current trend in ESS is to just start an inferior if there is no live one, switch to a single living inferior right away, or prompt the user when multiple live inferiors coexist. Also why did you make it a macro instead of a function? It doesn't install temporary side effects for body so could just be a function to call.

I avoided `(inferior-ess-r-force)` in this case because I thought that xref should still be present (but nonfunctional, so far) in buffers that are not attached to a running R session. I modelled this behaviour after the approach in `ess-r-completion.el`, but am happy to change it if you think it makes more sense to be explict in the need to start an inferior process.

(Also, I made it a macro because it seemed intuitive to operate on symbols as opposed to values in this case, and so it would reduce to a simple `if` form when byte-compiled. But I'm not married to it in any case.)

> When you call an R function make sure it is namespaced, even if it's a base function like cat(). Commands are run in the current process and we don't know what state it might be in. cat could be masked with another function.

I tried to fix these in the latest commit. Please let me know if I missed one.

> I agree with Vitalie that it's better to avoid docstrings for documenting internal functions. I prefer reserving docstrings to user documentation and using ;; comments for notes aimed at the developer. Of course the function name and argument names should ideally speak for themselves.

I strongly disagree with the notion that internal functions should not use docstrings, and am quite sure that many/most internal functions contain them in the packages that are part of Emacs itself. But I will remove them if this is a condition for acceptance of these patches.

> The Emacs authors mentioned on emacs-devel that a mode backend could fallback to the other backends if needed. Given the current EIEIO design I'm not sure what's the best way to accomplish this, maybe simply calling the backend methods directly. We can worry about that later.

I agree that this is what should happen, and am not clear on why it doesn't work out of the box. Do you remember the thread where this discussion happened?

atheriel:
Note: it looks like the namespace changes break evil's `evil-goto-definition` interaction with xref, but I actually think this is a bug in evil, and this approach is the correct one. I'll file an issue there if this otherwise looks ok.

atheriel:
I've also noticed that `ess--srcrefs` caches the position of the function in a buffer, but not its line number. This makes it more difficult than first appears to find the original function location, given the way that the Xref API currently works. I'll keep digging, though.

lionel-:
> This sounds a bit too complicated for the POC solution.

Agreed, the cache (if the right solution) can and should come later. I'll open an issue about it.

> How would you download? Using R or emacs? How would this work on remotes? Will you download to remote or local location?

Using whatever is easiest at first. We should primarily aim for local boxes and support remote ones if we can. Also we might be able to rely on R packages for maintaning the cache. E.g. `crancache`.

> In case of github source we might not even download at all. The simplest solution is to have a backend that browse-urls into the web browser.

That's a good stopgap but xref is good for jumping around and discovering a new code base quickly. Github is not good at this. The current trend in R pkg development is to disaggregate packages into multiple ones (cf devtools diaspora). It's currently a pain to navigate those and would be awesome to have a transparent and reliable solution for jumping from source to source.

> Maybe have it as an alist of some default values e.g. ((width . auto) (pager . auto)).

Good idea.

> I would propose to start with several simple backends for now. Hold them all in ess-r-xref-backends variable and install as local xref-backend-functions within R files.

Sounds good. What do you think about Stefan's notion that we should use `add-function :before-until` instead of a `-functions` hook?

lionel-:
>  I modelled this behaviour after the approach in ess-r-completion.el, but am happy to change it if you think it makes more sense to be explict in the need to start an inferior process.
>
> (Also, I made it a macro because it seemed intuitive to operate on symbols as opposed to values in this case, and so it would reduce to a simple if form when byte-compiled. But I'm not married to it in any case.)

I think forcing is a bit better because if the user wants to attempt an action with R we should give them the means. I think automatically starting a process or linking to one works generally well. In any case I wouldn't make it a macro because it suggests you're installing then cleaning up side effects or evaluating the code with special control flow.

> I strongly disagree with the notion that internal functions should not use docstrings

Maybe we can change our minds later. The important thing is to clearly signal internal functions with `--` and you have done that very consistently, thanks!


> Do you remember the thread where this discussion happened?

https://lists.gnu.org/archive/html/emacs-devel/2015-04/msg01199.html

But I think this thread predates the EIEIO implementation. One complication will be that some backends will expect their own identifier (e.g. to install metadata in string properties), so we'll have to thread these methods together manually.

lionel-:
@atheriel If you just remove the ess-r-source-mode I think this is good to merge! I'm not sure if you should keep the environment getter and bytecode predicate lying around as dead code. They seem like they could be useful but at the same time we know where to find them if we need them.

If you'd like to implement the scaffholding for multiple internal xref backends (e.g. the edit-to-buffer one) that could come in another PR to keep the discussion focused.

vspinu:
@atheriel

> I modelled this behaviour after the approach in ess-r-completion.el

Auto completion and eldoc are not triggered by the user and it would be
intrusive to ask for process each time the user opens a new buffer. With xref is
different, the user triggers it and expects a results.

> I made it a macro

I think this macro wouldn't be need needed if you call ess-force-buffer-current first.

>  But I will remove them if this is a condition for acceptance of these patches.

Don't bother with it, it's a minor and is not worth your time. Especially that it's such a subjective matter. 

> ess--srcrefs caches the position of the function in a buffer, but not its line  number. This makes it more difficult than first appears to find the original   function location,

No need to deal with ess--srcrefs. You can call ess--dbg-create-ref-marker on file,
line, column and get the location from that marker.

BTW, would it make sense to treat these two as separate backends or it's a complication?

@lionel-

> What do you think about Stefan's notion that we should use add-function :before-until

AFAICS xref-backend-functions is designed for backends and that's probably what we should do as well.

I am no longer familiar with current emacs trends but experience from clojure ecosystem indicates that wrapping is not such a good idea after all. It creates monolithic, hard to change or customize constructs which don't play well at all with async code. We should allow users to easily customize their backends and I don't see how this could be done with add-functions, but I am might be missing something.

atheriel:
Ok, I've pushed changes that leave me happy enough with the state of this branch to merge it. I can rebase the commits if you'd like, but the current history seems clean enough to me. This is a long comment, but it sums up my outstanding issues on the topic, so it can be my last.

> Auto completion and eldoc are not triggered by the user and it would be
intrusive to ask for process each time the user opens a new buffer. With xref is
different, the user triggers it and expects a results.

This makes sense to me. It's now using `(inferior-ess-r-force)`.

> No need to deal with ess--srcrefs. You can call ess--dbg-create-ref-marker on file,
line, column and get the location from that marker.

I was able to use `ess--dbg-find-buffer` 👍. The backend is now able to jump to functions eval'd into the interpreter by ESS (provided they are still in the same place, but that's an issue for another time).

### Alternative/Multiple Backends

I spent some time going through the emacs-devel threads and experimenting with multiple backends. Basically, I now think that it's a bad idea to go down this route, so I'll contribute my thoughts even though I don't have a real stake in the final design. The current Xref API only allows for cascading between backends when calling the actual backend function itself (which can return `nil` instead of symbol). So for example we could modify this backend to only work when an inferior R process is present, e.g.

```emacs lisp
(defun ess-r-xref-backend ()
  "An `xref-backend-functions' implementation for `R-mode'."
  ;; For now, we require an inferior R process. Otherwise, return nil to
  ;; delegate to the next xref backend, likely etags.
  (if (ess-make-buffer-current)
      'ess-r
    (message "No ESS process associated with current buffer. Delegating to \
the next xref backend...")
    nil))
```

(I thought about including this snippet above but decided against it, following similar lines of thought around forcing an inferior process.)

The problem is I don't think we can do this further down the line where it would actually be useful, e.g. defering to the next backend if the symbol is not an R function, etc, because most of that context only appears much later, usually in `xref-backend-definitions`, which can't bail out by returning `nil`.

All of this makes me think that a single monolithic backend is likely the direction we will have to take, and my cursory research on other modes implementing the backend (Emacs lisp, elpy, and some others) shows that this is what they seem to be doing as well at present. On emacs-devel, the discussion generally drifted towards the idea that backends would need to be composed within another backend -- but since we don't have a compelling third-party backend we need to work with at the moment I don't really see why we'd do this over using configuration options to control various aspects of a monolithic one.

The final reason is that the only backend we can actually fall back on at the moment is etags, which users are unlikely to be using out of the box and therefore will likely get little value out of. I think for those users that do want to use etags, we can provide a configuration option, which is what the current Emacs lisp mode does.

### R Source Mode

> f you just remove the ess-r-source-mode I think this is good to merge!

I have removed this code, as you both requested. However, I'll make one final plea for its inclusion:

1. Without the ability to jump to definitions outside of what has been sourced/eval'd into the interpreter, this backend currently works on *very, very few actual functions* -- namely, the ones you probably wrote yourself. (In fact, it basically doesn't do anything for the R code I actually write in practice, which contains few functions.) The original version allowed for jumping around most of what a user would understand as "functions I'd like to see the source of", even if, as both of you have fairly pointed out, it trod upon the functionality of `ess-r-xref-dump-into-edit-buffer`. As @lionel-  said

  > xref is good for jumping around and discovering a new code base quickly

  ... and yet this is not currently possible, unless you're source-ing a lot more functions into your R session that I am.

2. I'm trying to understand why you would reject an existing, working implementation, which could clearly be improved when someone has time to upgrade `ess-r-xref-dump-into-edit-buffer` in the future. Many of the plans/ideas discussed in this thread, even though I generally agree with them, seem as though they should build on this functionality, not denigrate it. Even if the implementation is eventually completely swapped out for something based on `dump()`, it's basically just an implementation detail, and I don't see how it would affect users.

3. It's not clear to me that `ess-r-xref-dump-into-edit-buffer`, which is really an interface to `dump()`, not `print()`, is the functionality that is desirable anyway. As I've said, I'm mostly interested in read only buffers, which is a real strength of Emacs through `special-mode`, and the idea behind ESS's help buffers, after all. And `ess-r-xref-dump-into-edit-buffer` is more limited at present than printing the function into the interpeter, so I'm more likely to do that... and lose out on the ability to xref at all.

4. For what it's worth, `ess-r-source-mode` is basically what Rstudio does when it can't find a srcref.

vspinu:
> and therefore will likely get little value out of. 

I use it all the time for C code. Most of the time I have at least one tag table loaded. It's frustrating that emacs folks couldn't figure out the xref pipelining so far :(

> I have removed this code, as you both requested. 
> I'm trying to understand why you would reject an existing

I think the idea was to integrate it with `dump-into-edit-buffer`, preferably as a backend, not to remove it. The mode should have been removed, not the functionality. 

> t's not clear to me that ess-r-xref-dump-into-edit-buffer, which is really an interface to dump(), not print(), is the functionality that is desirable anyway

We are all in favor of improving it.  Automatic namespaced evaluation is one such improvement. Suggestions and patches are always welcome.

`dump` is better than `print` because you can edit an object (not only functions) eval the buffer and have a new in your environment. (This contradicts "source is real" philosophy but that's another discussion). 

 > I'm mostly interested in read only buffers, which is a real strength of Emacs through special-mode,

I don't see read-only buffers as "strength" in this case. It's inconsistent. When you xref to a real source file you don't make it read only, right? Then why would that be different with functions which we haven't found srcrefs of? Especially that we have the ability to eval it into the original namespace/package. 

>  And ess-r-xref-dump-into-edit-buffer is more limited at present than printing the function into the interpeter

Why is that more concretely? And you mean `ess-dump-object-into-edit-buffer` here, right?

vspinu:
In any case, green light from me. The patch is in a very good state.

> All of this makes me think that a single monolithic backend is likely the direction we will have to take,

Ok. Let's do that for starters then. Cascade all our backends within that `cond` and disaggregate them later if we find a way to do so.

lionel-:
> I'm trying to understand why you would reject an existing, working implementation

Because it's redundant with another facility that already exists in ESS and could use some improvement. I also think that it needs to be an edit buffer, not a viewing buffer. The viewing buffer would not bring much over just printing the function in the REPL which is the main way of discovering sources.


> It's not clear to me that ess-r-xref-dump-into-edit-buffer, which is really an interface to dump(), not print(), is the functionality that is desirable anyway

In the case of a function with xrefs available it should use `print()` rather than `dump()` to be truer to the source.


> All of this makes me think that a single monolithic backend is likely the direction we will have to take, 

I think we can still call into the etags backend if needed.

lionel-:
Ok let's try this out. Thanks for this @atheriel!

lionel-:
After 2 mins of use I can already tell you this is going to improve my workflow a lot! Being able to jump around a `devtools::load_all()`'d package seamlessly is huge. 👍

lionel-:
Hi @atheriel. Did you have a chance to look into the copyright assignment paperwork? This would really help us move forward the inclusion of ESS into ELPA, which will make it easier to install for new users. Thanks!

lionel-:
@atheriel Another friendly reminder that we need a copyright assignment to the FSF so that we can be included in Emacs Elpa.

At least, please let us know whether we need to rewrite your contribution to sort this out.

atheriel:
@lionel- Thanks for following up, and for your patience. I did submit the paperwork; hopefully I will appear on the list soon.

lionel-:
Awesome, thank you!

sje30:
hi @lionel- how is this functionality working out?  Has anyone compared it to something like dump-jump?  What should we as ESS be recommending?

lionel-:
The functionality has become an essential part of my workflow for package development. It requires installing packages from a local directory, and with srcrefs, which is the main limiting factor. I'd say that for their own packages, users should use xref-jump, and for foreign packages they should use dump-jump.

In the future it should become more useful. You can already set an envvar to instruct R to install all packages with srcrefs, but because they are installed from a temporary directory, you can't srcref-jump to the sources. It'd make sense for the `tools` package to gain a function that unrolls all the srcrefs and recreates a source directory in a temporary location. Once we have that, we can use that in ESS, RStudio, or vim to srcref-jump to any installed packages. I currently don't have time to work on this, but that'd be a great project for someone wishing to contribute to R.

PR #476, opened 2018-01-31
Remove cl

jabranham:
The cl package was obsoleted in Emacs 24.3. ESS currently requires 24.4+, so this shouldn't negatively affect anyone.


vspinu:
Thanks!

PR #485, opened 2018-02-07
Support for flymake in R-mode and Emacs 26+

jabranham:
Flymake got a major rewrite in Emacs 26 and is quite nice now. It
supports on-the-fly syntax checking for errors, warnings, and notes.
This commit adds support for this in ESS's R-mode using the lintr
package.

As of now, you'll need the development version of `lintr` because it allows passing a string to `lint()`. You'll also need the pretest version of Emacs (26.0.91 I believe is the latest, but anything built from the `emacs-26` branch should work).

### DIFF
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -22,13 +22,16 @@
 
 ;;; Commentary:
 ;; Flymake is the built-in Emacs package that supports on-the-fly
-;; syntax checking. This file adds support for this in R-mode by
+;; syntax checking.  This file adds support for this in R-mode by
 ;; relying on the lintr package, available on CRAN and currently
 ;; hosted at https://github.com/jimhester/lintr.
 
 ;; It is enabled by default.
 
 ;;; Code:
+
+(eval-when-compile
+  (require 'cl-lib))
 
 (defcustom ess-r-flymake-linters "default_linters"
   "Default linters to use.
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -30,6 +30,13 @@
 
 ;;; Code:
 
+(defcustom ess-r-flymake-linters "default_linters"
+  "Default linters to use.
+
+See \"lintr::with_defaults\" for how to customize this."
+  :group 'ess-R
+  :type 'string)
+
 (defvar-local ess-r--flymake-proc nil)
 
 (defun ess-r-flymake (report-fn &rest _args)
@@ -61,7 +68,7 @@ REPORT-FN is flymake's callback function."
                            ;; commandArgs(TRUE) lets us access
                            ;; everything after --args as a string:
                            "try(lint(commandArgs(TRUE)"
-                           ", " "linters = default_linters"
+                           ", " "linters = " ess-r-flymake-linters
                            "))"))
                    "--args" ,(eval
                               ;; text string of the current buffer:
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -37,6 +37,11 @@ See \"lintr::with_defaults\" for how to customize this."
   :group 'ess-R
   :type 'string)
 
+(defcustom ess-r-flymake-lintr-cache t
+  "If non-nil, cache lintr results."
+  :group 'ess-R
+  :type 'string)
+
 (defvar-local ess-r--flymake-proc nil)
 
 (defun ess-r-flymake (report-fn &rest _args)
@@ -69,6 +74,7 @@ REPORT-FN is flymake's callback function."
                            ;; everything after --args as a string:
                            "try(lint(commandArgs(TRUE)"
                            ", " "linters = " ess-r-flymake-linters
+                           (when ess-r-flymake-lintr-cache ", cache = TRUE")
                            "))"))
                    "--args" ,(eval
                               ;; text string of the current buffer:
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -133,6 +133,8 @@ REPORT-FN is flymake's callback function."
 
 (defun ess-r-setup-flymake ()
   "Setup flymake for ESS."
+  (when (< 26 emacs-major-version)
+    (error "ESS-flymake requires Emacs version 26 or later"))
   (add-hook 'flymake-diagnostic-functions #'ess-r-flymake nil t)
   ;; Try not to enable flymake if flycheck is already running:
   (unless (bound-and-true-p flycheck-mode)
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -32,6 +32,7 @@
 
 (eval-when-compile
   (require 'cl-lib))
+(require 'ess-custom)
 
 (defcustom ess-r-flymake-linters "default_linters"
   "Default linters to use.
@@ -48,12 +49,10 @@ See \"lintr::with_defaults\" for how to customize this."
 (defvar-local ess-r--flymake-proc nil)
 
 (defun ess-r-flymake (report-fn &rest _args)
-  "A Flymake backend for ESS-R modes. Relies on the lintr package.
-
+  "A Flymake backend for ESS-R modes.  Relies on the lintr package.
 REPORT-FN is flymake's callback function."
-  (unless (executable-find
-           ;; TODO - check whether lintr package is available
-           "R") (error "Cannot find a suitable R"))
+  (unless (executable-find "R")
+    (error "Cannot find a suitable R"))
   ;; If a live process launched in an earlier check was found, that
   ;; process is killed.  When that process's sentinel eventually runs,
   ;; it will notice its obsoletion, since it have since reset
@@ -70,7 +69,8 @@ REPORT-FN is flymake's callback function."
        (make-process
         :name "ess-r-flymake" :noquery t :connection-type 'pipe
         :buffer (generate-new-buffer " *ess-r-flymake*")
-        :command `("R" "--slave" "--restore" "--no-save" "-e"
+        :command `(,inferior-R-program-name
+                   "--slave" "--restore" "--no-save" "-e"
                    ,(eval (concat
                            "library(lintr);"
                            ;; commandArgs(TRUE) lets us access
@@ -107,7 +107,7 @@ REPORT-FN is flymake's callback function."
                                   ;; type
                                   (group-n 3 (| "style: " "warning: " "error: "))
                                   ;; msg
-                                  (group-n 4 (one-or-more not-newline)) "\n")
+                                  (group-n 4 (one-or-more not-newline)) line-end)
                               nil t)
                        for msg = (match-string 4)
                        for (beg . end) = (flymake-diag-region
@@ -115,9 +115,9 @@ REPORT-FN is flymake's callback function."
                                           (string-to-number (match-string 1))
                                           (string-to-number (match-string 2)))
                        for type = (let ((str (match-string 3)))
-                                    (cond ((string-equal str "error: ") :error)
-                                          ((string-equal str "warning: ") :warning)
-                                          ((string-equal str "style: ") :note)))
+                                    (cond ((string= str "error: ") :error)
+                                          ((string= str "warning: ") :warning)
+                                          ((string= str "style: ") :note)))
                        collect (flymake-make-diagnostic source
                                                         beg
                                                         end
@@ -135,10 +135,11 @@ REPORT-FN is flymake's callback function."
   "Setup flymake for ESS."
   (when (< 26 emacs-major-version)
     (error "ESS-flymake requires Emacs version 26 or later"))
-  (add-hook 'flymake-diagnostic-functions #'ess-r-flymake nil t)
-  ;; Try not to enable flymake if flycheck is already running:
-  (unless (bound-and-true-p flycheck-mode)
-    (flymake-mode)))
+  (when (string= "R" ess-dialect)
+    (add-hook 'flymake-diagnostic-functions #'ess-r-flymake nil t)
+    ;; Try not to enable flymake if flycheck is already running:
+    (unless (bound-and-true-p flycheck-mode)
+      (flymake-mode))))
 
 ;; Enable flymake in Emacs 26+
 (when (<= 26 emacs-major-version)
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -1,0 +1,131 @@
+;;; ess-r-flymake.el --- A ess-r Flymake backend  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2018 J. Alexander Branham (alex DOT branham AT gmail DOT com)
+
+;; This file is not part of GNU Emacs.
+
+;; This is free software; you can redistribute it and/or modify it under
+;; the terms of the GNU General Public License as published by the Free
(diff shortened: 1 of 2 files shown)


jabranham:
Currently, it's enabled by default on Emacs 26+. If you don't have lintr installed, or the `lint()` command fails for some reason, flymake does the smart thing and disables this backend.

izahn:
Will this interfere with Flycheck for those of us currently using that instead?

jabranham:
No, I've written it so that it doesn't turn on by default if flycheck is active already.

Even if you turn both on, they won't mess with each other. I guess you could get both flycheck and flymake faces applied to the same stuff, which might be visually unappealling.

nverno:
Could it be autoloaded instead of required in ess-r-mode? There is already so much overhead when starting ESS.

jabranham:
It could be, sure. I have it enabled by default for now since I think it's useful, and it's easier for people to disable features they dislike than find new features in a package they've used for a while. The additional overhead of requiring this is pretty minimal, especially compared to loading the rest of ESS. I'll wait to hear from the ESS maintainers and go with whatever they decide.

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+(defcustom ess-r-flymake-lintr-cache t
+  "If non-nil, cache lintr results."
+  :group 'ess-R
+  :type 'string)
+
+(defvar-local ess-r--flymake-proc nil)
+
+(defun ess-r-flymake (report-fn &rest _args)
+  "A Flymake backend for ESS-R modes. Relies on the lintr package.
+
```
vspinu: Could you please make the doc conform to emacs requirements (`M-x checkdoc`). We are aiming for clean docs for new code. Also please remove empty line. Such fragmentation of the code for the sake of one-line-paragraph is not very useful.

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+                           "try(lint(commandArgs(TRUE)"
+                           ", " "linters = " ess-r-flymake-linters
+                           (when ess-r-flymake-lintr-cache ", cache = TRUE")
+                           "))"))
+                   "--args" ,(eval
+                              ;; text string of the current buffer:
+                              (concat (buffer-substring-no-properties
+                                       (point-min) (point-max)))))
+        :sentinel
+        (lambda (proc _event)
```
vspinu: Would you mind getting this sentinel into a separate function?

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+
+(defvar-local ess-r--flymake-proc nil)
+
+(defun ess-r-flymake (report-fn &rest _args)
+  "A Flymake backend for ESS-R modes. Relies on the lintr package.
+
+REPORT-FN is flymake's callback function."
+  (unless (executable-find
+           ;; TODO - check whether lintr package is available
+           "R") (error "Cannot find a suitable R"))
```
vspinu: I am surprised emacs doens't give you warning here. It's a non standard formatting. Would you mind moving the `(error ..` on a separate line?

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+  (let ((source (current-buffer)))
+    (save-restriction
+      (widen)
+      (setq
+       ess-r--flymake-proc
+       (make-process
+        :name "ess-r-flymake" :noquery t :connection-type 'pipe
+        :buffer (generate-new-buffer " *ess-r-flymake*")
+        :command `("R" "--slave" "--restore" "--no-save" "-e"
+                   ,(eval (concat
```
vspinu: Eval is not needed here, nor quoting. You can just do `(list "R" ... (concat ...))`.

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+                       for msg = (match-string 4)
+                       for (beg . end) = (flymake-diag-region
+                                          source
+                                          (string-to-number (match-string 1))
+                                          (string-to-number (match-string 2)))
+                       for type = (let ((str (match-string 3)))
+                                    (cond ((string-equal str "error: ") :error)
+                                          ((string-equal str "warning: ") :warning)
+                                          ((string-equal str "style: ") :note)))
+                       collect (flymake-make-diagnostic source
```
vspinu: Why each argument on separate line? Looks rather ugly to me.

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+  (when (< 26 emacs-major-version)
+    (error "ESS-flymake requires Emacs version 26 or later"))
+  (add-hook 'flymake-diagnostic-functions #'ess-r-flymake nil t)
+  ;; Try not to enable flymake if flycheck is already running:
+  (unless (bound-and-true-p flycheck-mode)
+    (flymake-mode)))
+
+;; Enable flymake in Emacs 26+
+(when (<= 26 emacs-major-version)
+  (add-hook 'ess-mode-hook #'ess-r-setup-flymake))
```
vspinu: This will enable it in all ess-modes (sas, stata etc). Could you please check for ess-dialect being "R" within `ess-r-setup-flymake`.

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+  ;; narrowing restriction.
+  (let ((source (current-buffer)))
+    (save-restriction
+      (widen)
+      (setq
+       ess-r--flymake-proc
+       (make-process
+        :name "ess-r-flymake" :noquery t :connection-type 'pipe
+        :buffer (generate-new-buffer " *ess-r-flymake*")
+        :command `("R" "--slave" "--restore" "--no-save" "-e"
```
vspinu: `inferior-R-program-name` should be used here instead of "R".

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+                              ;; Regex to match the output lint() gives us.
+                              (rx line-start "<text>:"
+                                  ;; row
+                                  (group-n 1 (one-or-more num)) ":"
+                                  ;; column
+                                  (group-n 2 (one-or-more num)) ": "
+                                  ;; type
+                                  (group-n 3 (| "style: " "warning: " "error: "))
+                                  ;; msg
+                                  (group-n 4 (one-or-more not-newline)) "\n")
```
vspinu: Shouldn't it be `line-end` instead of "\n"?

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+                              nil t)
+                       for msg = (match-string 4)
+                       for (beg . end) = (flymake-diag-region
+                                          source
+                                          (string-to-number (match-string 1))
+                                          (string-to-number (match-string 2)))
+                       for type = (let ((str (match-string 3)))
+                                    (cond ((string-equal str "error: ") :error)
```
vspinu: `string=` is shorter and is more commonly used I believe.

lisp/ess-r-flymake.el
```diff
@@ -0,0 +1,149 @@
...
+              ;; check's output.
+              (kill-buffer (process-buffer proc))))))))))
+
+(defun ess-r-setup-flymake ()
+  "Setup flymake for ESS."
+  (when (< 26 emacs-major-version)
+    (error "ESS-flymake requires Emacs version 26 or later"))
+  (add-hook 'flymake-diagnostic-functions #'ess-r-flymake nil t)
+  ;; Try not to enable flymake if flycheck is already running:
+  (unless (bound-and-true-p flycheck-mode)
```
vspinu: A message might be useful here to inform people why flymake is not activated.

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: How do you want it formatted?

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: Done

lisp/ess-r-flymake.el
jabranham: I think we should leave it as-is, since that's how e.g. python.el and elisp.el do it.

jabranham:
@vspinu  I made most of the changes you requested, let me know what you think. Thanks for the review!

lisp/ess-r-flymake.el
jabranham: I can't get it to work without the eval/quoting. Can you?

lisp/ess-r-flymake.el
```diff
@@ -48,12 +49,10 @@ See \"lintr::with_defaults\" for how to customize this."
...
 
 (defun ess-r-flymake (report-fn &rest _args)
-  "A Flymake backend for ESS-R modes. Relies on the lintr package.
-
+  "A Flymake backend for ESS-R modes.  Relies on the lintr package.
 REPORT-FN is flymake's callback function."
-  (unless (executable-find
-           ;; TODO - check whether lintr package is available
-           "R") (error "Cannot find a suitable R"))
+  (unless (executable-find "R")
```
vspinu: `inferior-R-program-name` here?

lisp/ess-r-flymake.el
vspinu: Sure. I will fix it.

lisp/ess-r-flymake.el
vspinu: Frankly I couldn't care less how they do it. Most likely one of them copied from the other and that's not a good reason to imitate them. Having modularized code helps readability and debugging. Currently, if I want to debug that lambda I have to instrument all of the ess-r-flymake sausage.

vspinu:
I am getting `Disabling backend flymake-proc-legacy-flymake because (error Can’t find a suitable init function)` in emacs 27.0.5. Any ideas?

jabranham:
Yes, that's normal. It's saying the old legacy flymake backed isn't getting
enabled.

The new flymake can have multiple backends.

vspinu:
I see. The r backend does work but doesn't do much because it throws undetected during linting

```
WARNING: ignoring environment value of R_HOME
Loading required package: grDevices
Error in setwd(path) : cannot change working directory
```
I think that `try` is not very useful. We need a way to catch and inform the user that something went wrong and disable the flymake backend.

jabranham:
How are you getting that error, and where is it reported (what buffer)?

vspinu:
In the process buffer (`*ess-r-flymake*`). It comes from lintr directly. I am looking into it right now.

vspinu:
Well, `lint` expects a filename, but we are passing actual code. I am using 

```
 lintr     * 1.0.2   2017-11-08 CRAN (R 3.5.0)
```

and getting:

```R
lint("a+b")
Error in file(con, "r") : cannot open the connection
In addition: Warning message:
In file(con, "r") : cannot open file 'a+b': No such file or directory
```

jabranham:
Yes, you'll need the development version of lintr, which accepts a string.

vspinu:
Which dev version? I am seeing same thing wtih master:

```R
> devtools::session_info()
Session info ------------------------------------------------------------------
 setting  value                                             
 version  R Under development (unstable) (2018-01-30 r74185)
 system   x86_64, linux-gnu                                 
 ui       X11                                               
 language                                                   
 collate  en_GB.UTF-8                                       
 tz       Europe/Amsterdam                                  
 date     2018-02-11                                        

Packages ----------------------------------------------------------------------
 package   * version    date       source                          
 base      * 3.5.0      2018-01-31 local                           
 compiler    3.5.0      2018-01-31 local                           
 datasets  * 3.5.0      2018-01-31 local                           
 devtools    1.13.4     2017-11-09 CRAN (R 3.5.0)                  
 digest      0.6.15     2018-01-28 CRAN (R 3.5.0)                  
 graphics  * 3.5.0      2018-01-31 local                           
 grDevices * 3.5.0      2018-01-31 local                           
 lazyeval    0.2.1      2017-10-29 CRAN (R 3.5.0)                  
 lintr     * 1.0.1.9000 2018-02-11 Github (jimhester/lintr@e926089)
 magrittr    1.5        2014-11-22 CRAN (R 3.5.0)                  
 memoise     1.1.0      2017-04-21 CRAN (R 3.5.0)                  
 methods   * 3.5.0      2018-01-31 local                           
 Rcpp        0.12.15    2018-01-20 CRAN (R 3.5.0)                  
 rex         1.1.2      2017-10-19 CRAN (R 3.5.0)                  
 stats     * 3.5.0      2018-01-31 local                           
 tools       3.5.0      2018-01-31 local                           
 utils     * 3.5.0      2018-01-31 local                           
 withr       2.1.1      2017-12-19 CRAN (R 3.5.0)                  
 xml2        1.2.0.9000 2018-02-11 Github (hadley/xml2@a571b0f)    
> lint("a+b", linters = default_linters)
Error in file(con, "r") : cannot open the connection
In addition: Warning messages:
1: In normalizePath(filename) : path[1]="a+b": No such file or directory
2: In file(con, "r") : cannot open file 'a+b': No such file or directory
```

jabranham:
Ah, apparently `lint` requires the string to end with a newline character. https://github.com/jimhester/lintr/issues/234

We could ensure this always happens by appending a `\n` to the buffer contents. We'd have to check that the's not one already though because lintr would then warn about trailing newlines.

vspinu:
`(buffer-substring ...)` seem to always append \n to the string, so it's not an issue. 

I think the default linters are too intrusive and some are not fully featured. For instance roxygen blocks stating with ##'  are not recognized by commented_code_linter, infix_spaces_linter doesn't allow for a subset of infix operators, line_length_linter is too old school with 80 cols as a default, object_usage_linter almost always will fail within a package because it doesn't see objects defined in other source files etc 

So I would propose to leave a subset of linters by default.  But it's also not entirely clear to me how to streamline such customization from within emacs. Customizing the behavior of the lintr by a, potentially, very long string seems ugly to me. I have added the following:

```el
(defcustom ess-r-flymake-linters
  '("commented_code_linter = NULL"
    "single_quotes_linter = NULL"
    "infix_spaces_linter = NULL"
    "line_length_linter(80)"
    "object_usage_linter = NULL")
  "Default linters to use.
Can be either a string with R expression to be used as
is (e.g. 'lintr::default_linters'). Or a list of strings where
each element is passed as argument to 'lintr::with_defaults'."
  :group 'ess-R
  :type '(choice string (repeat string)))
```

Please let me know if you have other ideas.

jabranham:
Makes sense to me. I think defaulting `object_usage_lintr` off is especially a good idea, since it actually evaluates code, which could be unsafe.

Another way to allow for lintr customization would be to add a defcustom for each lintr, but that seems like a nightmare to maintain since there are a lot of them and they occasionally change names.

Issue #488, opened 2018-02-12
function definition is void: projectile-project-p


nverno:
In ess-inf, there are checks for `projectile-project-root` being bound, 
but not for `projectile-project-p`.  However, `projectile-project-root` is autoloaded
and latter isn't, leading to problems trying to start R.

nverno:
I just realized this was actually due to my autoloading `projectile-project-root` 
in my own code. But, it still seems strange to check for one function and then 
call another.

PR #489, opened 2018-02-12
check for projectile-project-p before calling it

nverno:
changed check from projectile-project-root to projectile-project-p
to match the subsequent function call #488


lisp/ess-inf.el
```diff
@@ -409,7 +409,7 @@ name. PROJECTILE-ROOT is directory name returned by
 `projectile-project-root' is undefined, or no project directory
```
vspinu: It should be changed in the docs as well.

vspinu:
Thanks!

PR #491, opened 2018-02-15
flymake: use a space before buffer name

jabranham:
This way it doesn't show up in e.g. list-buffers


vspinu:
That's intentional for debugging purposes before the release. I will make it invisible once all the details are figured out. 

I am seeing loads of *ess-r-flymake*<i> buffers created at the moment so something is not quite right yet.

PR #493, opened 2018-02-27
Improve flymake docs

jabranham:
- Add flymake node to texi file
- Add flymake announcement to doc/newfeat.texi
- Address M-x checkdoc complaints in ess-r-flymake.el

### DIFF
--- a/doc/ess.texi
+++ b/doc/ess.texi
@@ -2920,6 +2920,7 @@ ESS has a few extra features, which didn't fit anywhere else.
 @c todo: sort alphabetically
 @menu
 * ESS ElDoc::                   Display function arguments in minibuffer
+* ESS Flymake::                 On-the-fly syntax checking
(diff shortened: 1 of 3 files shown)


vspinu:
Perfect. Thanks!

vspinu:
BTW, I think .rlint config file is not picked up from the current dir or package root. Could you please check this up? 

I am not even sure what should be the right behavior here. Should .rlint overwrite ESS setup completely, or only those values which are explicitly listed in .rlint?

jabranham:
This is probably due to https://github.com/jimhester/lintr/issues/65

Maybe setting the 'default-directory' of the flymake buffer to the 'default-directory' of the source buffer would fix it?

jabranham:
Hrm... maybe not, since that buffer already shares 'default-directory' with the source buffer. I'll look into it (and open a separate issue).

PR #495, opened 2018-02-28
Add --quiet option to wget

jabranham:
Otherwise wget writes output (loading CA cert, etc) along with
julia-mode.el

Also, why does ESS handle julia-mode this way? It seems like it would be better to use Emacs's built-in way of declaring package dependencies.


rsparapa:
Looks OK to me provided -q is portable.  wget is not available by default on macOS so it should be fine there.

vspinu:
> Also, why does ESS handle julia-mode this way? 

Because ESS is provided through several options (including a few linux distros packages).  I think it would be good to everyone if we move to ELPA style packaging once the cleanup is complete.

PR #497, opened 2018-03-08
Improve face customization options

jabranham:
- Define ess- faces for (nearly) every face used to let users
  customize them for ESS

- Define a new group ess-faces for easier customization

- Remove 2 old defvars that aren't used anywhere else in the codebase

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2970,6 +2970,7 @@ the variable `ess-help-own-frame' is non-nil."
   "Font lock face used to highlight F and T."
   :group 'ess-faces)
 
+(defconst ess-matrix-face 'ess-matrix-face)
 (defface ess-matrix-face
   '((default (:inherit font-lock-constant-face)))
   "Font lock face used to highlight matrices."
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2966,7 +2966,7 @@ the variable `ess-help-own-frame' is non-nil."
 
 (defconst ess-f-t-face 'ess-f-t-face)
 (defface ess-f-t-face
-  '((default (:inherit font-lock-type-face)))
+  '((default (:inherit ess-constant-face)))
   "Font lock face used to highlight F and T."
   :group 'ess-faces)
 
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -136,6 +136,12 @@
   "Extra utilities for ESS"
   :group 'ess
   :prefix "ess-")
+
+(defgroup ess-faces nil
+  "Faces and face options for ESS modes."
+  :group 'ess
+  :prefix "ess-")
+
 ;; Variables (not user-changeable)
 
 (defvar ess-version "17.11" ;; updated by 'make'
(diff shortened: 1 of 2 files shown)


lisp/ess-custom.el
```diff
@@ -2896,45 +2901,87 @@ the variable `ess-help-own-frame' is non-nil."
...
+(defconst ess-function-call-face 'ess-function-call-face)
 (defface ess-function-call-face
   '((default (:slant normal :inherit font-lock-function-name-face)))
   "Font Lock face used to highlight function calls in ess buffers."
-  :group 'ess)
+  :group 'ess-faces)
 
+(defconst ess-numbers-face 'ess-numbers-face)
 (defface ess-numbers-face
   '((default (:slant normal :inherit font-lock-type-face)))
```
vspinu: Why change `slant`?

lisp/ess-custom.el
```diff
@@ -2896,45 +2901,87 @@ the variable `ess-help-own-frame' is non-nil."
...
+(defconst ess-operator-face 'ess-operator-face)
+(defface ess-operator-face
+  '((default (:inherit font-lock-constant-face)))
+  "Font lock face used to highlight operators."
+  :group 'ess-faces)
+
+(defconst ess-modifiers-face 'ess-modifiers-face)
+(defface ess-modifiers-face
+  '((default (:inherit font-lock-constant-face)))
+  "Font lock face used to highlight modifiers."
```
vspinu: Could you please add examples of modifiers here. Examples would be also useful in other non-obvious places (constants, operators, keywords etc).

lisp/ess-custom.el
```diff
@@ -2896,45 +2901,87 @@ the variable `ess-help-own-frame' is non-nil."
...
+
+(defconst ess-f-t-face 'ess-f-t-face)
+(defface ess-f-t-face
+  '((default (:inherit font-lock-type-face)))
+  "Font lock face used to highlight F and T."
+  :group 'ess-faces)
+
+(defface ess-matrix-face
+  '((default (:inherit font-lock-constant-face)))
+  "Font lock face used to highlight matrices."
```
vspinu: I think this doc is imprecise. We probably highlight only row numbers. Could you please have a look?

lisp/ess-custom.el
vspinu: Orh. Nevermind, that's our default for whatever reason. You didn't introduce any change here.

lisp/ess-custom.el
jabranham: On my end, it highlights row and column labels when printing matrix output (that look like [1,] or [,1]) or the labels on vectors [1] when printing vector output. I've updated the doc so it's hopefully clearer.

lisp/ess-custom.el
jabranham: Done, let me know if you think I missed anything non-obvious

jabranham:
I think I addressed everything, @vspinu. Thanks for the review!

vspinu:
Perfect. Thanks for all the work!

PR #498, opened 2018-03-09
Look for xref symbols within the namespace of current package

lionel-:
This way:

* we can find non-exported functions
* we can find functions from other packages imported to the namespace
* there is no need to attach the package to the search path

This should make xref much more useful to ESS users allergic to `devtools::load_all()` ;)

cc @atheriel

### DIFF
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -65,11 +65,16 @@ srcrefs point to temporary locations."
 
 (defun ess-r-xref--srcref (symbol)
   (inferior-ess-r-force)
-  (with-current-buffer (ess-command (format ".ess_srcref(\"%s\")\n" symbol))
-    (goto-char (point-min))
-    (when (re-search-forward "(" nil 'noerror)
-      (goto-char (match-beginning 0))
-      (read (current-buffer)))))
+  ;; Look for `symbol' inside the package namespace
+  (let* ((pkg (ess-r-package-name))
+         (pkg (if pkg
+                  (concat "\"" pkg "\"")
+                "NULL")))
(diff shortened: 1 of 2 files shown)


lisp/ess-r-xref.el
```diff
@@ -65,11 +65,16 @@ srcrefs point to temporary locations."
...
 (defun ess-r-xref--srcref (symbol)
   (inferior-ess-r-force)
-  (with-current-buffer (ess-command (format ".ess_srcref(\"%s\")\n" symbol))
-    (goto-char (point-min))
-    (when (re-search-forward "(" nil 'noerror)
-      (goto-char (match-beginning 0))
-      (read (current-buffer)))))
+  ;; Look for `symbol' inside the package namespace
+  (let* ((pkg (ess-r-package-name))
+         (pkg (if pkg
```
vspinu: These 3 lines in principle could be avoided by having `if (pkg != "nil" ...` in the R code.

lionel-: It doesn't feel right to pass `NULL` as a string though.

vspinu: I meant to pas "nil" not NULL; that is, remove the if and have  `(format "...\"%s\" pkg)` so when pkg is nil, "nil" is gotten on the R side. Either way is fine but it feels to me that if we have to make assumptions then better make them on elisp side not on the R side. This NULL passing is pretty common and could safe more trivial lines like these and on R side it doesn't cost much.

lionel-: That's what I meant by passing NULL as a string. I'd prefer to avoid stringly typing like this and the ambiguity it brings, maybe if it's a common pattern we could have a helper.

vspinu: Ok then. Let someone have their "nil" package in peace.

PR #500, opened 2018-03-09
[WIP] lexical binding

jabranham:
*WIP Do not merge*

The good:
- lexical scoping should result in multiple [performance](http://nullprogram.com/blog/2016/12/22/) [advantages,](http://nullprogram.com/blog/2017/01/30/) and [easier to read code](https://www.emacswiki.org/emacs/DynamicBindingVsLexicalBinding#toc5). Dynamic binding is still available through the use of so-called special variables, so we don't lose anything by switching to lexical-binding.

- I'm also providing some documentation improvements, mostly by satisfying M-x checkdoc as I go along.

The bad:
- Some things rely on dynamic binding, meaning they break. :-(
- I've mostly been testing `R-mode`, because that's what I use most often

The not-so-bad
- Most things don't break :-)

So far what I know is broken:

- [ ] fill-paragraph advice in R-mode:
```
c(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18)
```

`(fill-paragraph)` results in:

```
c(0,
  1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18)
```

- [ ] indentation tests fail

### DIFF
--- a/lisp/ess-comp.el
+++ b/lisp/ess-comp.el
@@ -1,4 +1,4 @@
-;;; ess-comp.el --- setting for compiling, only.
+;;; ess-comp.el --- setting for compiling, only.  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 1997--2006 A. J. Rossini
 ;; Copyright (C) 1997--2006 A.J. Rossini, Richard M. Heiberger, Martin
--- a/lisp/ess-compat.el
+++ b/lisp/ess-compat.el
@@ -1,4 +1,4 @@
-;;; ess-compat.el --- simple determination of Emacs and version #.
+;;; ess-compat.el --- simple determination of Emacs and version #.  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2000--2017 A.J. Rossini, Richard M. Heiberger, Martin
 ;;      Maechler, Kurt Hornik, Rodney Sparapani, and Stephen Eglen.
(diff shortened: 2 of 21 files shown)


vspinu:
It's hard to judge the implication of this change. I would stick with new code being lexicaly scoped and leave the old code alone. Old code is so stuffy that it's bound to break on this change. A lot of code we are not routinely testing, so you might get issues popping years later.

lionel-:
Agreed, ESS does not have sufficient test coverage to apply large changes of this nature confidently.

vspinu:
@jabranham would you mind reducing the scope of this change a little bit and use lexical binding only for a handful of new components (notably files starting with `ess-r-...`).

jabranham:
@vspinu sure thing, I've done that and re-pushed the changes. There are two commits. The files in the second I was less sure you wanted included. Let me know if any of that looks wrong.

lionel-:
have you run `make all` in the test folder? I'm pretty sure this breaks indentation and other stuff.

lionel-:
I would be more comfortable if the maintainers of each file were to make this change.

jabranham:
Yes, and yes it does. That's why I said WIP don't merge. :-) 

I was trying to gauge interest in switching before fixing all the various little issues.

lionel-:
I would rather fix `ess-r-syntax.el` myself if you don't mind.

jabranham:
Sure thing, I've pushed a commit undoing changes in that file.

This doesn't fix any of the issues previously mentioned about fill-paragraph or indentation, though, I believe.

lionel-:
Thanks, I'll send a PR to your branch when I find some time to look into it.

PR #502, opened 2018-03-12
company backend for library name completion

nverno:
When added to company-backends, eg. 
`(company-R-args company-R-objects company-R-library :separate)`
installed library names are offered as completion. I don't know how if this
is the proper way to cache objects in ESS? It appears as though things are
cached in properties of the process.

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -391,7 +391,7 @@ fill=TRUE); try(traceback(), silent=TRUE)})\n")
 ;; Precedes R4 and allows spaces in file path
 (add-to-list 'compilation-error-regexp-alist-alist
              ;; Starts at bol or with ": " (patterns 3,4,5,6,9)
-             '(R3 "\\(?:^ *\\|: ?\\)\\([^-+[:digit:] \n]:?[^: ]*\\):\\([0-9]+\\):\\(?:\\([0-9]+\\):\\)?"  1 2 3 2 1))
+             '(R3 "\\(?:^ *\\|: ?\\)\\([^-+[:digit:] \t\n]:?[^: \t\n]*\\):\\([0-9]+\\):\\(?:\\([0-9]+\\):\\)?"  1 2 3 2 1))
 
 (add-to-list 'compilation-error-regexp-alist-alist
              ;; Don't start with digit; no spaces


lisp/ess-r-completion.el
```diff
@@ -367,6 +367,29 @@ To be used instead of ESS' completion engine for R versions >= 2.7.0."
...
+;; completion for library names -- only active within 'library(...)'
+(defun company-R-library (command &optional arg &rest ignored)
+  (interactive (list 'interactive))
+  (cl-case command
+    (interactive (company-begin-backend 'company-R-library))
+    (prefix (and (string= "library" (car-safe (ess--fn-name-start 'symbol)))
+              (let ((start (ess-symbol-start)))
+                (and start (buffer-substring start (point))))))
+    (candidates (all-completions arg (company-R-library-all-completions)))
+    (annotation "<library>")
```
vspinu: I think it should be `<package>`

lisp/ess-r-completion.el
```diff
@@ -367,6 +367,29 @@ To be used instead of ESS' completion engine for R versions >= 2.7.0."
 ;;; COMPANY
 ;;; http://company-mode.github.io/
 
+(defvar company-R-library--cache ())
+(defun company-R-library-all-completions ()
+  (unless company-R-library--cache
```
vspinu: More idiomatically would be `(or ...)`.

lisp/ess-r-completion.el
```diff
@@ -367,6 +367,29 @@ To be used instead of ESS' completion engine for R versions >= 2.7.0."
...
+    (let ((proc (ess-get-next-available-process)))
+      (when proc
+        (setq company-R-library--cache
+              (ess-get-words-from-vector
+               "local({ out <- try({rownames(installed.packages())}); print(out, max=1e6) })\n")))))
```
vspinu: I think this is very slow on platforms with huge number of libraries, especially if lib paths are located on network drives. This is the case for @mmaechler  for example.

lisp/ess-r-completion.el
```diff
@@ -367,6 +367,29 @@ To be used instead of ESS' completion engine for R versions >= 2.7.0."
 ;;; COMPANY
 ;;; http://company-mode.github.io/
 
+(defvar company-R-library--cache ())
```
vspinu: I think `installed.packages` has a built-in cache mechanism.

There is also `.packages` function. I wonder which one would be better for this purpose.

nverno: How would you check for non-local libPaths. What would they look liike? I'm sure there is a function, but I can't find documentation on it.

vspinu: I don't actually know and I am pretty sure those would look differently depending on the system or network path type. 

Let's go with this as it and see how it goes. We can add this to the default backends and see if people complain. If this change causes serious slowdown even on local machines then, I am afraid, we will have to keep it outside of default backends.

lisp/ess-r-completion.el
```diff
@@ -367,6 +367,28 @@ To be used instead of ESS' completion engine for R versions >= 2.7.0."
...
+(defun company-R-library-all-completions ()
+  (or company-R-library--cache
```
vspinu: We don't need this one on emacs side if `installed.packages` uses internal cache. Have you checked that out?

lisp/ess-r-completion.el
nverno: even if you had every package on cran it still seems like it wouldn't be slow

lisp/ess-r-completion.el
nverno: only as far as reading its documentation. I figured it would still be faster to cache in lisp since its tiny memory, but could remove if you want.

vspinu: The issue is that package list can change over time and different processes could end up having different lists (think of remotes). Even on the same machine different R processes could end up with different package list for a number of reasons.

nverno: good point, I will remove the cache

PR #509, opened 2018-03-25
company backend for library name completion

nverno:
just fixed up from previous pull request


lionel-:
It's alright to force-push on your own pull request rather than open another one ;). This way the PR keeps a reference to the discussions.

nverno:
yea my bad, I don't know git well enough to squash together three commits that are separated by others.

vspinu:
Ok. Let's go with this. I don't mind adding it by default to the list of backends. Few people who have issues can always disable it.

PR #510, opened 2018-03-29
Fix several defcustom :type statements

jabranham:
This makes their defaults not be type mismatches

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -565,7 +565,7 @@ lines or reboot emacs. The first call clears the default
 `ess-smart-S-assign' assignment and the second line re-assigns
 it to the customized setting. "
   :group 'ess-S
-  :type 'character)
+  :type 'string)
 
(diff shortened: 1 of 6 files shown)


lisp/ess-custom.el
```diff
@@ -552,7 +552,7 @@ buffer or end chunks etc.")
   :group 'ess-S
   :type 'string)
 
-(defcustom ess-smart-S-assign-key "_"
+(defcustom ess-smart-S-assign-key ?_
```
vspinu: Does the new value actually works?  All changes to the defaults should be checked. I think that in this case it would be more reasonable to do an `(or string character)` as people might have already had customized this.

Any ideas why travis fails?

jabranham:
@vspinu I tested it locally but forgot I had overridden the smart assign key. Fixed, thanks

vspinu:
Thank you! I wanted to do this myself for ages.

PR #513, opened 2018-04-02
dir-locals.el: Setup bug-reference variables

jabranham:
If you use bug-reference-mode (which magit automatically turns on, I think), this will make links from text like "Merge pull request #510" that will open up the appropriate PR/issue.


lionel-:
Thanks!

Issue #506, opened 2018-03-17
Fix compilation warnings


lionel-:

jabranham:
Closing this as the `next` branch should byte compile on the most recent Emacs major release without warnings. Travis now errors on warnings with Emacs 26.

PR #515, opened 2018-04-11
Eliminate some byte compile warnings

jabranham:
I saw in #506 that yall were interested in eliminating some byte compile warnings. This is a start. It eliminates the easiest ones. Before these commits, I see 169 byte compile warnings. After this, there are still 94 on my machine.

There are a lot of commits, so let me know if you'd rather I squash them together into just one. 

There are a lot of references to functions that start with `w32-`, but these aren't defined anywhere for me and I couldn't find info about them online. Does anyone know what these are from?


mmaechler:
Thank you very much for your work.

Did make sure that you are  **not** requiring the very latest version(s) of emacs?
I know that some compiler warnings are because of  deprecated / outdated  emacs lisp functions.
Using the new way may disable the use of old versions of emacs.
In the past we have had quite a few "company users" who have not had much control about  updating the version of emacs they have to use at work.   It would be a pity if the elimination of such deprecation warnings would disable using ESS with older emacsen -- as they are e.g., in  Ubuntu LTS releases.

jabranham:
I tried very hard not to. I think none of the changes I made require Emacs > 24.3, which is what ESS requires (according to requires.texi, anyway). 

Actually, now that I think about it, removing `(ido-init-completion-maps)` might matter for people on older versions. I'll go back through and check what version of Emacs aliases that to `(ignore)`.

jabranham:
Looks like that happened in 25.1 so I've reverted that change. As far as I know, everything else should be fine from 24.3 onward.

lisp/ess-bugs-d.el
```diff
@@ -26,6 +26,8 @@
 (require 'ess-bugs-l)
 (require 'ess-utils)
 (require 'ess-inf)
+(require 'ess-custom)
+(declare-function ess-setq-vars-local "ess-mode")
```
lionel-: Should this be `(require 'ess-mode)`?

lisp/ess-comp.el
```diff
@@ -37,6 +37,8 @@
 (add-to-list 'load-path nil)
 
 ;; defvar'ed to nil in ./ess-site.el
+;; TODO: ^^^^ This isn't true?
```
lionel-: ESS used to depend on ess-site but I've been trying to remove that dependency. The defvar here looks like the right thing to do.

@mmaechler Could we make the value of `ess-show-load-messages` during compilation dependent on an environment variable? I think those load messages are distracting.

lisp/ess-inf.el
```diff
@@ -46,6 +46,8 @@
 (require 'format-spec)
 (require 'ess-tracebug)
 
+(declare-function tramp-sh-handle-expand-file-name "tramp-sh")
```
lionel-: Should there be instead `(require 'tramp-sh)` inside of `ess-set-working-directory`?

lisp/ess-julia.el
```diff
@@ -46,6 +46,10 @@
 (require 'ess-utils)
 (require 'ess-r-mode)
 
+(defvar ac-prefix)
```
lionel-: Probably better to have `(require auto-complete)` inside of `ess-ac-julia-objects`.

lionel-: It is probably be a good thing to explicitly require packages inside the commands/functions that use them (ideally at entry points and not in all depended-on functions), but of course this won't actually help with the compilation warning. So these defvar/declare-function declarations are correct!

lisp/ess-r-mode.el
```diff
@@ -51,6 +51,7 @@
 (autoload 'ess-r-args-show      "ess-r-args" "(Autoload)" t)
 (autoload 'ess-r-args-auto-show "ess-r-args" "(Autoload)" t)
 
+(declare-function ess-sqpe-versions-create "ess-sp6w-d")
```
lionel-: Could you move this inside `ess-r-s-versions-creation` please.

lisp/ess-r-syntax.el
```diff
@@ -358,8 +358,7 @@ reached."
       (ess-token-after= '("," ";"))
       (and (ess-token-after= "identifier")
            (not (memq (char-syntax (char-before)) '(?w ?_))))
-      (progn (/= (skip-syntax-backward ".") 0)
-             (ess-token-operator-p (ess-token-after)))
+      (ess-token-operator-p (ess-token-after))
```
lionel-: What is the purpose here? I don't see a compilation warning.

lisp/ess-bugs-d.el
jabranham: Sure, done

lisp/ess-inf.el
jabranham: I don't think it is, since tramp is built in with Emacs we can assume it's available, unlike auto-complete. I've removed it and added `(require 'tramp-sh)` in that function.

lisp/ess-julia.el
jabranham: Done

lisp/ess-noweb-mode.el
```diff
@@ -89,8 +89,11 @@
 
 ;; Want to use these now in order to cater for all obscure kinds of emacsen
 (require 'ess-compat)
+(declare-function html-mode "sgml-mode")
 (autoload 'ess-write-to-dribble-buffer "ess")
 
+(defvar weave-process)
```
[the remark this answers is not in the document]
jabranham: It's used in `ess-noweb-weave`, but I don't see any place it's set either. I wonder if that function even works? I don't use sweave myself.

lisp/ess-noweb-mode.el
```diff
@@ -89,8 +89,11 @@
 
 ;; Want to use these now in order to cater for all obscure kinds of emacsen
 (require 'ess-compat)
+(declare-function html-mode "sgml-mode")
```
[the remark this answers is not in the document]
jabranham: If we don't put this here, the byte compiler complains about html-mode being undefined.

lisp/ess-r-mode.el
jabranham: Done

lisp/ess-r-syntax.el
jabranham: Without this change, I see:

```
ess-r-syntax.el:346:8:Warning: value returned from (/= (skip-syntax-backward
                                                        ".") 0) is unused
```

jabranham:
Thanks for the review, @lionel-. I think I responded to everything; let me know if I missed something. 

No one responded to my question about the `w32-` functions. These functions actually produce errors for me when using ESS. For example, trying to open BUGS or JAGS files results in 

```
(void-function w32-shell-dos-semantics)
  w32-shell-dos-semantics()
  ess-jags-mode()
```

I'm assuming those functions have something to do with the way windows handles shells, but I can't find any info about them online. Anyway, I'll open a separate issue about that.

lisp/ess-r-syntax.el
lionel-: oops of course, I wonder why I didn't see any warning.

Could you keep the `skip-syntax-backward` though? Not sure if it is needed (no tests seem to have broken) but better be safe.

lisp/ess-noweb-mode.el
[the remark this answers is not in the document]
lionel-: Even with the change in line 203? I don't see any other reference to `html-mode`.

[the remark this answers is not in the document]
jabranham: Yes, you're right of course. I misremembered and was thinking that it called html-mode but it just checks the value. Done.

lisp/ess-r-syntax.el
jabranham: Done

lionel-:
I see this function in https://github.com/emacs-mirror/emacs/blob/e0f18aa07fb900c1bb0fe25386336fd6a73c9b0d/lisp/w32-fns.el

But if I try `(require 'w32-fns)` I get an error. You will find lots of ESS is broken because it is more than 20 years old and we only started writing tests in 2015... So thanks again for helping making things better!

jabranham:
Ah, great. Hopefully that means I can just add some `(require)` statements and all will be good. I'll check that out quickly.

rsparapa:
What is the error?  (require 'w32-fns) should still work.  Although, admittedly, I don't care about Windows support and I don't know who does.

jabranham:
For me, it's (void-function set-message-beep). Doesn't seem related to ESS as I encounter it even with `emacs -q`. 

I don't really care about windows support either, and am currently working on silencing the byte compile warnings about these being undefined by just checking that they're defined first. Should have a commit to push soon.

jabranham:
OK, just pushed a commit that does away with all the `w32-` errors except those in `ess-sp6w-d.el` where it's a little trickier. 

This also closes #518 as BUGS/JAGS seem to work now

jabranham:
Just pushed a commit that fixes `ess-R-display-vignettes` on remote systems. 

Byte compile warning count: 87

lionel-:
> I have to admit though that I don't totally understand how containing-sexp gets set.

By dynamic scoping, this will require refactoring in the longer term. I think a `defvar` is reasonable for now.

jabranham:
@rsparapa  - OK, great, I'll just remove the function then. I've never used SAS so didn't know it wasn't on macs anymore.

@lionel- `defvar`s make the byte compiler complain about non-prefixed variables. OK to rename to `ess-containing-sexp`, `ess-indent-point`, etc?

lionel-:
> OK to rename to ess-containing-sexp, ess-indent-point, etc?

Hmm that's going to make lots of commit noise for something that needs to be refactored anyway.

Could you try `(setq-local byte-compile-warnings '(not lexical)` please?

jabranham:
Great, done. Down to just 25 compiler warnings on my computer!

jabranham:
I think the remaining byte compiler warnings are trickier to deal with and should probably be PRs on their own. So this is ready to be merged whenever you want. It's a lot of commits, so let me know if you want me to squash them.

Down to about 20 compiler warnings on my end.

vspinu:
This is great indeed!

57 commits is a bit too much noise, especially that many apply the same fix in different places. Not a big deal though, it reflects the amount of work you put into a task which no one else dared to attempt ;) I think we can merge as it is.

lisp/ess-comp.el
vspinu: Agreed. We should remove dependency on ess-site and streamline inner dependencies. The everything depends on everything is not the right approach.

jabranham:
Yeah, I kept thinking I had found fixed all warnings of type X, made a commit, did some more stuff, then found more warnings about type X. 

I'm happy that all this work did actually fix a few bugs :-)

lisp/ess-dde.el
```diff
@@ -96,7 +96,7 @@ PROC, VISIBLY and MESSAGE are ignored."
 If prefix arg is given, forces a query of the ESS process for the help
```
vspinu: I think no-one is using dde these days. I really doubt it works as originally intended. 

Maybe it's time to move this part to `./old/`?

lisp/ess-inf.el
vspinu: Would be good to investigate if that action could be done on remotes without using internal tramp functions.

lionel-:
@jabranham If possible could you squash together the commits that apply the same fix? This would make for a cleaner history. If this leads to too many rebase conflicts don't bother though!

While you are fixing stuff you can just amend existing commits and then force push to the PR branch btw.

jabranham:
@lionel- Sure, did so here: https://github.com/jabranham/ESS/tree/fix-byte-compiler-warnings. It still consists of a large number of commits (25), but I think they all do different things. Let me know if you want me to squash anything else together.

Would you rather I force-push that branch here or open a new PR?

lionel-:
You can force-push here, thanks. I also regularly rebase on master to keep history linear.

jabranham:
OK, I'll rebase this off master and force push here in a few minutes.

jabranham:
OK, done.

lisp/ess-dde.el
jabranham: That would make sense to me. I think another candidate is the mouseme stuff. Or do people use that?

jabranham:
Just FYI - I don't think the travis build failure is my fault. It looks broken on master after the recent commits. Since I rebased off master, it's now broken here too.

PR #523, opened 2018-04-16
Update documentation

### LINKED ISSUE #496: Documentation outdated in some parts
`ESS/doc/ess.texi` sections and related examples are primarily focused on S-PLUS and R is only seen as an alternative/special case. This was true in the past, now it should be right the opposite.  I was unable to find clear references to this product on TIBCO website, but according [wikipedia.org/wiki/S-PLUS](https://en.wikipedia.org/wiki/S-PLUS) the latest stable release is 7 years old and, "due to the increasing popularity of the open source S successor R", TIBCO Software itself has released a runtime for R. 

Similarly, the documentation speaks of "MSDOS prompt shell". The latest MSDOS was released in 2000. 
Even if keeping the old jargon for a while made sense, after 18 years from the last release, it doesn't help. Also, since 2009, PowerShell shell is an integral part of Windows, therefore examples should target this shell. Addressing the MSDOS shell might create confusion in new and younger users of Windows and ESS (who did not meet to the MSDOS era).

There are other similar instances (e.g. Cygwin, "new" features relatively old). I wonder whether the old sections are kept by design or there is simply a lack of time to update the documentation to devote more time to developing the project. In the latter case, one could volunteer to update the `ess.texi`.

jabranham:
This is a large update to ESS's documentation. Inspired by #496, I went through and made the documentation focus on R instead of S4/5/6/+. I also:

* updated the wording
* added sections on company.el and Xref 
* removed the ETAGS section
* fixed typos where I found them (probably introducing some new ones too)
* removed very old new features from the "new features" section (put them in onewfeats.texi)
* removed a lot of duplicated information from the help-s.texi file
* removed the extended examples of working with R/S where the objects are real and source code is a by-product. It's mentioned briefly that it's possible to work that way but you probably shouldn't be doing so.
* moved a lot of files to an archive/ folder just to clean up the doc/ folder a bit.
* added sentance-end-double-space t to the dir-locals file
* made a function obsolete via `make-obsolete` rather than in the docstring (make-obsolete will put it in the docstring too)

There might be one or two other things I'm not remembering at the moment. Let me know if something doesn't look good.

Closes #496.


miguelmorin:
Hi @jabranham , I went through the ESS documentation and the two Google Code Archive pages. I proposed to merge the two Google Code Archive pages into ESS documentation in #532, then @vspinu mentioned this issue, so I flag it here. I'll wait for this update, then merge the documentation if needed.

vspinu:
This looks good. I am not a big fan of double spacing for sentence ending. I think a dot and a space should be enough. I also think most of the docs in ESS are disregarding this old convention and there is no much point in changing that.

vspinu:
Setting `sentence-end-double-space` to nil should disable both checks.

PR #524, opened 2018-04-20
Use display-buffer-alist instead of obsolete variables

jabranham:
`display-buffer-alist' obsoleted many variables like
`special-display-regexps' in Emacs 24.3, and the byte compiler
complains about this.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -185,11 +185,7 @@ Alternatively, it can appear in its own frame if
 
       (let* ((infargs (or ess-start-args
                           inferior-ess-start-args))
-             (special-display-regexps nil)
-             (special-display-frame-alist inferior-ess-frame-alist)
              (proc (get-process procname)))
-        (if inferior-ess-own-frame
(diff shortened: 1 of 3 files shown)


vspinu:
I am not a frame user but AFAIC judge this one is clean and safe to merge. So I will merge and try it locally.

lisp/ess-inf.el
```diff
@@ -3185,8 +3188,11 @@ search path related variables."
 (defun ess-display-temp-buffer (buff)
   "Display the buffer BUFF using `temp-buffer-show-function' and respecting
 `ess-display-buffer-reuse-frames'."
-  (let ((display-buffer-reuse-frames ess-display-buffer-reuse-frames))
-    (funcall (or temp-buffer-show-function 'display-buffer) buff)))
+  (if temp-buffer-show-function
```
vspinu: Why was this change needed? Unless I miss something, the code within fboundp is not the correct application of `temp-buffer-show-function`.

PR #525, opened 2018-04-20
ess-mode: Set prog-mode as the parent keymap & run prog-mode hooks

### LINKED ISSUE #219: Does not inherit from prog-mode
Is there a particular reason ESS does not inherit from prog-mode? See: https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Major-Modes.html.  All programming modes are "supposed" to inherit from this, so that they can be configured together with prog-mode-hook.

jabranham:
I think this closes #219 as ess-mode will now have the prog-mode keymap and run prog-mode hooks. Are we missing anything else?


lionel-:
Thanks! Could you add a couple tests please?

lionel-:
oh and a news item

jabranham:
Sure, I just added some tests. There's already a news item, do I need to add something to it?

test/ess-r-tests.el
```diff
@@ -4,11 +4,17 @@
 (require 'ess-r-tests-utils)
 
 
-;;; Inferior R
+;;; R
 
 (ert-deftest ess-r-inherits-prog-mode ()
-  (with-r-file nil
-    (should (derived-mode-p 'prog-mode))))
+  (when (getenv "TRAVIS")
```
lionel-: hmm I don't think we should be testing only on Travis. How about something like

```lisp
(let ((map (current-local-map))
      found)
  (while (and map (not found))
    (if (eq (keymap-parent map) prog-mode-map)
        (setq found t)
      (setq map (keymap-parent map))))
  found)
```

This kind of test might be useful when/if we refactor ESS modes.

PR #526, opened 2018-04-22
Make writing to the dribble buffer optional

jabranham:
I'm betting most users never look at this buffer. Since its stated use is for debugging, it should probably not be on by default. You can keep the old behavior by setting the new variable to non-nil (and if you've previously set `ess-verbose`, the behavior won't change).

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -35,6 +35,9 @@ So, for example, R buffers will now show ESS[R] rather than ESS[S].
 @item @ESS{[R]}: @code{Makevars} files are now automatically opened
 with @code{makefile-mode}.
 
+@item New varaible @code{ess-write-to-dribble}.
+This allows users to disable the dribble (@code{*ESS*}) buffer if they wish.
+
 @end itemize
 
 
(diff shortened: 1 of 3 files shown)


mmaechler:
We have had it on by default very much on purpose:
Try  M-x ess-submit-bug-report  (which I think also has a menu entry): It crucially uses the contents of *ESS*.
What's the big drawback of having it on by default ??
I'd pretty strongly like to keep the current behavior.

jabranham:
Ah, I forgot about that. Would this PR be OK if I changed it to keep the current behavior by default?

vspinu:
> Would this PR be OK if I changed it to keep the current behavior by default?

I think this doesn't harm. Especially during compilation we could use it to disable messages, right?

> Try M-x ess-submit-bug-report (which I think also has a menu entry):

I personally never used dribble buffer it to track bugs, nor I think it's that useful for bug reports. Even for those rare occasions when people do submit bugs with `ess-submit-bug-report` that information is useless because it doesn't track a lot of stuff. The benefit of having clean code out-weights the benefit of having ess-submit-bug-report. So, if after me, I would even remove all that dribble buffer code.

jabranham:
OK, pushed a different commit that keeps the old behavior while allowing users to disable it.

mmaechler:
Thank you @jabranham. This now looks good to me --> hence closing and merging

Issue #528, opened 2018-04-23
Using compilation-mode buffers for devtools commands


atheriel:
At the moment when you use the various `ess-r-devtools-*` commands, these will run in the linked R session. It makes a lot of sense to me for processes like running tests, building documentation, and running package checks to happen in isolated environments/sessions, and Emacs already has nice tooling for this using `compilation-mode`. As a bonus, we'd get things like jumping to errors in files automatically, since this is provided by `compilation-mode`.

Is there any interest in having some sort of flag -- say `ess-r-devtools-use-compile` -- that would run these commands in a compile buffer instead? I'd be happy to do some of the implementation work, if so. Where it makes sense, we could also ensure that packages are reloaded in the linked session.

lionel-:
I think that would make sense. Certainly worth some experimenting!

lionel-:
Also might be nice to have a test command that sources files in `tests`. At the moment we only have an interface around testthat which is too specific for ESS.

vspinu:
I actually don't see much added value in this. It would complicate the code for not much benefit.

>  It makes a lot of sense to me for processes like running tests, building documentation, and running package checks to happen in isolated environments/sessions,

I think the insulation that devtools provide is already pretty good. If we add compilation style package methods it would require adding ways to control which R is selected. I might want to run those in R, R-dev, R old rel etc. With connected R processes this is easy (and will be easier in the future); just re-associate with the inferior in which you want to run the test.

> As a bonus, we'd get things like jumping to errors in files automatically, since this is provided by compilation-mode.

Compilation mode is activated in iESS. Error tracking should work better in iESS than in generic compilation mode because it targets R specific errors. Try M-g n/p when you are in a script buffer and get errors.

lionel-:
> I think the insulation that devtools provide is already pretty good.

I'd say there is currently no insulation? It makes sense to run check / tests in a separate process just like RStudio does.

lionel-:
oops of cours R CMD check is run in a separate process, I was thinking about tests.

lionel-:
I think you're right to be prudent. Also if compilation occurs in a different process this poses the question of reloading restoring the interactive session, which might not be straightforward.

atheriel:
[...]

PR #531, opened 2018-04-25
Allows some devtools commands to be run in a compilation-mode buffer.

atheriel:
This is a proof-of-concept implementation of the feature discussed in #528. I've only implemented `check()`, `test()`, and `document()` for now, and obviously the documentation could use some work as well.

How does this general approach look?

(Known issue: compilation mode buffers do not render the shell-escaped characters produced by `devtools::test()` correctly.)


lisp/ess-r-package.el
```diff
@@ -289,9 +312,13 @@ With prefix ARG ask for extra args."
   "Interface for `devtools::check()'.
 With prefix ARG ask for extra args."
   (interactive "P")
-  (ess-r-package-eval-linewise
-   "devtools::check(%s)\n" "Checking %s" arg
-   '("" (read-string "Arguments: " "vignettes = FALSE"))))
+  (let ((actions '("" (read-string "Arguments: " "vignettes = FALSE"))))
```
vspinu: Too much copy pasting. I would add a wrapper which would dispatch to `ess-r-package-eval-linewise`.

lisp/ess-r-package.el
```diff
@@ -268,6 +273,24 @@ arguments, or expressions which return R arguments."
...
+    (compilation-start command nil
+                       (lambda (name-of-mode)
+                         (concat "*" (downcase name-of-mode) ":" pkg-name "*"))
+                       ess-r-error-regexp-alist)))
+
+(defun ess-r-package-eval-in-compile-buffer (expr)
+  "Evalate R expression EXPR in a standalone `compilation-mode'
+buffer."
+  (let* ((procname inferior-ess-r-program-name))
+    (ess-r-package-compilation-start
```
vspinu: I think this extra helper is not warranted. Could be inlined here.

lisp/ess-r-package.el
```diff
@@ -268,6 +273,24 @@ arguments, or expressions which return R arguments."
         args
       (concat ", " args))))
 
+(defun ess-r-package-compilation-start (command)
+  "Run `compilation-start' in the package's root directory."
+  (let* ((pkg-info (or (ess-r-package-project)
+                       (ess-r-package-set-package)))
+         (pkg-name (ess-r-package-name))
+         (default-directory (cdr pkg-info)))
+    (compilation-start command nil
```
vspinu: `t` instead of nil would result in compilation-shell-minor-mode which should solve the term signals issue.

lisp/ess-r-package.el
vspinu: Indeed. Let's strive for minimal code for now.

atheriel: Fixed.

lisp/ess-r-package.el
atheriel: Unfortunately this also puts it in `comint-mode`, which is quite a different experience than the usual `compilation-mode`. I don't see an obvious way to get term signal handling outside of `comint-mode` -- do you have any inkling of what could be done?

atheriel:
Hey, sorry for the long silence on this. I've made a minor update to the patch, but I'm not sure how to proceed.

lisp/ess-r-package.el
atheriel: I had a hard time coming up with a design for such a wrapper -- do you have an idea about how to do it nicely?

lisp/ess-r-package.el
vspinu: Why does that matter? If compilation mode is based on comint this is how it is.

lionel-:
FYI I have extracted `ess-r-eval-in-compile-buffer` from `ess-r-package-eval-in-compile-buffer`:

```lisp
(defun ess-r-eval-in-compile-buffer (expr)
  "Evalate R expression EXPR in a standalone `compilation-mode' buffer."
  (let* ((procname inferior-ess-r-program-name)
         (command (format "%s --slave --no-readline -e \"%s\"" procname expr)))
    (compilation-start command nil
                       (lambda (name-of-mode)
                         (concat "*" (downcase name-of-mode) "*"))
                       ess-r-error-regexp-alist)))
```

I use it for rendering markdown documents:

```lisp
(defun my-ess-rmarkdown-render ()
  (interactive)
  (let ((cmd (format "rmarkdown::render('%s')" buffer-file-name)))
    (ess-r-eval-in-compile-buffer cmd)))
```

I quite like that this uses a separate process at each time rendering. I haven't tried it yet for devtools commands but I think we should explore compilation-mode further after the next release.

vspinu:
@atheriel  sorry for the late reply. We will merge this after the release in September. A bit more thinking is needed here.

atheriel:
Yes, I appreciate that. I'd like to get back to it. One thing I hadn't considered (but is pertinent to my own use cases) is how to ensure that this will interoperate with `ess-remote` correctly.

jabranham:
What's the status of this? Are we implementing it for 19.04?

lionel-:
I don't think so. But I'd still like to integrate compilation buffers in ESS at some point.

atheriel:
I'm still a big fan of this approach, although there are a few outstanding issues mentioned in the thread:

(1) Rendering of escape characters.
(2) Whether to use `compilation-mode` or `comint-mode`; and
(3) How this will work with `ess-remote`.

As well as any actual design questions. Any thoughts on these?

vspinu:
I had some very concrete thoughts back then but didn't have time to get to it. And now I am even more busy to have a look at this. 

One thing for sure, the functionality should be brought in once we have the local cache for prog + params which is also needed for restarts.

kguidonimartins:
This would be a nice feature! Any update?

Issue #406, opened 2017-02-20
nil in load-history


tagteam:
with emacs 26.0.50.1 (or emacs 25.x) after loading ess (16.10-1 [elpa: 20170206.310]) via

```lisp
(use-package ess-site :ensure ess)
```

I get a value 'nil' in the load-history

```lisp
(mapcar 'car load-history)

"~/emacs-genome/genes/ess-20170206.310/lisp/ess-site.elc"
nil
"~/emacs-genome/genes/ess-20170206.310/lisp/ess-toolbar.elc"
"~/emacs-genome/genes/ess-20170206.310/lisp/ess-mouse.elc"
```

the 'nil' occasionally causes trouble, e.g., when `help--loaded-p` does this:

```lisp
(dolist (x load-history) (if (string-match-p re (car x))
(setq done t)))
```

mmaechler:
Note that  `(use-package .)` is not in my emacs 25.x, but the same phenomenon appears when using
`(require 'ess-site "<path-to-ess>/lisp/ess-site.el")`    after starting "vanilla emacs"  (=  `emacs -Q`).

jabranham:
Do you still experience this? I can't seem to reproduce it using today's ESS.

tagteam:
yes, I still have it to after upgrade to 2018-04-26:

(mapcar 'car load-history)
...
"/home/tag/emacs-genome/genes/ess-20180426.1302/lisp/ess-site.elc" nil "/home/tag/emacs-genome/genes/ess-20180426.1302/lisp/ess-toolbar.elc"

but, yesterday was the 2nd of May. where can I get this version? thanks!

vspinu:
Yes, please. It's a very good idea. That buffer pasting and evaluation is just an R-programmer doing lisp :)

vspinu:
How about `ess-r-create-runner` instead of `ess-r-create-version`?

jabranham:
Yeah, the name is gross. `ess-r-create-runner` is better. 

It's not immediately clear to me how `inferior-R-version` actually tells the `R` function which binary to call. Do you happen to know that off the top of your head?

On a somewhat-related note, can we rename `R` to something like `ess-run-R` and alias `R` and `run-R` to it? Sometimes when I'm testing out functions I do something like `defun R` as a placeholder (while not thinking about R/ESS), which is annoying since I then have to go find the ESS R function again. 

[...]

PR #537, opened 2018-05-03
Rename R to ess-run-R

jabranham:
- ess-run-R is a cleaner namespace
- we're aliasing run-R to ess-run-R to match e.g. run-python,
  run-octave
- keep R as an alias for ess-run-R

I mentioned this in #406 but I occasionally am fiddling with Emacs and do something like `(defun R <test-code>`. This makes a namespace friendlier function and a few aliases.


lionel-:
I dislike capital letters in elisp symbols. `R` makes sense but I'd prefer `run-r` just like we have `run-python` and not `run-Python`.

I also don't like having both `run-r` and `ess-run-r`. I think it should be a single `run-ess-r` for consistency with our namespace. Though I've thought for a long time we should just grab the `r-` namespace.

vspinu:
I agree with @lionel-. Even without run-ess-r aliases we have way too many runners. 

Shall we stick with 2 versions `run-r-xyz` and `R-xyz`?

lionel-:
> Shall we stick with 2 versions run-r-xyz and R-xyz?

Sholud this be `run-ess-r-xyz` for consistency? Or do you plan to slowly switch to the `r-` namespace? If we go for the latter we'll complicate the situation.

vspinu:
> Sholud this be run-ess-r-xyz for consistency? 

Sure. I don't think people will be using it much anyways.


> Or do you plan to slowly switch to the r- namespace? 

I don't think it matters much and changing everything into r- namespace would also mean changing sas- and stata-. Too much hustle for no much benefit. 

I still haven't abandoned the idea of a clean rewrite of r-mode not hindered by backward compatibility. But, given the renewed development on ESS (from @jabranham and you) the prospects of ESS modernization look pretty good :).

lionel-:
> I don't think it matters much and changing everything into r- namespace would also mean changing sas- and stata-. Too much hustle for not much benefit.

I see `STA-`, `stata-`, `ess-stata-`, `sas-`, `SAS-` and `ess-sas-` namespaces /o\

> I still haven't abandoned the idea of a clean rewrite of r-mode not hindered by backward compatibility. But, given the renewed development on ESS (from @jabranham and you) the prospects of ESS modernization look pretty good :).

Right we should wait to see long term plans before switching to `r-` (if we ever do). I think modernising ESS is a better option though. The big question is whether it's worth it to be burdened by language genericity.

In any case, let's stick to `ess-r-` for now.

lionel-:
> But, given the renewed development on ESS (from @jabranham and you) the prospects of ESS modernization look pretty good :).

Let's not forget the new genericity mechanism that you contributed ;)

vspinu:
It was a quick attempt to streamline generic dispatch, but given that most emacs tools (navigation particularly) don't understand it, it's hard to work with. Emacs 25 has cl-generic which dispatches on context, once we drop 24 dependency we should probably move to that system.

mmaechler:
DEFINITELY do keep  R-<x.y.z> etc --- this is *THE* most used M-x functionality in ESS, and ingrained in thousands of brains.    Sorry but that namespace cleanness really is not the only criterion.   Backwards compatibility (and ease of use / remembering/.. notably for beginners) *is* an important criterion too.

lionel-:
Martin, I think everyone has been on the same page in that thread. `R` aliases should remain despite the shift to clean namespaces.

mmaechler:
@liionel- thank you for the clarification -- and excuse my not carefully reading everything.

jabranham:
So if I understand correctly, yall want the function to be named `run-ess-r` and alias `R` to that, right?

lionel-:
Yes please.

jabranham:
OK, done

PR #538, opened 2018-05-05
Improvements to fontification

lionel-:
* Distinguish between function-like keywords that take parameters such as `if` or `for` and simple keywords like `break` or `next`. Only fontify the former if they precede an opening parenthesis. This should fix many spurious fontifications. The same fix is also applied to search path modifiers such as `require()` and `library()`.

* Fontify comma-separated `@param` keywords in roxygen blocks properly.

* Fontify `tryCatch()`, `invokeRestart()`, `withRestarts()`, `withCallingHandlers()` and `signalCondition()` as keywords since these functions (at least potentially) affect control flow.

* Add literate unit tests for fontification. Fix an issue in the roxy font-lock regexps in the process.

* ~~Start moving font-lock regexp variables to the private namespace `ess-r--`. It should be clear to our users that modifying these regexps might require some maintenance when updating ESS in case we change things.~~

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -7,9 +7,10 @@ Changes and New Features in development version:
 @item @ESS{[R]}: Fontification of roxygen @code{@@param} keywords now
 supports comma-separated parameters.
 
-@item @ESS{[R]}: Function-like keywords such as @code{if} or @code{stop}
-are no longer fontified as keyword if not followed by an opening
-parenthesis.
+@item @ESS{[R]}: Function-like keywords such as @code{if ()} or
+@code{stop()} are no longer fontified as keyword if not followed by an
+opening parenthesis. The same holds for search path modifiers like
+@code{library()} or @code{require()}.
 
 @item ESS modes now inherit from @code{prog-mode}.
 
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -10,7 +10,8 @@ supports comma-separated parameters.
 @item @ESS{[R]}: Function-like keywords such as @code{if ()} or
 @code{stop()} are no longer fontified as keyword if not followed by an
 opening parenthesis. The same holds for search path modifiers like
-@code{library()} or @code{require()}.
+@code{library()} or @code{require()}. This feature is only available in
+Emacs >= 25.
 
 @item ESS modes now inherit from @code{prog-mode}.
 
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,6 +4,10 @@
 Changes and New Features in development version:
 @itemize @bullet
 
+@item @ESS{[R]}: Function-like keywords such as @code{if} or @code{stop}
+are no longer fontified as keyword if not followed by an opening
+parenthesis.
+
 @item ESS modes now inherit from @code{prog-mode}.
 
 @item @ESS{[R]}: The package development minor mode now only activates
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,6 +4,9 @@
 Changes and New Features in development version:
 @itemize @bullet
 
+@item @ESS{[R]}: Fontification of roxygen @code{@@param} keywords now
+supports comma-separated parameters.
+
 @item @ESS{[R]}: Function-like keywords such as @code{if} or @code{stop}
 are no longer fontified as keyword if not followed by an opening
 parenthesis.
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2633,16 +2633,16 @@ This variable has no effect. Customize
   (append ess-RS-constants
           '("T" "F")))
 
-(defvar ess-r--keywords
-  '("in" "else" "break" "next"))
+(defvar ess-R-bare-keywords
+  '("in" "else" "break" "next")
+  "Keywords that do not precede an opening parenthesis.")
 
-(defvar ess-r--fn-like-keywords
+(defvar ess-R-keywords
   '("while" "for" "if" "switch" "function" "return" "message" "warning" "stop")
   "Keywords that precedece an opening parenthesis.")
-(defvaralias 'ess-R-keywords 'ess-r--fn-like-keywords)
 
 (defvar ess-S-keywords
-  (append ess-r--keywords '("terminate")))
+  (append ess-R-keywords '("terminate")))
 
 ;; only some of these keywords "look like functions but are not":
 (defvar ess-S-non-functions
@@ -2771,13 +2771,15 @@ default or not."
         '(1 font-lock-function-name-face nil))
   "Font-lock keyword - function defintions for R.")
 
-(defconst ess-r--fl-keyword:keywords
-  (cons (regexp-opt ess-r--keywords 'words)
-        'ess-keyword-face))
+(defconst ess-R-fl-keyword:bare-keywords
+  (cons (regexp-opt ess-R-bare-keywords 'words)
+        'ess-keyword-face)
+  "Font-lock keywords that do not precede an opening parenthesis.")
 
-(defconst ess-r--fl-keyword:fn-like-keywords
-  (cons (concat (regexp-opt ess-r--fn-like-keywords 'words) "\\s-*(")
-        'ess-keyword-face))
+(defconst ess-R-fl-keyword:keywords
+  (cons (concat (regexp-opt ess-R-keywords 'words) "\\s-*(")
+        'ess-keyword-face)
+  "Font-lock keywords that precede an opening parenthesis.")
 
 (defvar ess-R-fl-keyword:assign-ops
   (cons (regexp-opt ess-R-assign-ops) 'ess-assignment-face)
@@ -2798,8 +2800,8 @@ default or not."
 (defcustom ess-R-font-lock-keywords
   '((ess-R-fl-keyword:modifiers  . t)
     (ess-R-fl-keyword:fun-defs   . t)
-    (ess-r--fl-keyword:keywords . t)
-    (ess-r--fl-keyword:fn-like-keywords . t)
+    (ess-R-fl-keyword:bare-keywords . t)
+    (ess-R-fl-keyword:keywords . t)
     (ess-R-fl-keyword:assign-ops . t)
     (ess-R-fl-keyword:constants  . t)
     (ess-fl-keyword:fun-calls)
@@ -2854,8 +2856,8 @@ system described in `inferior-ess-font-lock-keywords'.")
     (ess-R-fl-keyword:messages  . t)
     (ess-R-fl-keyword:modifiers . t)
     (ess-R-fl-keyword:fun-defs  . t)
-    (ess-r--fl-keyword:keywords . t)
-    (ess-r--fl-keyword:fn-like-keywords . t)
+    (ess-R-fl-keyword:bare-keywords . t)
+    (ess-R-fl-keyword:keywords . t)
     (ess-R-fl-keyword:assign-ops	. t)
     (ess-R-fl-keyword:constants . t)
     (ess-fl-keyword:matrix-labels	. t)
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2635,11 +2635,11 @@ This variable has no effect. Customize
 
 (defvar ess-r--keywords
   '("in" "else" "break" "next"))
-(defvaralias 'ess-R-keywords 'ess-r--keywords)
 
 (defvar ess-r--fn-like-keywords
   '("while" "for" "if" "switch" "function" "return" "message" "warning" "stop")
   "Keywords that precedece an opening parenthesis.")
+(defvaralias 'ess-R-keywords 'ess-r--fn-like-keywords)
 
 (defvar ess-S-keywords
   (append ess-r--keywords '("terminate")))
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2762,7 +2762,7 @@ default or not."
 
 ;;; fl-keywords R
 (defvar ess-R-fl-keyword:modifiers
-  (cons (regexp-opt ess-R-modifyiers 'words)
+  (cons (concat (regexp-opt ess-R-modifyiers 'words) "\\s-*(")
         'ess-modifiers-face)     ; modify search list or source (i.e. directives)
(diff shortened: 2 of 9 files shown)


lisp/ess-custom.el
```diff
@@ -2587,14 +2587,16 @@ This variable has no effect. Customize
   (append ess-RS-constants
           '("T" "F")))
 
-(defvar ess-R-keywords
-  ;; "Reserved Words" -- part 2 --
-  '("while" "for" "in" "repeat" "if" "else" "switch" "break" "next" "function"
-    ;; note that these are *NOT* reserved words in R:
-    "return" "message" "warning" "stop"))
+(defconst ess-r--keywords
+  '("in" "else" "break" "next"))
```
vspinu: I don't think we should changer/remove this. It's a very old variable and people are probably customizing it. I would also not down-case it for now. Either downcase all such variables (and alias) or none. 

Why don't you just create a new defconst for what you need?

doc/newfeat.texi
```diff
@@ -4,6 +4,13 @@
 Changes and New Features in development version:
 @itemize @bullet
 
+@item @ESS{[R]}: Fontification of roxygen @code{@@param} keywords now
+supports comma-separated parameters.
+
+@item @ESS{[R]}: Function-like keywords such as @code{if} or @code{stop}
+are no longer fontified as keyword if not followed by an opening
```
vspinu: Why just `if`, `stop` and not all keywords?

lisp/ess-custom.el
lionel-: > Why don't you just create a new defconst for what you need?

We have to split this variable in two parts to fix its fontification. Maybe we can add an alias so the old symbol points to the new function-like keywords?

I _think_ it's better to gradually move towards a consistent naming scheme when we make modifications.

lionel-:
Also I would remove `switch`, `message` and `warning` from there because these are regular functions that don't affect control flow.

vspinu:
How control flow is related to highlighting? (switch is a control flow function).

doc/newfeat.texi
vspinu: If your taks is to fix spurious fontification,  then I see no reason at all to distinguish between "function like" keywords and other keywords. 

What is "function-like" keywords in the first place?

lisp/ess-custom.el
lionel-: > Maybe we can add an alias so the old symbol points to the new function-like keywords?

Done. Should we obsolete the old variable?

> I think it's better to gradually move towards a consistent naming scheme when we make modifications.

Otherwise convergence will likely never happen.

vspinu: >  it's better to gradually move towards 

That creates confusion and annoyance. At least it should be done within blocks of related functionality (fontlock, indentation, interaction etc).

vspinu: This R-r dichotomy is generally not bothering me as much as it does you. I think as long as within related groups of functions notation is consistent, it's not that bad.

lionel-:
> How control flow is related to highlighting? (switch is a control flow function).

I guess that's true. In a way all R functions are control-flow functions because of lazy evaluation (they all decide which of their arguments are evaluated). Control flow seems to be a red herring but I think we should only highlight:

- Keywords with special syntax like `for` or `if`
- `return()` as it causes an unconditional jump if evaluated. Following that logic we should add `invokeRestart()` to keywords. Maybe `tryCatch()` as well.

doc/newfeat.texi
lionel-: A keyword that precedes an opening parenthesis. I've been greatly annoyed by spurious fontifications of `if`, `message` etc. Now they are only fontified if an opening parenthesis follows.

> I see no reason at all to distinguish between "function like" keywords and other keywords.

What do you mean? The goal is to keep fontifying `next`, `break` and other non-function-like keywords that could cause a jump.

vspinu: With `next`, `break` no chance, but with `while`, `for`, `while`, `switch` etc. it's possible.

lionel-: What do you mean?

lisp/ess-custom.el
```diff
@@ -2587,11 +2587,11 @@ This variable has no effect. Customize
   (append ess-RS-constants
           '("T" "F")))
 
-(defconst ess-r--keywords
+(defvar ess-r--keywords
   '("in" "else" "break" "next"))
 (defvaralias 'ess-R-keywords 'ess-r--keywords)
 
-(defconst ess-r--fn-like-keywords
+(defvar ess-r--fn-like-keywords
```
vspinu: `defconst` is really not a const in emacs:)

`--` is really not warranted here. It's a kinda user level variable even if it's not a custom.

doc/newfeat.texi
vspinu: What is not clear more concretely? while, switch, stop etc. are function like, just like `if` and `for`. So all I am saying is that your fix should apply to those as well, not just to `if` and `for`.

lisp/ess-custom.el
lionel-: I think these variables should be modifiable (I have now switched it back to defvar for this reason) but that they should not be part of a configuration UI. I.e. they should not constrain possible updates to how we fontify things.

I have reverted the naming back to the old scheme though so the `--` is gone.

lionel-:
> What about else, break, next etc?

They keep being highlighted. They fall in the "jumping" group. I am only suggesting to remove `switch`, `message` and `warning` for consistency. Without great hopes of convincing anyone else though ;)

> To my mind, if you just leave the ess-R-keywrods alone and fix spurious highlighting (parenthesis fix) and roxygen @param it would be just right.

That's basically the state of this PR now that the naming changes have been reverted in the last commit.

doc/newfeat.texi
lionel-: > So all I am saying is that your fix should apply to those as well, not just to if and for.

That is the case, it's now even covered by unit tests.

vspinu:
Cool. Thanks for being flexible! I am fine with all the changes.

lionel-:
I have added `tryCatch()`, `invokeRestart()` and `signalCondition()` to the keyword list. The latter to for consistency with `message()` and `warning()`.

doc/newfeat.texi
vspinu: > That is the case

I see. Should have been more careful with reading your PR :/, sorry.

lionel-:
@vspinu @mmaechler I am experimenting with adding a new group of keywords for fontification. The goal is to separate the keywords that form the backbone of a function from those that merely signal possible control flow effects.

With this scheme:

* `return()`, `stop()` and `invokeRestart()` are kept in the original keywords list because they cause an unconditional jump.

* A new keyword list for the condition system holds all the following functions: `message()`, `warning()`, `signalCondition()` and `tryCatch()`, `withRestarts()`, `withCallingHandlers()`. These keywords are fontified the same way as search path modifiers (which should be a weaker colour than the keywords lists in properly configured ESS).

I find this substantially improves readability.

warning-message
withcallinghandlers
trycatch


What do you think?

vspinu:
An overengineering I would say. Having a special face for 3 keywords is an overkill IMO. Have you tried activating `ess-R-fl-keyword:fun-calls`? This is what I use and having 3 kind of colors for function calls don't quite appeal to me. 

`stop`, `message`, `warning` form a conceptual group in my mind (signals), and the proposed segmentation is way too technical to be worth the distinction. Other conceptual issues: `signalCondition` can be an interrupt as well, `warning` can exit if `options(warn)` is set etc.

lionel-:
> Have you tried activating ess-R-fl-keyword:fun-calls?

Yes this is what I use in my config as well. Even so, I have been using the new colour group all weekend and it's been a real pleasure. With this scheme the fontification is informative of the evaluation semantics of the function. At the very least I would like to introduce the new group and have it inherit from the keywords face. However I would strongly suggest to make it inherit from the modifiers face by default. Having them as a separate colour is a feature.

> Other conceptual issues: signalCondition can be an interrupt as well, warning can exit if options(warn) is set etc.

This is not a conceptual issue: the new group is about functions potentially affecting control flow. It is consistent (unlike the current keywords group) and makes sense.

vspinu:
@mmaechler 

> started to emphasize more and more that all R syntax is about function calls and that most "keywords" in traditional languages really are R functions.

From this I can infer that you are actually with me :) My view is that we shouldn't split keywords even further and create salad for the sake of fuzzy definition of non-local exit. To be clear, Lionel proposes to color function calls with 4 faces, one for `stop`, one for `signalCondition`, `tryCatch` ..., one for `for`, `if` ..., and the final one for all other function calls. 

If you believe that `stop` is a function and is different from `next` and `break` then you should't be agreeing.

@lionel- 

> Sure, but the keywords group is not about signals it is about hard control flow.

Who said that?

>  the new group is about functions potentially affecting control flow. 

 Every function call _potentially_ affects program flow because it can throw.

 `for`, `if` are control flow functions, and with your proposal they will be highlighted (with default faces) as `stop` `return` `break` `next` etc. Right? I would say, if we go with your proposal we should create new group for `stop`, `return`, `break`, `next`, not the other way around.

If I understand correctly, your proposal boils down to the following:

  * signals `stop()`, `warning()`, `invokeRestart()`, `simpleWarning` will be in 3 different colors.
  * functions (as per Martins's classification) will be in 4 different colors.
  * control flow `for`, `if`, `tryCatch`, `invokeRestart`, `break`, `next` will be in 3 different colors.

I do see the point of your non-local exit proposal but it's just one possible taxonomy, and, probably, not the most common. 

For example, here is the taxonomy provided by R's own documentation:

  * `Control`: if, for, while, repeat, break, next
  * `conditions`: tryCatch, withCallHandlers, signalCondition, stop, warning, message etc. 
  * `function`: function and return

Another taxonomy is one used by Rstudio (if we want to mimic them),

> However I would suggest to make it inherit from the modifiers face by default. Having them as a separate colour is a feature.

Which group? Can you please post the exhaustive list of keywords (and faces) which will be highlighted after your proposal?

lionel-:
(On a side note, I think we could get rid of this new `ess-R-bare-keywords` variable. We should just check for `"in" "else" "break" "next"` in `ess-R-keywords` and generate the appropriate regexp variables accordingly. This way the implementation details of fontification don't leak into the configuration UI.)

vspinu:
> @vspinu This is a long comment based on false premises. 

If you could have taken care to explain it clearly I wouldn't have had to guess.

> To avoid diluting the discussion, could you please review 

Given that it takes 30 sec for you to reply, I guess I shouldn't be even trying to analyze the problem from different angles. Indeed, please review. 

> hat I do not suggest 4 groups but just 2:

I didn't say that you propose 4 groups. I said that after your implementation functions will be highlighted with 4 faces. It wasn't clear from your proposal if break/next are different from for/if and if they are different from stop/return etc. So I assumed those are different. 

In any event, there will be 3 colors for functions. Does that make a difference?

> Hard control-flow keywords: switch, if, for, break, next, invokeRestart, return, stop. Soft control-flow 
 eywords: tryCatch, withCallingHandlers, withRestarts, signalCondition, message, warning

Thanks for clarifying the proposal. It's a long way from what you wrote in the beginning. 

> he third group (ordinary function calls) is not fontified by default and should not affect this discussion.

It does because there will be functions not in your list but which do behave similarly (`simpleWarning`, `simpleError`, `conditionCall` + all custom control flows like `do`, `dopar`, `when`, `times` etc). What about `while`?

How do you explain to people that `message` is a control flow? Why not `print`, `cat`?

lionel-:
> If you could have taken care to explain it clearly I wouldn't have had to guess.

No need for snappy comments, please. I thought I had explained my proposal clearly.


> It's a long way from what you wrote in the beginning.

This is what I originally wrote:

> * return(), stop() and invokeRestart() are kept in the original keywords list because they cause an unconditional jump.
>
> * A new keyword list for the condition system holds all the following functions: message(), warning(), signalCondition() and tryCatch(), withRestarts(), withCallingHandlers().

Is this such a long way?

> It does because there will be functions not in your list but which do behave similarly (simpleWarning, simpleError, conditionCall + all custom control flows like do, dopar, when, times etc). What about while?

The condition constructors and accessors are ordinary functions and should never be fontified differently. The custom control flow functions should be set up in user config (as a heavy user of rlang this is what I do).


> How do you explain to people that message is a control flow? Why not print, cat?

`message()` and `warning()` send catchable conditions and their fontification is a reminder that it might cause a jump and code below it might not be called, though typically will. They potentially affects control flow unlike `cat()` and `print()`. There might be other signallers out there (e.g. in rlang) but the default keyword lists are about base functions. Maybe someday we can figure out a way to communicate special keywords to ESS from the onLoad hooks of packages, perhaps the MPI facilities would help with this.

lionel-:
I'm not sure `withCallingHandlers()` belongs in the new group because that's the only function in the list that does not represent a potential jump or jump target.

lionel-:
> I'm not sure withCallingHandlers() belongs in the new group because that's the only function in the list that does not represent a potential jump or jump target.

But the fact that it has an effect on the call stack (installs handlers) is sufficient to have it there.

Where would `on.exit()` belong? It seems clear to me that `on.exit()` needs special fontification as well. I think it'd belong in the original keywords group because its argument is always evaluated on stack unwinding (so conceptually causes a jump).

lionel-:
Just in case this isn't clear, of course all R functions might not return, but the suggested keyword groups are about functions that are deterministically or potentially _original causes_ of jumps. If there were no such functions anywhere in the call tree, control flow would be perfectly linear.

lionel-:
@vspinu I think I should have used softer words and said that your comment was based on a misunderstanding rather than on false premises. Sorry about that.

vspinu:
Yerh ... things escalate rather quickly recently. But well, written word is unnuanced and better not be taken literally. It's enough that we have to type so much; choosing careful words would add a major inefficiency.    

As I said earlier, I do see the benefits of the proposal.  Enhanced readability seems to be more important than keeping all signals in one face, and control-flow based categorization is surely more useful than the current "reserved keywords" semantics. The "salad" thing is a matter of taste and as long as it's customizable we should be fine. In any case, a separate PR should be made for this to put the change into a better prospective. 

If `browser` is added, a less known `recover` should be there as well. `on.exit` should surely be in because it contains a "non-continuous" section of the program.  

A slight deviation from the definition is that `function` and `in` are non control reserved words, but well, I guess we can live with that.

PR #542, opened 2018-05-12
Respect Emacs keybinding conventions

### LINKED ISSUE #539: Respect emacs "C-c <letter>" keybinding conventions
I noticed that ess-noweb-mode binds "C-c h" to "ess-handy-commands". This is bad practice according to the emacs manual:

> Don't define C-c letter as a key in Lisp programs. Sequences consisting of C-c and a letter (either upper or lower case) are reserved for users; they are the only sequences reserved for users, so do not block them.
> 
> Changing all the Emacs major modes to respect this convention was a lot of work; abandoning this convention would make that work go to waste, and inconvenience users. Please comply with it.

Source: https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html

Here are the keybindings I found violating this convention:
```
ess-rd.el:192:    (define-key map "\C-ch" 'ess-handy-commands)
ess-sas-d.el:203:  (define-key sas-mode-local-map "\C-ci" 'ess-eval-line-and-step-invisibly)
ess-help.el:611:    (define-key map "\C-ch"   'ess-handy-commands)
ess-s-lang.el:750:  (define-key ess-mode-map          "\C-cf" 'ess-insert-function-outline)
ess-s-lang.el:751:  (define-key inferior-ess-mode-map "\C-cw" 'ess-execute-screen-options)
ess-noweb-mode.el:275:    (define-key map "\C-ch" 'ess-handy-commands)
```

jabranham:
- Don't bind keys to C-c [letter]
- Remove ess-add-MM-keys (not referenced elsewhere in docs/code)

Closes #539

### DIFF
--- a/lisp/ess-s-lang.el
+++ b/lisp/ess-s-lang.el
@@ -747,7 +747,6 @@ underscore character."
   (interactive)
   (require 'ess-mode); typically unnecessary
   (require 'ess-inf); dito
-  (define-key ess-mode-map          "\C-cf" 'ess-insert-function-outline)
   (define-key inferior-ess-mode-map "\C-cw" 'ess-execute-screen-options)
 
   ;; Make M-- : [Alt] + [-] (in addition to / instead of  "_" = (on US-keyboard) [Shift]+ [-]
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -43,6 +43,15 @@ with @code{makefile-mode}.
 @item New varaible @code{ess-write-to-dribble}.
 This allows users to disable the dribble (@code{*ESS*}) buffer if they wish.
 
+@item ESS now respects Emacs conventions for keybindings.
+This means that The @code{C-c [letter]} bindings have been
+removed. This affects @code{C-c h}, which was bound to
+@code{ess-eval-line-and-step-invisibly} in @code{sas-mode-local-map};
+@code{C-c f}, which was bound to @code{ess-insert-function-outline} in
+@code{ess-add-MM-keys}; and @code{C-c h}, which was bound to
+@code{ess-handy-commands} in @code{Rd-mode-map},
+@code{ess-noweb-minor-mode-map}, and @code{ess-help-mode-map}
+
 @end itemize
 
 
--- a/lisp/ess-help.el
+++ b/lisp/ess-help.el
@@ -611,7 +611,6 @@ For internal use.  Take into account variable `ess-help-own-frame'."
     (define-key map "k" 'kill-this-buffer)
     (define-key map "?" 'ess-describe-help-mode)
     ;;-- those should be "inherited" from ess-mode-map ( ./ess-mode.el )
-    (define-key map "\C-ch"   'ess-handy-commands)
     (define-key map "\C-c\C-s" 'ess-switch-process)
     (define-key map "\C-c\C-r" 'ess-eval-region)
     (define-key map "\C-c\M-r" 'ess-eval-region-and-go)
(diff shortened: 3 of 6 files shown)


lisp/ess-s-lang.el
```diff
@@ -741,22 +741,6 @@ underscore character."
 ;; -----  The user can always customize `ess-S-assign' ...
 (ess-toggle-S-assign 'force-to-S-assign)
```
vspinu: This shouldn't be deleted. I believe @mmaechler is using this for teaching.

lionel-: That's a weird function to have in the package. Should this be in `ess-site`? Looking into its contents, wasn't that function-outline thing moved to /old?

jabranham: IMO stuff like this should definitely not be in core files like ess-s-lang or ess-site. Perhaps we could rename ess-font-lock.el, which also contains personal settings, to something like ess-personal and put this function (and others like it) there? 

And yes, the function-outline thing no longer exists for most users.

vspinu: Let's wait for @mmaechler to tell us if he still needs that function. Moving it to ess-site is a reasonable compromise indeed. 

ess-font-lock is ancient. I would also favor moving it to `./old` but I am afraid plenty of people might still rely on those.

lionel-: Perhaps we should load `lisp/old/` files from ess-site, and as soon as someone complains something doesn't work we remove it for good? This way we can keep old stuff around but out of sight and we make no promise to keep them working.

jabranham: @lionel- One reason I'm in favor of removing stuff like this is because it's all just one more thing to check when we rename stuff or fix new byte compile warnings.

Can we just `make-obsolete` and then remove it in the subsequent release?

vspinu: I think we should remove all at once. Gradual removals are extremely unnoing because they will break people's setup every time they update ESS. We don't want that. So maybe indeed, old/unesed stuff should just go into `./lisp/old` and be loaded from there from ess-site. 

At some point we will make melpa work without ess-site, entirely based on autoloads.

mmaechler: I'd pretty strongly vote for `make-obsolete`  -- fits what Emacs typically does and fits the quite consistent  deprecation -> defunctation  cycle (with sufficient waiting time in between) that we use in R.

Specifically as you asked:  I'm pretty sure I'm using  `ess-font-lock` quite extensively, am slightly less sure about the toggle-ess-assign / underscore issue.
I'm too busy these days, travelling all day Sunday finishing a talk to be held, and will not be able to look into this... and yes, I would be  really grateful if you don't remove these just now: I'm constantly working the github version of ESS and when it breaks during my preparation of talk, I'd be extremely unhappy... in particular as I know much less how to go back to a previous version than with svn where that was trivial for me.

jabranham:
OK, pushed a new commit that only removes the keybindings. I'll open up a new issue about the deprication/removal of old functions/files.

vspinu:
Could you please add a news entry on this? All breaking changes should be documented.

jabranham:
Done, thanks

lisp/ess-s-lang.el
```diff
@@ -747,7 +747,6 @@ underscore character."
   (interactive)
   (require 'ess-mode); typically unnecessary
   (require 'ess-inf); dito
-  (define-key ess-mode-map          "\C-cf" 'ess-insert-function-outline)
```
vspinu: Sorry, but why is this one deleted? Whoever calls `ess-add-MM-keys` expects that key to be defined. I think it should be left alone. Plus the next line binds `C-c w` which you left as it is.

jabranham: Because that function doesn't exist anymore (unless users load old/ or add it manually).

vspinu: That's weird :/ Then we should surely delete the key, indeed.

vspinu:
All clean. Thanks!

PR #544, opened 2018-05-20
Reduce Makefile verbosity

lionel-:
* Define a shorter `COMPILE` envvar
* Don't display `require` messages when compiling

The latter makes compiling less verbose and by the same token compilation warnings stand out.


vspinu:
:+1:

PR #545, opened 2018-05-20
Fix versioning of Emacs in Makefile

lionel-:
Includes #544 

This avoids warnings on older Emacsen.

In most cases the warnings were harmless because these files are not used on old Emacs versions. However sometimes `emacs` does not point to the right version and `make` compiles with an older emacs than intended. This then leads to issues when running ESS with a newer Emacs than used for compiling. Avoiding compilation in that case is safer.

### DIFF
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -28,9 +28,8 @@
 
 ;;; Code:
 
-(when (>= emacs-major-version 25)
-  (require 'subr-x)
-  (require 'xref))
+(require 'subr-x)
+(require 'xref)
 
 (require 'ess-utils)
 (require 'ess-r-package)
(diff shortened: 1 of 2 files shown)


lisp/ess-r-xref.el
```diff
@@ -28,9 +28,8 @@
 
 ;;; Code:
 
-(when (>= emacs-major-version 25)
-  (require 'subr-x)
-  (require 'xref))
+(require 'subr-x)
```
vspinu: This file could be loaded on emacs 24 as well. Also ELPA compiles it. Maybe `(require xref t)`?

lionel-: I thought it wouldn't be loaded because requires are guarded. MELPA doesn't use our Makefile?

vspinu: I think they just compile all el files within the directory.

vspinu: @lionel- what's the state of this? I would suggest we leave this part as it is. It's a good marker for the time when we finally drop emacs 24.

PR #546, opened 2018-05-21
Prettify-symbols should be the same width as what they replace

jabranham:
Otherwise we get the situation where

```R
a <- mean(mtcars,
          na.rm = TRUE)
```

is indented as

```R
a <- mean(mtcars,
         na.rm = TRUE)
```

when prettify-symbols-mode is on.

This is taken more or less directly from:
http://endlessparentheses.com/using-prettify-symbols-in-clojure-and-elisp-without-breaking-indentation.html


lionel-:
I think this fix is worthy of a news entry.

jabranham:
OK, added

vspinu:
Cool. What do those Br . Bl mean?

jabranham:
They're a way of creating reference points:

```
The meaning of glyph reference point codes is as follows:

    0----1----2 <---- ascent	0:tl or top-left
    |         |			1:tc or top-center
    |         |			2:tr or top-right
    |         |			3:Bl or base-left     9:cl or center-left
    9   10   11 <---- center	4:Bc or base-center  10:cc or center-center
    |         |			5:Br or base-right   11:cr or center-right
  --3----4----5-- <-- baseline	6:bl or bottom-left
    |         |			7:bc or bottom-center
    6----7----8 <---- descent	8:br or bottom-right
```

For more info, see `compose-region` and `reference-point-alist`.

PR #547, opened 2018-05-22
Obsolete ess-r-args.el

### LINKED ISSUE #428: ess-r-args-insert removes quotation marks (badly)
The utility function `ess-r-args-insert` in "ess-r-args.el", which inserts parameters and their default values to R functions, removes quotation marks (`"`) from default values. This leads to syntactically invalid expressions and strange results for nested quotes (including problems with code highlighting).

For example, the result for `read.table` is this (after `quote=` the highlighting of strings gets confused):
```
read.table(file=, header=FALSE, sep=, quote="', dec=., numerals=c("allow.loss", "warn.loss", "no.loss"), row.names=, col.names=, as.is=!stringsAsFactors, na.strings=NA, colClasses=NA, nrows=-1, skip=0, check.names=TRUE, fill=!blank.lines.skip, strip.white=FALSE, blank.lines.skip=TRUE, comment.char=#, allowEscapes=FALSE, flush=FALSE, stringsAsFactors=default.stringsAsFactors(), fileEncoding=, encoding=unknown, text=, skipNul=FALSE)
```

jabranham:
With the excellent eldoc support, I don't see any reason to keep `ess-r-args`. This PR moves it to `old/`, demotes a few `defcustom`s to `defvar`s and makes them obsolete. 

Closes #428 (by smashing the problem with a hammer, admittedly)

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -85,7 +85,6 @@ removed. This affects @code{C-c h}, which was bound to
 @code{ess-handy-commands} in @code{Rd-mode-map},
 @code{ess-noweb-minor-mode-map}, and @code{ess-help-mode-map}
 
-
 @item prettify-symbols-mode no longer breaks indentation
 This is accomplished by having the pretty symbols occupy the same
 number of characters as their non-pretty cousins.  You may customize
@@ -94,6 +93,9 @@ behavior.
 
 @item Variable @code{ess-s-versions-list} is obsolete and ignored.
 Use @code{ess-s-versions} instead.  You may pass arguments by starting the inferior process with the universal argument.
+
+@item The @code{ess-r-args.el} library has obsoleted and will be removed in a future release.
+Use @code{eldoc-mode} instead, which is on by default.
 
 @item All of the @code{*-program-name} variables have been renamed to @code{*-program}.
 Users who previously customized e.g. @code{inferior-ess-R-program-name}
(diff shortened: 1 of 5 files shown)


lionel-:
I think the functions should be marked obsolete. From last discussion, we're going to load all /old files from `ess-site` for a few releases before removing them all at once eventually.

jabranham:
OK, done

jabranham:
BTW if we rename `old/` to `obsolete/`, then Emacs automatically gives obsolescence warnings when those files get loaded.

vspinu:
Things in old are not available on melpa, which makes moving-to-old strategy somewhat unsatisfactory. Shall we have an obsolete dir which is pulled by melpa, and in due time move stuff from obsolete to old?

jabranham:
Don't we want to remove these things from MELPA, though? By putting them in old, they're still available for people to download or copy/paste. We just make it harder to use obsolete things that will break soon (or, in this case, are already partially broken). 

I don't see how putting things in old/ but still including & loading them in every install is different from just continuing to support them. I assume most people aren't looking to see if ESS files are loaded from `/path/to/ess/lisp` or `/path/to/ess/lisp/old`.

lionel-:
This is to deprecate stuff before removing them. Putting them in a subfolder is just a convenience for developers. This way they are easy to track and don't produce warnings since they are not compiled.

lionel-:
Yeah, this is about a new deprecation strategy suggested by Vitalie so that things break all at once rather than little by little.

jabranham:
Gotcha. In that case, ESS will have to add `obsolete/` to the `load-path` manually, I think. I guess `ess-site` is the best place to do that. 

We'll also need to adjust the MELPA recipe. Assuming that #543 gets merged, the `:files` bit should look like this I think:

```
  ("*.el" "lisp/*.el" ("etc" "etc/*") ("obsolete" "lisp/obsolete*.el")
      "doc/*.texi" "doc/info/dir")
```

jabranham:
Anyone object to moving the `ess-r-args-*` (now-obsolete) variables out of `ess-custom.el` and into `obsolete/ess-r-args.el`?

vspinu:
Sorry, I missed your question. Sure, let's do that.

jabranham:
Done and rebased off master. Ready to be merged after #543 is

PR #549, opened 2018-05-23
Improve finding binaries

### LINKED ISSUE #113: ESS/Julia Crashes Aquamacs
I am running OS X 10.9.4, Julia v0.3.

I installed the latest Aquamacs, deleted the ESS that comes with it and installed the Git development branch. My `Preferences.el` contains

```
   (add-to-list 'load-path "~/Dropbox/Admin/Aquamacs/ESS/lisp/")
   (load "ess-site")
```

Julia is already in the path; the version of ESS installed is 14.08.

When I start julia with `M-x julia RET`, a  Julia console opens in the _Terminal_ (not in Emacs - is that correct?), and Aquamacs becomes unresponsive to any input. I have to kill it.

The same happens with Vincent Goulet's version of Emacs from
 http://vgoulet.act.ulaval.ca/en/emacs/mac/.

Thank you for any pointers.

jabranham:
An attacker could currently get ESS users to unwittingly run a malicious program by setting the `-program-name` variable in a file. This renames all of them to `-program`, which makes them risky variables automatically. 

A few other improvements:
- Remove `inferior-R-version`, since it's not used anywhere
- Fix defining R runners (mentioned in #536)
- Use absolute paths for executables when possible (closes #113, I think, since I don't think we can do any better to correctly detect the executable).

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -60,6 +60,12 @@ number of characters as their non-pretty cousins.
 @item Variable @code{ess-s-versions-list} is obsolete and ignored.
 Use @code{ess-s-versions} instead.  You may pass arguments by starting the inferior process with the universal argument.
 
+@item All of the @code{*-program-name} variables have been renamed to @code{*-program}.
+Users who previously customized e.g. @code{inferior-ess-R-program-name}
+will need to update their customization to
+@code{inferior-ess-R-program}. These variables are treated as risky
+variables.
+
 
 @end itemize
 
--- a/lisp/ess-arc-d.el
+++ b/lisp/ess-arc-d.el
@@ -50,7 +50,7 @@
     (ess-help-sec-keys-alist       .  " ")
     (inferior-ess-primary-prompt   .  "> ?"               )
     (comint-use-prompt-regexp      . t)
-    (inferior-ess-program          .  inferior-ARC-program-name)
+    (inferior-ess-program          .  inferior-ARC-program)
     (inferior-ess-help-command     .  "(help '%s)\n"      )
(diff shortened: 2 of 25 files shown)


doc/newfeat.texi
```diff
@@ -60,6 +60,12 @@ number of characters as their non-pretty cousins.
 @item Variable @code{ess-s-versions-list} is obsolete and ignored.
 Use @code{ess-s-versions} instead.  You may pass arguments by starting the inferior process with the universal argument.
 
+@item All of the @code{*-program-name} variables have been renamed to @code{*-program}.
```
vspinu: Well, this will mean a lot of users actually. I think we should alias. I am not sure about -program postfix either. I think `-executable` is way more common.  Let's see what others think.

jabranham: > I think we should alias

Sure, that's easy enough. My thinking in not doing so was to make them aware of it in the least painful way possible. I'll add the (obsolete?) aliases shortly though.

> I think -executable is way more common. 

On my system, I have more -program than -executable variables. I only have 8 `-executable`, and perhaps 50ish `-program` (not counting ESS's, of course).  Another reason I (slightly) prefer `-program` is because elisp automatically marks these as risky, whereas it doesn't with -executable.

At the end of the day though, I really don't care that much about the name.

jabranham:
OK, aliases added

doc/newfeat.texi
lionel-: > Another reason I (slightly) prefer -program is because elisp automatically marks these as risky, whereas it doesn't with -executable.

I like this so I'd say to go with `-program`.

vspinu:
Looks good. Thanks!

PR #559, opened 2018-06-08
Simplify ess-smart-S-assign and ess-smart-S-assign-key

### LINKED ISSUE #558: add a configure about ess-smart-S-assign
If one don't want to use `ess-smart-S-assign-key`, he must use some `toggle` function.  But this is not explicit. For example, if some one write it twice in init file, it will start again.

Would you like to add some function like `(ess-smart-S-assign-mode nil)` to disable it indeed.

Or let we set `ess-smart-S-assign-key` to `nil`. I try to this but it will throw a exception when `require` this package.

jabranham:
This changes is so that users can (setq ess-smart-S-assign-key nil) at
any time to disable the smart assignment. To change which key is used,
they must set ess-smart-S-assign-key before ESS is loaded.

If ess-smart-S-assign-key is nil, ess-smart-S-assign behaves like
self-insert-command.

It keeps around the old functions so that they don't cause errors, but
they're ignored in favor of having users set ess-smart-S-assign-key
themselves.

Closes #558

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1957,12 +1957,10 @@ for `ess-eval-region'."
 (defvar inferior-ess-mode-map
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map comint-mode-map)
-
     (define-key map "\C-y"              'ess-yank)
     ;; Use syntax valid *both* for GNU emacs and XEmacs :
     (define-key map "\r"       'inferior-ess-send-input)
     (define-key map "\C-a"     'comint-bol)
-
     ;; 2010-06-03 SJE
     ;; disabled this in favour of ess-dirs.  Martin was not sure why this
     ;; key was defined anyway in this mode.
@@ -1992,6 +1990,8 @@ for `ess-eval-region'."
     (define-key map "\C-c\C-d"   'ess-doc-map)
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
+    (when ess-smart-S-assign-key
+      (define-key map ess-smart-S-assign-key #'ess-smart-S-assign))
     map)
   "Keymap for `inferior-ess' mode.")
 
--- a/lisp/ess-s-lang.el
+++ b/lisp/ess-s-lang.el
@@ -219,18 +219,6 @@
 (defconst ess-help-S+-sec-regex "^[A-Z. ---]+:$"
   "Reg(ular) Ex(pression) of section headers in help file.")
 
-;;; S-mode extras of Martin Maechler, Statistik, ETH Zurich.
-;;; See also ./ess-utils.el
-
-;; Seth's idea; see ess-toggle-S-assign-key below
-(defvar ess-S-assign-key [?\C-=] ;; = "\C-c=" ; old-default:  "_"
-  "This key is mapped to insert `ess-S-assign' (by default '<-'),
-when \\[ess-toggle-S-assign-key] is called.")
-
-(defvar ess-S-assign-key-last nil
-  "This caches the previous value (binding) of `ess-S-assign-key'.  It allows
- \\[ess-toggle-S-assign-key] to toggle back to the previous definition.")
-
  ; Function Definitions
 
 (defun S-comment-indent ()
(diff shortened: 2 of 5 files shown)


lisp/ess-s-lang.el
```diff
@@ -694,52 +658,18 @@ In that case, it is removed and replaced by `ess-smart-S-assign-key'.
...
+(make-obsolete 'ess--unset-smart-S-assign-key
+               "it does nothing. Set `ess-smart-S-assign-key' to nil." "2018-06-08")
+(defalias 'ess-disable-smart-S-assign 'ignore "")
+(make-obsolete 'ess-disable-smart-S-assign
+               "it does nothing. Set `ess-smart-S-assign-key' to nil." "2018-06-08")
+(defalias 'ess--activate-smart-S-assign-key 'ignore "")
+(make-obsolete 'ess--activate-smart-S-assign-key
+               "it does nothing. Set `ess-smart-S-assign-key' instead." "2018-06-08")
+(defalias 'ess-toggle-S-assign 'ignore "")
+(make-obsolete 'ess-toggle-S-assign
```
vspinu: I like the change but this one should be setting ess-smart-S-assign-key to nil. Otherwise this will lead to a lot of breakage. This  "feature" is already annoying to a lot of users, let's not irritate them yet again on this occasion.

jabranham: By "this one" you mean `ess-toggle-S-assign` should actually just do something like `(setq ess-smart-S-assign-key nil)`, right? That's an easy enough change to make.

jabranham:
@vspinu OK, I pushed a commit that provides `ess-disable-smart-S-assign` as a transition function. `ess-toggle-S-assign` and `ess--unset-S-assign` are aliased to it.

lisp/ess-s-lang.el
```diff
@@ -777,21 +774,21 @@ If 'sp-for-help-changed?' process variable is non-nil or
...
-          (ess-process-get 'sp-for-help-changed?))
-      (ess-process-put 'sp-for-help-changed? nil)
-      (setq ess-help-topics-list
-            (ess-uniq-list
-             (append (ess-get-object-list name 'exclude-1st)
-                     (ess-get-help-files-list)
-                     (ess-get-help-aliases-list)))))
-     (t
-      ess-help-topics-list))))
+                           (ess-write-to-dribble-buffer
```
lionel-: Why the whithespace change?

lisp/ess-inf.el
```diff
@@ -1992,6 +1990,8 @@ for `ess-eval-region'."
     (define-key map "\C-c\C-d"   'ess-doc-map)
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
+    (when ess-smart-S-assign-key
```
lionel-: This gives me a strange feeling because it seems people should use standard ways of settings keybindings instead. But in this case I guess it's a good compromise...

I wonder if we should have some sort of parent map that both `ess-mode-map` and `inferior-ess-mode-map` would inherit from. The common keybindings would be mapped there and this way you only need to customise one place.

lisp/ess-s-lang.el
jabranham: Sorry, didn't notice that. Should be fixed now.

lisp/ess-inf.el
jabranham: I agree it's a bit odd. Other modes do define them separately, though. For example, there's `python-mode`, derived from `prog-mode`, and `inferior-python-mode`, derived from `comint-mode`.

lionel-: oh yeah, that is the reason, thanks!

MilesMcBain:
Hi, I've got errors preventing ESS mode loading relating to this change. E.g:

`File mode specification error: (wrong-number-of-arguments #[nil \301\211\207 [ess-smart-S-assign-key nil] 2 (/home/miles/.emacs.d/elpa/ess-20180616.216/lisp/ess-s-lang.elc . 12236)] 1)`

I guess because I call `ess-toggle-underscore nil` in `with-eval-after-load 'ess-site`. 

The documentation added with this PR does not help me understand what I need to do to fix this.

jabranham:
@MilesMcBain Are you trying to disable `ess-smart-S-assign-key`? If so, the best way to do so is with `(setq ess-smart-S-assign-key nil)`.

MilesMcBain:
I've added a PR that improves the documentation of the new feature for others in a similar boat: https://github.com/emacs-ess/ESS/pull/578

PR #560, opened 2018-06-08
Test on 26.1

jabranham:
Add 26.1 to travis test.

What's the plan to drop support for older Emacs versions? We currently say that 24.3 and older are supported. 24.3 was released in March 2013, over 5 years ago. Even Debian stretch is shipping 24.5 (from 2015) now as their oldest Emacs. They also ship Emacs 25.

I vote we add a note to `newfeat` that says that the upcoming release (whenever that is) will be the last to support Emacs < 25.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: emacs-lisp
 env:
   - EVM_EMACS=emacs-24.3-travis
   - EVM_EMACS=emacs-25.3-travis
+  - EVM_EMACS=emacs-26.1-travis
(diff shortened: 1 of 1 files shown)


vspinu:
Indeed. Let's do that.

lionel-:
Let's drop support for 24.3 right away. I stumbled on an issue with ERT conditions on 24.3 and spent too much time trying to fix it.

vspinu:
We cannot drop it without releasing first. There has been too many additions since last release.

Just disable those tests on 24.3. Or remove 24.3 from the travis matrix if that helps.

PR #561, opened 2018-06-08
Add announcement that we are dropping support for Emacs < 25.1

jabranham:
As discussed in #560


vspinu:
Merged through a rebase. Thanks!

PR #566, opened 2018-06-11
Rethink control flow keywords

lionel-:
* `ess-R-keywords` now includes `on.exit()`, `tryCatch()`, `withRestarts()`, `invokeRestart()`, `recover()`, `browser()`.

* Add new `ess-R-weak-keywords` with a new face inheriting from `ess-modifiers-face`. It includes: `message()`, `warning()`, `signalCondition()`, `withCallingHandlers()`.

The goal is to fontify differently keywords that (a) cause a control flow jump or represent a jump target and (b) might cause a control flow jump.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -14,15 +14,17 @@ opening parenthesis. The same holds for search path modifiers like
 Emacs >= 25.
 
 @item @ESS{[R]}: We have improved fontification of keywords so they
-better reflect the semantics of the R language. The following variables
-have been added to @code{ess-R-keywords}: @code{on.exit()},
-@code{tryCatch()}, @code{withRestarts()}, @code{invokeRestart()},
-@code{recover()} and @code{browser()}. This variable now refers to
-functions that either cause a jump in control flow or represent a jump
-target.
+better reflect the semantics of the R language. @code{ess-R-keywords}
+now only contains words reserved by the R parser.
+@code{ess-R-control-flow-keywords} contains words of base functions that
+cause non-contiguous control flow, such as @code{return()} and
+@code{stop()}. It includes the following variables that were previously
+not fontified: @code{on.exit()}, @code{tryCatch()},
+@code{withRestarts()}, @code{invokeRestart()}, @code{recover()} and
+@code{browser()}
 
-In addition, the new @code{ess-R-weak-keywords} variable holds functions
-that might cause a jump but not necessarily. It includes
+Finally, @code{ess-R-signal-keywords} contains functions part of the
+condition system that only potentially impact control flow:
 @code{message()}, @code{warning()} (moved from @code{ess-R-keywords}),
 @code{signalCondition()} and @code{withCallingHandlers()}. These
 keywords inherit from @code{ess-modifiers-face} (the face used for
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -21,6 +21,13 @@ have been added to @code{ess-R-keywords}: @code{on.exit()},
 functions that either cause a jump in control flow or represent a jump
 target.
 
+In addition, the new @code{ess-R-weak-keywords} variable holds functions
+that might cause a jump but not necessarily. It includes
+@code{message()}, @code{warning()} (moved from @code{ess-R-keywords}),
+@code{signalCondition()} and @code{withCallingHandlers()}. These
+keywords inherit from @code{ess-modifiers-face} (the face used for
+@code{library()} etc).
+
 @item ESS modes now inherit from @code{prog-mode}.
 
 @item @ESS{[R]}: The package development minor mode now only activates
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -13,6 +13,14 @@ opening parenthesis. The same holds for search path modifiers like
 @code{library()} or @code{require()}. This feature is only available in
 Emacs >= 25.
 
+@item @ESS{[R]}: We have improved fontification of keywords so they
+better reflect the semantics of the R language. The following variables
+have been added to @code{ess-R-keywords}: @code{on.exit()},
+@code{tryCatch()}, @code{withRestarts()}, @code{invokeRestart()},
+@code{recover()} and @code{browser()}. This variable now refers to
+functions that either cause a jump in control flow or represent a jump
+target.
+
 @item ESS modes now inherit from @code{prog-mode}.
 
 @item @ESS{[R]}: The package development minor mode now only activates
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2647,7 +2647,13 @@ This variable has no effect. Customize
 
 (defvar ess-R-keywords
   '("in" "else" "break" "next" "while" "for" "if" "switch"
-    "function" "return" "message" "warning" "stop"))
+    "function" "return" "on.exit" "stop"
+    "tryCatch" "withRestarts" "invokeRestart"
+    "recover" "browser")
+  "Keywords that impact control flow.
+These keywords either cause a control flow jump or establish a
+jump target.")
+
 
 (defvar ess-S-keywords
   (append ess-R-keywords '("terminate")))
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2646,17 +2646,20 @@ This variable has no effect. Customize
           '("T" "F")))
 
 (defvar ess-R-keywords
-  '("in" "else" "break" "next" "repeat" "while" "for" "if"
-    "switch" "function" "return" "on.exit" "stop"
+  '("if" "else" "repeat" "while" "function" "for" "in" "next" "break")
+  "Reserved words in the R language.")
+
+(defvar ess-R-control-flow-keywords
+  '("switch" "function" "return" "on.exit" "stop"
     "tryCatch" "withRestarts" "invokeRestart"
     "recover" "browser")
   "Keywords that impact control flow.
 These keywords either cause a control flow jump or establish a
 jump target.")
 
-(defvar ess-R-weak-keywords
+(defvar ess-R-signal-keywords
   '("message" "warning" "signalCondition" "withCallingHandlers")
-  "Keywords that possibly impact control flow.
+  "Keywords for condition signalling.
 These keywords might cause a control flow jump but do not necessarily.")
 
 (defvar ess-S-keywords
@@ -2812,9 +2815,13 @@ default or not."
             '(1 ess-keyword-face))
       "Font-lock keywords that precede an opening parenthesis.")))
 
-(defvar ess-R-fl-keyword:weak-keywords
-  (cons (concat "\\(" (regexp-opt ess-R-weak-keywords 'words) "\\)\\s-*(")
-        '(1 ess-weak-keyword-face)))
+(defvar ess-R-fl-keyword:control-flow-keywords
+  (cons (concat "\\(" (regexp-opt ess-R-control-flow-keywords 'words) "\\)\\s-*(")
+        '(1 ess-r-control-flow-keyword-face)))
+
+(defvar ess-R-fl-keyword:signal-keywords
+  (cons (concat "\\(" (regexp-opt ess-R-signal-keywords 'words) "\\)\\s-*(")
+        '(1 ess-r-signal-keyword-face)))
 
 (defvar ess-R-fl-keyword:assign-ops
   (cons (regexp-opt ess-R-assign-ops) 'ess-assignment-face)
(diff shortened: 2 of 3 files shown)


lionel-:
> What do you think about "soft-keywords" instead of "weak-keywords"? It's a bit more suggestive of the softer color for them.

I have a slight preference for "weak" but don't feel strongly either way.

vspinu:
Ok, "weak" then. Neither is particularly suggestive, but I guess we are unlikely to do better here. Waiting for @mmaechler to approve.

lionel-:
> on.exit() looks wrong in this list to me:

It definitely causes a jump in control flow, and I think it is right to have it in that list. Even if it weren't the case, it would deserve to be highlighted because that's the place where you clean up stuff. `on.exit()` has a very special place in R semantics.

>  found that list is missing repeat
> another one, not a reserved word, but somewhat special is Recall().

Makes sense.

> In some sense "keyword" really only applies to such reserved words ... and we should use a different term for simple functions such as "tryCatch()"... and I am still not sure they should be all listed.

I think making that list about actual keywords is not interesting. R has few keywords and we'd still like more symbols highlighted. Making it about control flow provides a simple rationale.

`tryCatch()`, unlike `try()`, is a primitive function (not in the sense of PRIMSXP but still) that installs handlers and provides a *jump target* for handlers. The keywords list in this proposal is about functions that either cause a jump or provide a jump target.

> What about assertError(), assertCondition() from "tools" package ? Why not also treat stopifnot() specially? The list seems to never end ...

In my mind these are all wrappers, at least conceptually (think of functions calling `Rf_error()` as if they were calling `stop()`). It is true of all R functions that they might not return but the list is about functions that actually cause the control flow change. `stop()` is special in a way that `stopifnot()` is not. So the list does have a manageable end.

gdkrmr:
what about making `&&` and `||` control flow keywords?

lionel-:
> what about making && and || control flow keywords?

Maybe, though I _think_ these should stay fontified like other infix operators.

>> another one, not a reserved word, but somewhat special is Recall().
>
> Makes sense.

Hmm on second thought it doesn't represent a cflow jump in the sense defined in the proposal but just a function call with some introspective magic.

lionel-:
> In some sense "keyword" really only applies to such reserved words ... and we should use a different term for simple functions such as "tryCatch()"

Maybe we should have `ess-R-strong-keywords` for the non-reserved keywords, which would go well with the `-weak-` variant? The default would be about cflow but could be customised.

vspinu:
I think if we widen the definition of "control flow" to "non-contiguous-flow" then `on.exit` would integrate just fine. 

AFAIU @lionel- original motivation is to highlight "message" and "warning" in a different font than the strong control flow operators. It's defenitely true that all functions can cause jump, but if we focus on the most commonly used control flow keywords the proposed definition leads to enhanced fontification by emphasizing non-contiguous sections of the code.  

> ... or provide a jump target.

Why is this useful? The code within tryCatch is localized and doesn't cause a jump outside of the body of the function, so I share @mmaechler's doubts here. On the other hand tryCatch is similar to switch in the sense that any of the installed handlers can be called, and `switch` is a keyword already. But then, withCallingHandlers is also similar to `switch` in this regard. So `withCallingHandlers`, `tryCatch`, `withRestarts` and `switch` belong to the same group.

How about calling `keywords` all control-flow and switch-like functions and having `ess-signal-keywords` for all those functions which do signal and are not already in `ess-keywords`? So basically, the current PR but moving `withCallHandlers` into keywords?

vspinu:
> I think if we widen the definition of "control flow" to "non-contiguous-flow" then on.exit would integrate just fine.

Actually `tryCatch`, `switch`, `withCallHandlers` are non-contiguous-flow constructs as well.

lionel-:
> any of the installed handlers can be called

I think this might be too large and vague a criterion. With lazy evaluation any arguments can be evaluated or not, and with first class functions any function passed as arguments can be evaluated or not. If I were to redesign the list from scratch I think I wouldn't include `switch`.

> tryCatch, switch, withCallHandlers are non-contiguous-flow constructs as well.

I'm not sure we should think about those in that way. Otherwise any higher order function like `lapply()` falls into the category.

> Why is this useful? The code within tryCatch is localized and doesn't cause a jump outside of the body of the function

hmm... otoh this reasoning also pertains to `for()` which always returns. `tryCatch()` install handlers on the stack and as such is not a normal function call. It is also significant that it provides a jump target and causes signals to jump I think.

vspinu:
> ith lazy evaluation any arguments can be evaluated or not, 
> Otherwise any higher order function like lapply() falls into the category.

I believe we should think of these issues in terms of the lexical code layout, not in terms of what is the evaluation semantics of the function is. To my mind, the real benefit of your proposal is to make visual detection of non-contiguous code easier. `for`, `switch`, `tryCatch`, `on.exit` all introduce non-contiguous fragments.

Just like any function can throw, any function can evaluate arguments in different order. Evaluation style arguments are unlikely to bring us to a satisfactory conclusion.

lionel-:
> for, switch, tryCatch, on.exit all introduce non-contiguous fragments.

I mostly agree (just that that switch falls into non-contiguous because of lazy eval and we don't want to consider lazy eval effects as non-contiguous flow for the purpose of these keywords, but that's a detail I'm happy to overlook).

On the other hand `withCallingHandlers()` is different to all the rest. It is merely an indication that cflow might jump because we are installing handlers whose typical purpose is to jump (in order to handle the condition). That's why I thought it'd fit in the weak keywords list.

vspinu:
>  switch falls into non-contiguous because of lazy eval

I don't really understand why lazy eval is important for defining switch as non-congihous. Switch can evaluate any of the branches and that does mean "jump over portions of code without evaluating those".

> On the other hand withCallingHandlers() is different to all the rest. 

I surely have looser understanding of these issues than you do, but to my mind withCallingHandlers is no more different than tryCatch, or withRestarts. There are branches and any of those could be or will be executed at some point in time. This is why I abstract them into "switch" branching category. I think any more technical distinction is not warranted for the purpose at hand, especially that 99.9% of users will never use those. 

One big benefit of moving withCallingHandlers is that we could stick with a more suggestive `ess-signal-keywords` instead of `ess-weak/soft-keywords`.

mmaechler:
> One big benefit of moving withCallingHandlers is that we could stick with a more suggestive ess-signal-keywords instead of ess-weak/soft-keywords.

I cannot get at ease with "keywords" in this context.  Maybe  `ess-signalling-functions` ?

lionel-:
>  Switch can evaluate any of the branches and that does mean "jump over portions of code without evaluating those".

Any function can do that thanks to lazy evaluation. It allows functions to choose which argument is evaluated. Lazy eval is a major cause of non-contiguous flow in R, that's why the backtraces can be a little wonky.

> withCallingHandlers is no more different than tryCatch, or withRestarts

tryCatch and withRestarts are jump targets while withCallingHandlers is not. They all have in common the side effect of registering functions on the context stack. tryCatch is a cause of jumps while withRestarts and withCallingHandlers are merely providers for causes of jumps.

> we could stick with a more suggestive ess-signal-keywords instead of ess-weak/soft-keywords

I'm not necessarily against hard-coding the semantics into the variable names, but keeping them generics would have the advantage of letting the user decides what semantics to give to those keywords. Otoh they can also build and install their own sets of fontlock keywords, so for clarity of UI your suggestion might be better. But `withCallingHandlers()` would still nicely fit in a `signal-keywords` variable (it installs handlers for signals).

lionel-:
> I cannot get at ease with "keywords" in this context. Maybe ess-signalling-functions ?

It really should be keywords. This is not about keywords in the `?keywords` sense but in the font-lock sense. `-keywords` is the conventional name pattern.

vspinu:
> Any function can do that thanks to lazy evaluation. It allows functions to choose which argument is evaluated.

Any function can throw and that doesn't make it a candidate for cflow keywords.  Any function can evaluate argument in different order but that doesn't make it a candidate for cflow either. Trying to define a taxonomy based on internal evaluation semantics of the function is not a fruitful direction. I am trying to focus on the syntax here and immediately apparent top-level evaluation differences. 

I think syntax is the most important criteria for the font-lock decisions. Someone with zero knowledge of R should be able to see patterns and meaning in font-lock colors. 

>  tryCatch and withRestarts are jump targets while withCallingHandlers is
not. They all have in common the side effect of registering functions on the context stack.

Most of the users won't know how these functions work internally. I think we can do better than classifying functions based on specific internals.

> tryCatch is a cause of jumps 

Not sure what you mean by this.  "tryCatch is a provider of jump targets" and the cause of jump is the actual function throwing an error.

The fact that tryCatch and withRestarts provide jump targets is not that relevant. All these functions (including `withCallingHandlers`) provide hooks (functions) which are called at some point in time if some condition (signals) happened. They all lead to non-contiguous flow which is immediately apparent from the syntax of those functions.

Of course, one can jump and say "any function which accepts a callback does that", but well...

> but keeping them generics would have the advantage of letting the user decides what semantics to give to those keywords

True, but how likely is that to happen? And how much will they care when they put a non-signaling function into that list? Most likely they will register their own keywords directly with `font-lock-add-keywords` as most of the examples online do. 

> But `withCallingHandlers()` would still nicely fit in a`signal-keywords` variable (it installs handlers for signals).

True, but it's not signaling anything. It's a tradeoff that we have to make. Imagine for a moment that you don't know anything about the internals of those functions. Would you still see withCallingHandlers naturally fit into that group?

lionel-:
>  Trying to define a taxonomy based on internal evaluation semantics of the function is not a fruitful direction. I am trying to focus on the syntax here and immediately apparent top-level evaluation differences.

I agree about all this and I was making the same points. I just think this should lead us to remove `switch()` from the keywords because it's a regular function on all counts. But as I said I'm ok to keep it because it's a pure special case of selective evaluation and is clearly used for cflow.

> Most of the users won't know how these functions work internally. I think we can do better than classifying functions based on specific internals.

Again I don't want to classify based on internals but on semantics. Jumps + jump target is purely about semantics.

> They all lead to non-contiguous flow

`withCallingHandlers()` really doesn't, it's merely an indication. If you ignore the effects on the stack (installing handlers for signals) it's equivalent to any higher order function. It loops over all handlers it was passed and call each of them in turn, discarding results. I still think it should be in the weak keywords because of its effect on the stack, which makes it a special function in the language, and also because the calling handlers it installs can be expected to jump.


> Imagine for a moment that you don't know anything about the internals of those functions. Would you still see withCallingHandlers naturally fit into that group?

Someone using withCallingHandlers should read about what it does (and not its internals, which I don't care about for the purpose of these variables). Knowing that it installs handlers for signals is the least minimum for using that function.

lionel-:
It seems we are largely on the same page except for withCallingHandlers. I'm ok with renaming the variables to `ess-R-control-flow-keywords` and `ess-R-signal-keywords`. But I strongly feel that (a) it should be highlighted but (b) not as part of tho cflow keywords. I think it makes sense to treat it as a signal keyword. tryCatch and stop are also signal keywords but the cflow semantics should supplant the signal semantics and so belong to the latter keyword group.

vspinu:
Fine with me. But, we should alias ess-R-keywrods to ess-R-control-flow-keywords if you want to rename that one.

lionel-:
Thanks. As I understood Martin's comments, the idea would be that `ess-R-keywords` would only contain reserved keywords, then we'd add the function keywords in the new variables.

vspinu:
That might be a neat idea actually, but it will make it messy for folks who already customized their ess-R-keywords. The end effect should be the same if we keep the same font lock for ess-R-keywords and ess-R-control-flow-keywrods.

lionel-:
> but it will make it messy for folks who already customized their ess-R-keywords.

I don't think there's many users that do that because it's not possible to easily append to keywords right now, since the regexp will not be updated. So you have to define the whole list of keywords before loading ESS. I was planning to look into the "custom" features for dynamically updating dependent variables and make it easier to add keywords.

vspinu:
This PR broke highlighting for me. I no longer see stop, message and warning highlighted. All of these are highlighted in `ess-function-call-face`. Any ideas?

lionel-:
Hmm I don't see this, and I use function-call highlighting as well.

lionel-:
I am using this to enable it:

```
(setf (cdr (assq 'ess-fl-keyword:fun-calls ess-R-font-lock-keywords)) t)
```

Perhaps you append to the the fl keywords list and changing the order has an effect? Though the face is set with `keep` which should prevent it from overtaking other faces in all cases, I think.

vspinu:
I guess it has somethting to do with it. I have just selected keywords which I want from the menu and then "Save to custom". This is a canonical way of doing it.

lionel-:
@vspinu I have just tried it and it works for me. I can't think of another reason why it stopped working for you. Can you reproduce this on an empty config? And if yes, with an empty custom.el?

lionel-:
I meant add a new customisation on an empty custom.el. To find out if something in your custom.el triggers the issue.

PR #567, opened 2018-06-11
New defcustom ess-r-prettify-symbols

### LINKED ISSUE #132: Remove R arrow prettification
What is the proper way to customise `R-customize-alist`?

I need to delete the element relative to unicode prettification but I couldn't find a proper hook to run

```
(lambda () (setq R-customize-alist
                          (delete '(prettify-symbols-alist
                                    quote
                                    (("<-" . 8592)
                                     ("<<-" . 8606)
                                     ("->" . 8594)
                                     ("->>" . 8608)))
                                  R-customize-alist)))
```

The hooks I tested either run too early (`R-customize-alist` is not yet defined), or too late (the first R buffer I open, the one that causes ESS to load, has unicode arrows and I have to manually reload it to get rid of them).

Or is there some other way?

jabranham:
Allows users to change (and, if set to nil, stop) which symbols get
prettified in R buffers.

Closes #132

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -64,7 +64,9 @@ removed. This affects @code{C-c h}, which was bound to
 
 @item prettify-symbols-mode no longer breaks indentation
 This is accomplished by having the pretty symbols occupy the same
-number of characters as their non-pretty cousins.
+number of characters as their non-pretty cousins.  You may customize
+the new variable @code{ess-r-prettify-symbols} to control this
+behavior.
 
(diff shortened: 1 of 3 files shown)


lionel-:
👍

vspinu:
:+1:

jabranham:
Unless I'm misunderstanding something, I think ligatures should work out of the box. Did you try changing `ess-r-prettify-symbols` so that the values are the characters you want displayed?

PR #569, opened 2018-06-11
Simplify setting the working directory

### LINKED ISSUE #550: Changing directory should be visible and set `default-directory`
Not visible in the inferior:

* `ess-use-dir`

Doesn't update `default-directory`:

* `ess-use-dir`
* `ess-set-working-directory`

jabranham:
alias ess-use-dir to ess-set-working-directory since they do
essentially the same thing. Move ess-use-this-dir to ess-inf since it
can be used by non-S/R languages now.

Closes #550

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -3078,12 +3078,15 @@ NO-ERROR prevents errors when this has not been implemented for
   'ess-use-dir 'ess-set-working-directory "2018-06-11")
 
 (defun ess-use-this-dir (&optional no-force-current)
-  "Set the current process directory to `default-directory'.
-If that buffer has no associated *R* process, use
-\\[ess-force-buffer-current], unless prefix argument
-NO-FORCE-CURRENT is non-nil."
+  "Set the current process directory to the directory of this file.
+`default-directory' is used as a fallback.  If that buffer has no
+associated *R* process, use \\[ess-force-buffer-current], unless
+prefix argument NO-FORCE-CURRENT is non-nil."
   (interactive "P")
-  (ess-set-working-directory default-directory))
+  (let ((dir (if buffer-file-name
+                 (file-name-directory buffer-file-name)
+               default-directory)))
+    (ess-set-working-directory dir)))
 
 (defun ess-get-working-directory (&optional no-error)
   "Retrive the current working directory from the current ess process."
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -3053,8 +3053,9 @@ list."
 
  ; directories
 (defun ess-set-working-directory (path &optional no-error)
-  "Set the current working directory to PATH for both ESS
-subprocess and Emacs buffer `default-directory'."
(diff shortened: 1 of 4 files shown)


jabranham:
It would make more sense (to me) if `ess-use-this-dir` set the working directory to where the file is located, not where `default-directory` is. OK to make that change?

vspinu:
That makes sense. But it should fall back on default directory, something like

```el
(if buffer-file-name
      (file-name-directory buffer-file-name)
    default-directory)
```

jabranham:
Thanks, done

lisp/ess-inf.el
```diff
@@ -3073,6 +3074,20 @@ subprocess and Emacs buffer `default-directory'."
       (error "Not implemented for dialect %s" ess-dialect))))
 
 (defalias 'ess-change-directory 'ess-set-working-directory)
+(define-obsolete-function-alias
+  'ess-use-dir 'ess-set-working-directory "2018-06-11")
+
+(defun ess-use-this-dir (&optional no-force-current)
+  "Set the current process directory to the directory of this file.
+
```
lionel-: Small nitpick: there shouldn't be a newline after the leading sentence.

jabranham: Removed

mmaechler:
Hmm, but we already have the functionality of  your `ess-use-this-dir` -- even with the exact same name in `ess-s-lang.el` .. and I have been using it every day via `M-[Enter]` ... ???

jabranham:
@mmaechler Right, this moves the function into `ess-mode.el` since it can now be used by languages other than R/S. It doesn't change the `M-RET` binding in S mode, so it'll still be there.

mmaechler:
aah.. now I understand.  Ok.

jabranham:
@lionel-  I'm trying to add a test:

```emacs-lisp
(ert-deftest ess-set-working-directory ()
  (with-r-running nil
    (ess-set-working-directory "/")
    (insert "getwd()\n")
    (should (output= (ess-eval-buffer nil)
              "setwd('/')\n>\n> getwd()\n[1] \"/\""))))
```

But it's failing because the output comes back with fontification stuff:

```
F ess-set-working-directory
    (ert-test-failed "Expected: 
setwd('/')
>
> getwd()
[1] \"/\""
                     #("Result: 
setwd('/')
> 
> getwd()
[1] \"/\"" 9 19
(fontified nil font-lock-face comint-highlight-input front-sticky t field input)
19 20
(fontified nil font-lock-face comint-highlight-input)
20 22
(fontified nil)
22 33
(fontified nil)
33 40
(fontified nil)))
```

Is there a better way to do this?

jabranham:
Thanks. In the meantime, I pushed a test that tests `default-directory` gets changed.

lionel-:
Also you should send a command to the inferior rather than insert text in the buffer.

jabranham:
Thanks, that seems to make the test work, for me at least. We'll wait to see if travis complains.

lionel-:
Great, one more test towards good coverage!
Thanks for the PR.

PR #573, opened 2018-06-13
Do not make ess-transcript-mode interactive

### LINKED ISSUE #572: ess 17.10 fails to load {S,R} transcript file
On a GNU Emacs 25.3.1 (x86_64-w64-mingw32) of 2017-09-26,  ess-version: 17.11 [elpa: 17.11] (loaded from c:/Users/csrabak/.emacs.d/elpa/ess-17.11/)ess-version: 17.11 [elpa: 17.11] (loaded from c:/Users/csrabak/.emacs.d/elpa/ess-17.11/)

Loading a script file (extension .Rt) fails with following message (after an Alt-X ess-transcript-mode issuing in the already loaded file at current buffer):

funcall-interactively: Wrong number of arguments: #[(alist &optional proc) "ÆÇ!È ÉÊ	!ËÌÍ!Î


jabranham:
Closes #572

### DIFF
--- a/lisp/ess-trns.el
+++ b/lisp/ess-trns.el
@@ -155,8 +155,6 @@ Type \\[ess-transcript-clean-region] to delete all outputs and prompts
 in the region, leaving only the S commands.  Other keybindings are:
 
 \\{ess-transcript-mode-map}"
-  (interactive)
-  (require 'ess-inf)
   (kill-all-local-variables)
   (setq buffer-read-only t) ;; to protect the buffer.
   (ess-setq-vars-local alist); (current-buffer))


jabranham:
The file already requires ess-inf, so that can be eliminated too

vspinu:
:+1:

mmaechler:
Hmm, really?   I do want ess-transcript-mode  interactive in the sense that I send R code from transcript files.  You are not disabling that, are you?

jabranham:
Nope, this is about preventing users from calling M-x ess-transcript-mode since that will always error (they should be using R-transcript-mode instead)

mmaechler:
aah, yes, of course!  thank you!

PR #574, opened 2018-06-13
Remove ess-electric-brace

### LINKED ISSUE #389: ess-electric-brace is broken
[I recently posted an earlier version of this to ess-help, but that mail
was returned for some reason. And I've since been advised that this is a
better place to raise the issue. So please forgive the cross-posting if
that message does show up eventually.]

In recent versions of ESS (currently ESS 16.10-1, Emacs 25.1.1),
the electric braces do not work at all for me: no automatic pairing
even when skeleton-pair is t and no reindenting ever. Here I'm 
suggesting a fix for this and an extension that I think is even
more useful.

There are two immediate issues with the current `ess-electric-brace`.
First, it starts by checking if `(fboundp 'skeleton-pair-insert-maybe)`,
but it does not check the value of `skeleton-pair`, which if nil
suppresses pairing in general. It seems a bad idea to give precedence to
skeleton when that feature is effectively turned off. Second and more
seriously, `ess-electric-brace` passes "{" to `skeleton-pair-insert-maybe`, 
but that function expects an optional prefix arg, so passing "{" *always*
suppresses pairing. The following simple patch fixes both these issues:

    --- /tmp/original-ess-brace.el    2016-12-18 21:29:59.000000000 -0500
    +++ /tmp/modified-ess-brace.el    2016-12-18 21:29:46.000000000 -0500
    @@ -2,8 +2,9 @@
       "Insert character and correct line's indentation."
       (interactive "P")
       ;; skeleton-pair takes precedence
    -  (if (fboundp 'skeleton-pair-insert-maybe)
    -      (skeleton-pair-insert-maybe "{")
    +  (if (and (fboundp 'skeleton-pair-insert-maybe)
    +           skeleton-pair)
    +      (skeleton-pair-insert-maybe nil)
         ;; else
         (let (insertpos)
           (if (and (not arg)

However, this function could do a lot more for the user. Either with
skeleton pairing or without, the added braces do not position well for
quick typing. The variable `ess-auto-newline`, which apparently is only
used in this function, is used quite inflexibly -- it can only insert
newlines both before *and* after braces, which does not fit well with
many coding styles in active use.

With a minor change, we get additional functionality and flexibility:
automatic positioning of the braces consistent with a range of desired
styles. Here is a version of `ess-electric-brace` that I quickly wrote
up that has been working for me, with a corresponding new defcustom 
and docstring for `ess-auto-newline`:

    (defcustom ess-auto-newline nil
      "Alist controlling automatic newline insertion around braces in S code.
    
    This specifically controls the behavior of the function
    `ess-electric-brace'. If the alist is non-nil, its keys should be
    characters, typically braces, representing key events, and the
    values should be lists containing either or both of the symbols
    `before' or `after'. A `before' or `after' in the list for a key
    indicates that a newline (and reindentation) is inserted before
    or after the character, respectively.
    
    For example, setting this to
    
       ((?{ after) (?} before))
    
    inserts newlines and auto-indents at the braces to give the style
    
       if ( test ) {
           truebody
       }

    whereas setting it to 

       ((?{ before after) (?} before after))

    gives

       if ( test ) 
       {
           truebody
       }
 
    another common style.

    If this alist is nil, no newline insertion or re-indentation is
    done. If in addition the variable `skeleton-pair' is t, then the
    character (e.g., left brace) and its match (e.g., right brace)
    are inserted as a pair."
      :type '(alist :key-type character
                    :value-type (repeat (choice (const before)
                                                (const after))))
      :group 'ess-edit)
    
    (defun ess-electric-brace (arg)
      "Insert character and correct line's indentation."
      (interactive "P")
      ;; skeleton-pair takes precedence unless auto newlines configured
      (if (and (null ess-auto-newline)
               (fboundp 'skeleton-pair-insert-maybe)
               skeleton-pair)
          (skeleton-pair-insert-maybe nil)
        ;; else
        (let* (insertpos
               (brace (if (featurep 'xemacs)
                          (event-to-character last-command-event)
                        last-command-event))
               (auto-newline (cdr (assoc brace ess-auto-newline))))
          (when (and (not arg)
                     (eolp)
                     (or (save-excursion
                           (skip-chars-backward " \t")
                           (bolp))
                         auto-newline))
              (progn
                (when (memq 'before auto-newline)
                  (ess-indent-line)
                  (newline))
                (insert brace)
                (ess-indent-line)
                (when (memq 'after auto-newline)
                    (newline)
                    ;; (newline) may have done auto-fill
                    (setq insertpos (- (point) 2))
                    (ess-indent-line))
                (save-excursion
                  (if insertpos (goto-char (1+ insertpos)))
                  (delete-char -1))))
          (if insertpos
              (save-excursion
                (goto-char insertpos)
                (self-insert-command (prefix-numeric-value arg)))
            (self-insert-command (prefix-numeric-value arg))))))
        
This allows finer control of newline insertion and indentation in a way
that can match common styles. It allows `ess-auto-newline` to
suppress skeleton's precedence instead of relying only on the general
`skeleton-pair`. It is simple to set and can be set in the ess-style
structure.

This is actually a bit conservative for my taste. I've been playing with
a version that allows configurable actions (e.g., insert a space before
left-brace if no space is there and not at bol) and that configurably
pairs with indentation. But I kept this (posted) version close to the current 
version intentionally.

Comments welcome. I'm happy to submit a PR or a patch with this if it
would help. I've attached the patch and a file with the code to this
issue.

Thanks, Chris

[ess-electric-brace-patch.txt](https://github.com/emacs-ess/ESS/files/662820/ess-electric-brace-patch.txt)
[new-ess-electric-brace-el.txt](https://github.com/emacs-ess/ESS/files/662824/new-ess-electric-brace-el.txt)

jabranham:
(fboundp 'skeleton-pair-insert-maybe) returns t starting with Emacs
... 22? 23? It's unclear. In any case, in all recent Emacs versions,
it returns t, which means this whole function just
called (skeleton-pair-insert-maybe "{"), which in turn just calls
self-insert-command so long as skeleton-pair-alist is nil.

Closes #389


jabranham:
Alternatively, if we think a lot of people have `{` as a skeleton (I don't, but maybe?) we could keep it and check whether `skeleton-pair-alist` contains `{`. If so, call skeleton-pair-maybe-insert, otherwise call self-insert-command.

jabranham:
It's not really being used right like that. The argument is supposed to be a number, not a character: `Insert the character you type ARG times.`.

jabranham:
OK, done

lisp/ess-mode.el
```diff
@@ -100,9 +100,9 @@
     (define-key map "\C-c\C-k"   'ess-force-buffer-current)
     (define-key map "\C-c`"      'ess-show-traceback)
     (define-key map [(control ?c) ?~] 'ess-show-call-stack)
-    (define-key map "{"          'ess-electric-brace)
-    (define-key map "}"          'ess-electric-brace)
     (define-key map "\C-\M-q"    'ess-indent-exp)
+    (define-key map "{"          #'skeleton-pair-insert-maybe)
```
vspinu: Any special reason to quote them as functions (#')?

jabranham: Habit, more than anything else. I don't think it'll make a difference (except this way the byte compiler can warn if they aren't defined). I know `"'` looks weird to some people though. Do you want me to remove the sharp quote?

vspinu: I do use #' when it does matter, for keymaps I think everyone use simple quite. My concern is that it should be either #' or ' everywhere in keymaps. Otherwise it's ugly and people raise questions. I think It won't work for nested keymaps, those are not functions. So let's stick to ' throughout. Did you introduced a couple of these in the earlier PRs?

jabranham: OK, switched it back to `'`. I don't *think* I did this elsewhere?

vspinu:
:+1:

PR #575, opened 2018-06-13
Remove old emacsen support

jabranham:
This PR removes a whole bunch of compatibility code for Emacs < 24.3 and XEmacs. It also removes ess-compat.el, which held nothing after this cleanup.


vspinu:
Nice!

Issue #409, opened 2017-03-07
ess-eval-visibly nil or 'nowait has extra `+` in output


jabranham:
Evaluating this code: 

```R
library(dplyr)
mtcars %>%
  group_by(cyl) %>%
  summarize(m = mean(wt),
            n = n())
```

with `(setq ess-eval-visibly t)` (the default) produces this in the iESS buffer: 

```
# A tibble: 3 × 3
    cyl        m     n
  <dbl>    <dbl> <int>
1     4 2.285727    11
2     6 3.117143     7
3     8 3.999214    14
```

While setting `ess-eval-visibly` to `'nowait` or `'nil` produces this:

```
> + + + # A tibble: 3 × 3
    cyl        m     n
  <dbl>    <dbl> <int>
1     4 2.285727    11
2     6 3.117143     7
3     8 3.999214    14
```

Including the `>` and `+` symbols can make the output more confusing to read when it's more complicated since alignment gets messed up.

I'm running Emacs 25.1, ess-version 16.10-1 [elpa: 20170227.802] on a linux system.

vspinu:
There is already a mechanism that tries to insert new lines after +++, but it fails on some occasions, yours being one of them. I am reasonably sure that that mechanism could be improved.

jabranham:
Thanks. If you can point me to that, I can take a look to see if I can improve it. I'm having problems finding that on my own.

PR #576, opened 2018-06-14
Strip extra + signs from output depending on ess-eval-visibly

jabranham:
```
When ess-eval-visibly is nil or 'nowait, the inferior process outputs
+ as a part of the output. We don't want these, though, because they
are already inserted when putting the input in. They also throw off
alignment of the first column of the output, which can be annoying.
```
Closes bug #409

### DIFF
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -1211,8 +1211,14 @@ value from EXPR and then sent to the subprocess."
    (ess-eval-visibly prompt)
    ((eq inferior-ess-replace-long+ 'strip) "> ")
    ((eq inferior-ess-replace-long+ t)
-    (replace-regexp-in-string "\\(\\+ \\)\\{4\\}\\(\\+ \\)+"
-                              ess-long+replacement prompt))
+    (let ((prompt (replace-regexp-in-string "\\(\\+ \\)\\{4\\}\\(\\+ \\)+"
+                                            ess-long+replacement prompt))
+          (len (length prompt)))
+      (if (and (> len 2)
+               (eq (elt prompt (- len 2)) ?+))
+          ;; append > for aesthetic reasons
+          (concat prompt "> ")
+        prompt)))
    (t prompt)))
 
 (defun ess--flush-accumulated-output (proc)
@@ -1280,9 +1286,6 @@ mirrors the output into *ess.dbg* buffer."
                           (string-match ess--dbg-regexp-skip string)
                           (not (string-match ess--dbg-regexp-no-skip string))))
          (match-dbg (or match-skip (and match-input (not match-selection))))
-         ;;check for main  prompt!! the process splits the output and match-end == nil might indicate this only
-         ;; (prompt-regexp "^>\\( [>+]\\)*\\( \\)$") ;; default prompt only
-         (prompt-replace-regexp "\\(^> \\|^\\([>+] \\)\\{2,\\}\\)\\(?1: \\)") ;; works only with the default prompt
          (is-ready (not (inferior-ess-set-status proc string)))
          (new-time (float-time))
          (last-time (process-get proc 'flush-time))
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -498,8 +498,8 @@ If 'strip, remove all such instances.  Otherwise, if non-nil, '+
   "Replacement used for long + prompt.
 Customization of this variable is not recommended. You can set it
 to '. '. If you set it to anything else you will have to change
-`inferior-S-prompt' to assure the correct prompt navigation
-in inferior buffers.  ")
+`inferior-S-prompt' to assure the correct prompt detection
+in inferior buffers.")
 
 (defmacro ess-copy-key (from-map to-map fun)
   `(define-key ,to-map
@@ -525,7 +525,7 @@ in inferior buffers.  ")
     (set (make-local-variable 'compilation-error-regexp-alist)
          ess-error-regexp-alist)
     (let (compilation-mode-font-lock-keywords)
-     (compilation-setup t))
+      (compilation-setup t))
     (setq next-error-function 'ess-tracebug-next-error-function)
     ;; new locals
     (make-local-variable 'ess--tb-last-input)
@@ -1150,7 +1150,7 @@ Kill the *ess.dbg.[R_name]* buffer."
 (defvar ess--dbg-regexp-debug  "\\(\\(?:Browse[][0-9]+\\)\\|\\(?:debug: \\)\\)")
 (defvar ess--dbg-regexp-selection "\\(Selection: \\'\\)")
 (defvar ess--dbg-regexp-input (concat ess--dbg-regexp-debug "\\|"
-                                     ess--dbg-regexp-selection))
+                                      ess--dbg-regexp-selection))
 
 (defvar ess--suppress-next-output? nil)
 
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -479,11 +479,6 @@ can use `ess--busy-slash', `ess--busy-B',`ess--busy-stars',
 (defvar ess--busy-timer nil
   "Timer used for busy process indication")
 
-(defcustom inferior-ess-split-long-prompt t
-  "If non-nil, long prompt '> > > > > + + + + > ' is split."
-  :group 'ess-tracebug
-  :type 'boolean)
-
 (defcustom inferior-ess-replace-long+ t
   "Determines if ESS replaces long + sequences in output.
 If 'strip, remove all such instances.  Otherwise, if non-nil, '+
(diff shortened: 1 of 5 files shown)


jabranham:
It also fixes stuff like when you evaluate a buffer, you might end up with:

```

+ . + > 
+ . + + . + + . + + . + > 
```

before this patch and only

```
> 
> 
```

after this patch

lisp/ess-tracebug.el
```diff
@@ -1262,6 +1262,10 @@ If in debugging state, mirrors the output into *ess.dbg* buffer."
       ;; FIXME: this should be in comint filters!!
       ;; insert \n after the prompt when necessary
       (setq string (replace-regexp-in-string prompt-replace-regexp " \n" string nil nil 1))
+      ;; Replace + that gets printed twice in the output:
+      (when (or (not ess-eval-visibly)
+                (eq ess-eval-visibly 'nowait))
```
vspinu: this should be intergrated within the following `(when infrior-ess-replace-long+ ...`.

I think a better idea would be to multiple options for `inferior-ess-replace-long+` (nil, t, strip).

vspinu:
I think we want them actually. Length of `+ . +` is proportional to the size of the code evaluated. Otherwise it would have been removed long time ago.

lisp/ess-tracebug.el
```diff
@@ -1262,6 +1262,10 @@ If in debugging state, mirrors the output into *ess.dbg* buffer."
...
+        (setq string (replace-regexp-in-string "\\+[[:space:]]" "" string)))
```
vspinu: This regexp cannot be. It will strip "1 + 3".

jabranham:
OK, in its new incarnation it depends on the value `inferior-ess-replace-long+`. It can now take the value `'strip`, which means that all the `+`'s will get stripped from output. I updated the regex so that it's smarted and doesn't strip the `+` from e.g. `"1 + 3"`.

lisp/ess-tracebug.el
```diff
@@ -1264,8 +1269,14 @@ If in debugging state, mirrors the output into *ess.dbg* buffer."
       (setq string (replace-regexp-in-string prompt-replace-regexp " \n" string nil nil 1))
 
       ;; replace long prompts
-      (when inferior-ess-replace-long+
-        (setq string (replace-regexp-in-string "\\(\\+ \\)\\{4\\}\\(\\+ \\)+" ess-long+replacement string)))
+      (cond ((and (eq inferior-ess-replace-long+ 'strip)
+                  (or (not ess-eval-visibly)
```
vspinu: Wouldn't this `or` condition also make sense for the next branch? In other words can we have 

```el
(when (and inferior-ess-replace-long+
                   (or (not ess-eval-visibly)..))
   (if ...))
```

lisp/ess-tracebug.el
```diff
@@ -1264,8 +1269,14 @@ If in debugging state, mirrors the output into *ess.dbg* buffer."
...
+                      (eq ess-eval-visibly 'nowait)))
+             (setq string (replace-regexp-in-string "\\(\\+[[:space:]]\\).*\\'"
+                                                    "" string)))
```
vspinu: Won't work either ;).  `+ + +` need not occur at the end of the process filter `string`. I think the best you can do here is to match `+ ` at least twice. 

Also could you please use a plain space instead of `[[:space:]]`. We know for sure that it's a space there, it's more readable and it's likely to be slightly more efficient.

lisp/ess-tracebug.el
jabranham: Sure, I'll add it there

lisp/ess-tracebug.el
jabranham: The newest version (just pushed) seems to work for me, though it does occasionally put the output on the line containing the prompt instead of the line after that. Can you give an example of where it fails?

vspinu: For example if the string is `+ + + + > some-output >`  you will strip all of the output. It will also strip prompt in usual `+ + + >` in which case there will be no indication of evaluation in the REPL. Some emacs ides do that but I don't like it personally.   

I would rather use `\\(\\+ \\)\\{2,\\}` there. In any case you will still have the "offset" problem. 

The solution is not as simple as it looks. I don't think it could be solved with reg-exp in process filter only. One needs to look at what follows after + + + and insert an "\n". It also should be efficient. The best way I can think of is to search for + + + just before the accumulated output is inserted into the repl (see accumulation code nearby) .

vspinu: To clarify it further, the regexp handles the limiting case when "+ + >" arrives in one chunk and the output in another. That is probably happening in your tests but doesn't hold in general. 

I will pull your branch over the weekend and integrate the fix on the accumulated output which I described above. 

Besides the mtcars example do you have other examples with shifted output which I can try out?

mmaechler:
So will the remaining output by default remain proportional to the amount of code sent?

As @vspinu I have been "using" that to check if (large or small) chunks of code had been sent..

vspinu:
Yes. That should stay. Even though it's quite common in emacs based IDEs not to reflect evaluation in the REPL but show only a message in the minibuffer. But as long as we output at least one ">" on each evaluation, it's uniformly better to display the full "+ + . + >".

vspinu:
@jabranham I have pushed some stuff to your branch. It's quite a rewrite of the output accumulation and flushing mechanism, so I would apreciate if you could test it before the merge.

jabranham:
@vspinu Thanks!

When I execute

```
library(dplyr)
mtcars %>%
  group_by(cyl) %>%
  summarize(m = mean(wt),
            n = n())
```

I get

```
> + + + 
# A tibble: 3 x 3
    cyl     m     n
  <dbl> <dbl> <int>
1     4  2.29    11
2     6  3.12     7
3     8  4.00    14
```

This is better than the current situation since it doesn't mess up alignment, but it would be nice if the extra `+` symbols weren't displayed.

vspinu:
That you can do by setting `inferior-ess-replace-long+` to 'strip.

jabranham:
I still get those `+` sequences after the prompt even after setting that to `'strip` though

vspinu:
Not on your example, right? It might be that I overlooked something. Would be nice if you could isolate it.

jabranham:
Yes, on that example. Do you not see the `> + + +` ?

I haven't had time yet to look into what's happening but hopefully can later today

vspinu:
Actually,  this is one example:

```R
library(dplyr)
mtcars %>%
  group_by(cyl) %>%
  summarize(m = mean(wt),
            n = n())
library(dplyr)
mtcars %>%
  group_by(cyl) %>%
  summarize(m = mean(wt),
            n = n())
```
I am pretty sure it's because the prompt is split between two chunks, but I am not sure that we can deal with that satisfactory.

vspinu:
I think stripping is not a great idea actually, at least not with current implementation. How do you see incomplete input, when process waits for input?

lionel-:
FWIW I think the main inconvenience is to have these `+ + +` mess up output alignment. If we could find a way of adding a newline at the right place (just before output), we wouldn't really need to strip.

jabranham:
> I think stripping is not a great idea actually

I'm starting to agree with you the more I think about it, especially now that we've fixed the alignment issue.

vspinu:
> FWIW I think the main inconvenience is to have these + + + mess up output alignment. If

This is exactly what this PR is fixing. 

> I am pretty sure it's because the prompt is split between two chunks,

Turned out not to be the case, but a logical bug.

> I think stripping is not a great idea actually

I am looking into how to (almost) fix that at the moment. I think users should be fine with rare false-negatives when the process output chunks are split exactly on the prompt.

vspinu:
This is what I get with the latest fixes:

```R

(1 +
   3)
library(dplyr)
mtcars %>%
  group_by(cyl) %>%
  summarize(m = mean(wt),
            n = n())
library(dplyr)
mtcars %>%
  group_by(cyl) %>%
  summarize(m = mean(wt),
            n = n())
"+ + + + > some-output >"
writeLines("+ + + + > some-output >")
```
With the default `t`:

```R
> + > 
[1] 4
> > . + > 
# A tibble: 3 x 3
    cyl     m     n
  <dbl> <dbl> <int>
1    4.  2.29    11
2    6.  3.12     7
3    8.  4.00    14
> > . + > 
# A tibble: 3 x 3
    cyl     m     n
  <dbl> <dbl> <int>
1    4.  2.29    11
2    6.  3.12     7
3    8.  4.00    14
> 
[1] "+ + + + > some-output >"
> . + > 
some-output >
> 
```

And with 'strip:

```R
> > 
[1] 4
+ > 
# A tibble: 3 x 3
    cyl     m     n
  <dbl> <dbl> <int>
1    4.  2.29    11
2    6.  3.12     7
3    8.  4.00    14
+ > 
# A tibble: 3 x 3
    cyl     m     n
  <dbl> <dbl> <int>
1    4.  2.29    11
2    6.  3.12     7
3    8.  4.00    14
> 
[1] "+ + + + > some-output >"
> 
some-output >
> 
```

The difference is minimal; the 'strip case being only slightly cleaner. The +ses in the 'strip case are unavoidable unless one were to manipulate comint buffer directly. Thing which I would like to avoid due to extra complexities and loss of efficiency.

PR #578, opened 2018-06-17
Improve documentation for new smart S assignment change.

MilesMcBain:
I had to go to the PR for this change to find the new function I needed to call. I think it would be kind to reference this in the documentation. Calling `ess-toggle-underscore` after ESS is loaded seems to be a fairly common practice - indeed it is done in the Spacemacs ESS layer.


lionel-:
Thanks!

PR #580, opened 2018-06-17
Drop official support for Emacs 24.3

lionel-:
Includes and closes #561.

I stumbled onto issues with Emacs 24.3 while improving the test framework and spent too much time trying to fix these. As 24.3 was released in March 2013, more than 5 years ago, it is likely not worth it to bother maintaining support for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: emacs-lisp
 env:
-  - EVM_EMACS=emacs-24.3-travis
+  - EVM_EMACS=emacs-24.4-travis
   - EVM_EMACS=emacs-25.3-travis
   - EVM_EMACS=emacs-26.1-travis
   - EVM_EMACS=emacs-git-snapshot-travis
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,7 +4,10 @@
 Changes and New Features in development version:
 @itemize @bullet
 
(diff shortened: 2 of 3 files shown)


mmaechler:
No!  I think we must have another release with 24.3 support.   
If it is only because some very long time users work in environments where they don't want or more importantly cannot update their system easily.  Think of companies with super stable environments (from their own internal guidelines). Believe me: they do exist,  e.g.,  in pharma and insurance [yes, both important here in Switzerland].   Yes, it's hard do believe particularly if you'r a young Ph.d working in a thriving "edge-type" environment.    Even more reason to aim for a  18.07 or 18.08 release -- the last with Emacs 24.3 support

lionel-:
I don't doubt these environments exist. I just think they can keep using an old ESS with their old Emacs.

Also note: this is official support we are removing, it only means we stop testing systematically for 24.3 compatibility, it does not mean ESS stops working on this version. If we are having a release soon, there's virtually no regression for Emacs 24.3 except it won't be able to run our suite of unit tests (the incompatiblity that posed me issues is in the condition signalling/catching for ERT).

lionel-:
No need to have a discussion about this, I'll just separate the test skipping from the multiple cursors. I only need the latter at the moment.

> the last with Emacs 24.3 support

I think the plan is to drop 24.4 as well in the next version, hence dropping 24.3 now would have made sense.

PR #582, opened 2018-06-17
Add generator for keyword regexps

lionel-:
It is currently quite tricky to modify variables like `ess-R-keywords`. If you modify those

* after loading ESS, this has no effect because the font-lock regexps are not updated;
* before loading ESS, you can't append to these variables as they are not defined yet

With `ess-r-generate-font-lock-regexps` implemented in this PR, users can update the keyword variables and regenerate the regexps manually afterwards.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2647,7 +2647,9 @@ This variable has no effect. Customize
 
 (defvar ess-R-keywords
   '("if" "else" "repeat" "while" "function" "for" "in" "next" "break")
-  "Reserved words in the R language.")
+  "Reserved words in the R language.
+If you modify this variable, call `ess-r-generate-font-lock-regexps'
+for the changes to take effect.")
 
 (defvar ess-R-control-flow-keywords
   '("switch" "function" "return" "on.exit" "stop"
@@ -2655,12 +2657,18 @@ This variable has no effect. Customize
     "recover" "browser")
   "Keywords that impact control flow.
 These keywords either cause a control flow jump or establish a
-jump target.")
+jump target.
+
+If you modify this variable, call `ess-r-generate-font-lock-regexps'
+for the changes to take effect.")
 
 (defvar ess-R-signal-keywords
   '("message" "warning" "signalCondition" "withCallingHandlers")
   "Keywords for condition signalling.
-These keywords might cause a control flow jump but do not necessarily.")
+These keywords might cause a control flow jump but do not necessarily.
+
+If you modify this variable, call `ess-r-generate-font-lock-regexps'
+for the changes to take effect.")
 
 (defvar ess-S-keywords
   (append ess-R-keywords '("terminate")))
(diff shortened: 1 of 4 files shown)


lisp/ess-custom.el
```diff
@@ -2782,21 +2790,52 @@ default or not."
...
+  "Font-lock regexp for control flow keywords.")
 
-(defvar ess-R-fl-keyword:fun-defs
-  (cons ess-R-function-name-regexp
-        '(1 font-lock-function-name-face nil))
-  "Font-lock keyword - function defintions for R.")
+(defvar ess-R-fl-keyword:signal-keywords nil
+  "Font-lock regexp for signal keywords.")
+
+(defvar ess-R-fl-keyword:assign-ops nil
```
vspinu: I really dislike that all these variables are initialized with `nil` and then set by side effect elsewhere. Cannot we do better here?

lionel-: How would you do it without code duplication? A defvar is side-effectful as well. Or are you concerned about the semantics of initialisation?

lionel-: If the latter, we could have an `&optional init` argument to signal we shouldn't override variables that are already set.

lisp/ess-custom.el
```diff
@@ -2647,20 +2647,28 @@ This variable has no effect. Customize
...
+
+If you modify this variable, call `ess-r-generate-font-lock-regexps'
+for the changes to take effect.")
 
 (defvar ess-R-signal-keywords
   '("message" "warning" "signalCondition" "withCallingHandlers")
   "Keywords for condition signalling.
-These keywords might cause a control flow jump but do not necessarily.")
+These keywords might cause a control flow jump but do not necessarily.
+
```
vspinu: A side note. let's try to avoid fragmentation of doc strings. For long docs it might be warranted but not here. Blanks eat vertical space, interfere with structural navigation and fragment the code. Plus it's not the best practice to create one-sentence-paragraphs. Lisp is beautiful because it's compact, and it's a pity if rarely read docs would spoil that.

lisp/ess-custom.el
vspinu: I was concerned mostly with the stylistic part of it. When I navigate to the var definition and I see nil, it's disturbing. I have to go and figure where and how it is initialized.

The initialization issue is even more imporant. If people "customized" those variables we will be overwriting them.

vspinu: Is the sole motivation to reduce the burden of restarting emacs after setting those vars? If so, I don't see the encured complications worth the tradeoff.

vspinu: Plus add to this that we are introducing one extra function and making both users and developer read and understand it.  With all those questions naturally raised "Why? How? Why not standard?". It's complicated. We are in cutting and cleaning season,  not adding extra complexities.

There are probably other variables in ESS which cannot be easily re-initialized.  I don't think we should introduce an init function for each of those.

lionel-: > Is the sole motivation to reduce the burden of restarting emacs after setting those vars? If so, I don't see the encured complications worth the tradeoff.

No, it's to make it possible to _modify_ the default list of keywords without having to copy the default list (and freeze it for good, missing out on upstream changes) in the user config.

There is currently no way to modify these. This function makes things simpler, not more complicated.

lisp/ess-custom.el
lionel-: Also I think the general solution for navigability of elisp code is to have the docstring folded by default.

lisp/ess-custom.el
lionel-: I don't think `font-lock-remove-keywords` is practical. IIUC you have to provide the exact keyword spec right? But this spec has been constructed by ESS and is not trivially reproducible. So while it's easy to add new keywords, it's hard to remove them. Letting users customise `-keywords` variables is a good feature from that point of view.

vspinu: > Also I bet we leave many users perplexed as to why their modification of the keywords variables (after having loaded ESS) do not have any effect. 

That's pretty standard and I thinks users are used to that. 

> missing out on upstream changes

Well, that is the case with all customized defcustoms and all `setq`ed variables. I wouldn't put too much weight on that. Your proposal doesn't really help with that. If a user already have them customized, our new changes won't affect them.

Actually, in the next version, we can issue a warning if we detect that someone had those keywords set. We might also go for some dynamic version. If some packages do provide special set of keywords, we can pick them up automatically. CIDER has such a feature for macros. Or maybe a special set `ess-user-custom-keywords` is warranted here.

I do agree with @jabranham here. User's shouldn't ideally be customizing those variables, and if they do it's sort of their problem to solve. Plus, they should use `font-lock-add-keywords` for new stuff. Why users would need to remove keywords?

lisp/ess-custom.el
vspinu: I don't know why people think that having a sentence in a separate paragraph improves readability. In any case, this one is not a defcustom, but an internal var. We should strive to keep internal vars and functions documented to the minimum. As a developer you read the code 100 times and the docs once.

> Also I think the general solution for navigability of elisp code is to have the docstring folded by default.

True. Do you happen to have a good setup for this?

lisp/ess-custom.el
lionel-: > That's pretty standard and I thinks users are used to that.

They may be used to being perplexed and frustrated (I know I was, when learning ESS) but that's not a good state of affairs. We should make it easy to configure ESS and clearly document things that are not meant to be configured and the alternative process. `ess-R-keywords`-like variables are currently deceptive.

> Why would users remove keywords?

Because they don't like them? The main use case is to add to the keywords though, without having to worry about regexps (matching the open paren etc).

If the only thing supported user interaction for those keywords is to set them to nil, this should be documented. And we should move all these variables to `ess-r--` if they are internal.

Overall I don't see it a big deal to have the keywords variables as part of the configuration UI. And it is standard to have a follow-up function to execute after setting a variable (which is automatically run by the customisation system when the defvar is promoted to defcustom).

In any case we should be clear about what's part of the UI and what's private.

lionel-: > that is the case with all customized defcustoms and all setqed variables.

That is not true, you can usually modify variables and customs.

vspinu:
To make a long story short, there is a good chance that with a a bit of thinking we can have simpler, more elegant and a more general solution here. For example, we could re-initialize all font-lock keywords for each file from scratch by making the internal `ess-r-keyword:xyz`  lisp forms or functions to be evaluated on ess-mode initialization.

lisp/ess-custom.el
vspinu: >>    that is the case with all customized defcustoms and all setqed variables.

> That is not true, you can usually modify variables and customs.

Don't know what is so much "not true" here. If a user customizes a value your change upstream will not be propagated into the user config. It's a null argument.

lionel-: You can usually both read from and write to a variable, which is currently not the case with the keywords variables. What is a null argument?

vspinu: We are talking about different things. I was referring to your comment that they will miss out on upstream change - "having to copy the default list (and freeze it for good, missing out on upstream changes) in the user config." . You focus on how within session changes propagate into the application.  Anyways, let's move into the main thread and let's think how we can make changes propagate without adding any new  "generators" and code duplication.

lionel-: There is currently no way to read from the default keyword lists and write a new list. You can either read from it after loading ESS, or write to it before loading ESS. Doing both is not possible, not within session, and not even after restarting Emacs. Hence there is no way to pick up the default upstream list and append to it, which I don't think is an unusual case (for example there are packages out there that provide additional control-flow functions).

I agree it is preferable for the changes to be automatically picked up, I'll think about your suggestion of introducing laziness, thanks.

vspinu: Ok. I see. "upstream" usually means the code in the main repo out there. So I thought you differentiate between "upstream" (our settings) and user settings. 

> Hence there is no way to pick up 

Yes. I think we all agree this is a problem. We disagree how to deal with it. 

It makes me wonder if we have similar "hardwiring" elsewhere in ESS. I think we do, but I am too busy right now to dwell and look for those. In any case the font-lock is probably the most important case.

PR #585, opened 2018-06-21
Rework ess-insert-S-assign, ess-smart-S-assign, ess-cycle-assignment

### LINKED ISSUE #584: new smart assign behaves weird
I have set `ess-smart-S-assign-key` to `"\M--"`, now `M--` inserts `<-`,  but if I press `M--` two times, `<-` changes to `\255`. Can this be reproduced by anyone?

jabranham:
* ess-insert-S-assign now inserts ess-S-assign, and only replaces it
  if new defcustom ess-maybe-replace-S-assign is non-nil. Since
  ess-maybe-replace-S-assign is nil by default, this changes current
  behavior. It is removed from ess-mode-map and inferior-ess-mode-map.

* New function ess-cycle-assignment cycles through the assignment
  operators in new defcustom ess-assign-list. It is bound to C-c C-=
  in ess-mode-map and inferior-ess-mode-map

Closes #584

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -107,15 +107,21 @@ will need to update their customization to
 @code{inferior-ess-R-program}. These variables are treated as risky
 variables.
 
-@item Customization of ess-smart-S-assign-key has been reworked.
-Use @code{(setq ess-smart-S-assign-key nil)} to disable smart
-assignment at any time instead of @code{(ess-toggle-underscore nil)}.
-To use another key, you should set the value of
-@code{ess-smart-S-assign-key} before ESS is loaded. The following
-functions have been made obsolete. You should customize
-ess-smart-S-assign-key instead: ess-toggle-S-assign,
-ess-toggle-S-assign-key, ess--unset-smart-S-assign-key,
-ess--activate-smart-S-assign-key, ess-disable-smart-S-assign
+@item @code{ess-smart-S-assign} changed name to @code{ess-insert-S-assign}.
+By default, it provides similar functionality. Since many users
+dislike this functionality, it is easier to disable; set
+@code{ess-smart-S-assign-key} to nil. The following functions have
+been made obsolete and will be removed in the next release of ESS:
+ess-smart-S-assign, ess-toggle-S-assign, ess-toggle-S-assign-key,
+ess-disable-smart-S-assign.
+
+@item The option ess-S-assign has been removed.
+Customize new option @code{ess-assign-list} instead.
+
+@item @code{C-c C-=} is now bound to @code{ess-cycle-assignment} by default
+See the docstring for details. New user customization option
+@code{ess-assign-list} controls what assignment operators are cycled
+through.
 
 @end itemize
 
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -549,12 +549,6 @@ ess-smart-comma only, but will be enriched in the near future.")
 Used to avoid annoying jumping by ess-eval.*-and-step to end of
 buffer or end chunks etc.")
 
-(defcustom ess-S-assign " <- "
-  "String used for left assignment in all S dialects.
-Used by \\[ess-smart-S-assign]."
-  :group 'ess-S
-  :type 'string)
-
 (defcustom ess-smart-S-assign-key "_"
   "Key used by `ess-smart-S-assign'.
 Should be nil or a \"simple\" key, in other words no key
@@ -564,6 +558,19 @@ You may change this to nil at any time. However, if you change it
 to another string, it must be set before ESS is loaded."
   :group 'ess-S
   :type '(choice (const :tag "Nothing" :value nil) string))
+
+(defcustom ess-assign-list (cons (if (boundp 'ess-S-assign) ess-S-assign " <- ")
+                                 '(" <<- " " = " " -> " " ->> "))
+  "List of assignment operators.
+`ess-cycle-assignment' uses this list.  These strings must
+contain spaces on either side."
+  ;; Note that spaces on either side is not strictly true (as in the
+  ;; function won't error), but matching <-/<<- is broken without
+  ;; them.
+  :type '(repeat string)
+  :group 'ess)
+(defvar ess-S-assign)
+(make-obsolete-variable 'ess-S-assign 'ess-assign-list "2018-07-01")
 
 (defcustom ess-r-prettify-symbols
   '(("<-" . (?\s (Br . Bl) ?\s (Bc . Bc) ?←))
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1952,12 +1952,12 @@ for `ess-eval-region'."
     (define-key map "\M-?"     'ess-list-object-completions)
     (define-key map "\C-c\C-k" 'ess-request-a-process)
     (define-key map ","        'ess-smart-comma)
-
+    (define-key map (kbd "C-c C-=") 'ess-cycle-assignment)
     (define-key map "\C-c\C-d"   'ess-doc-map)
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
     (when ess-smart-S-assign-key
-      (define-key map ess-smart-S-assign-key #'ess-smart-S-assign))
+      (define-key map ess-smart-S-assign-key 'ess-insert-assign))
     map)
   "Keymap for `inferior-ess' mode.")
 
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -107,7 +107,8 @@
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
     (when ess-smart-S-assign-key
-      (define-key map ess-smart-S-assign-key #'ess-smart-S-assign))
+      (define-key map ess-smart-S-assign-key 'ess-insert-assign))
+    (define-key map (kbd "C-c C-=") 'ess-cycle-assignment)
     map)
   "Keymap for `ess-mode'.")
 
--- a/lisp/ess-s-lang.el
+++ b/lisp/ess-s-lang.el
@@ -575,67 +575,51 @@ and one that is well formatted in emacs ess-mode."
     (ess-if-verbose-write "ess-fix-misc __end__\n");___D___
     ))
 
-(defvar polymode-mode)
-(defun ess-smart-S-assign (&optional N)
-  "Act as smart `ess-S-assign' key.
-Insert `ess-S-assign', unless in string/comment.  If the
-underscore key is pressed a second time, the assignment operator
-is removed and replaced by the underscore.  `ess-S-assign',
-typically \" <- \", can be customized.  In ESS modes other than
-R/S, the underscore is always inserted.
-
-If `ess-smart-S-assign-key' is nil, just call
-`self-insert-command'. You can pass N as usual."
+(defun ess-cycle-assignment ()
+  "Cycle between assignment symbols in `ess-assign-list'.
+On consecutive calls, replace the assignment symbol before point
+with the next symbol from that list. This function sets the last
+keypress to repeat it, so if it is bound to \"C-c C-=\" pressing
+\"=\" again cycles to the next assignment."
   (interactive)
+  (if (eq last-command this-command)
+      (let ((slist ess-assign-list)
+            str)
+        (while (and (setq str (car slist)
+                          slist (cdr slist))
+                    (not (and (re-search-backward str
+                                                  (- (point) (length str)) t)
+                              (not (replace-match (car slist))))))))
+    (insert (car ess-assign-list)))
+  (set-transient-map
+   (let ((map (make-sparse-keymap))
+         (key (format "%c" (event-basic-type last-input-event))))
+     (define-key map (kbd key) #'ess-cycle-assignment)
+     map)))
+
+(defun ess-insert-assign (arg)
+  "Insert the first element of `ess-assign-list' unless in string or comment.
+If the character before point is the first element of
+`ess-assign-list', replace it with the last character typed. If
+`ess-smart-S-assign-key' is nil, do `self-insert-command' using
+ARG as the number of times to insert."
+  (interactive "p")
   (if ess-smart-S-assign-key
-      (save-restriction
-        (ignore-errors
-          (when (and (eq major-mode 'inferior-ess-mode)
-                     (> (point) (process-mark (get-buffer-process (current-buffer)))))
-            (narrow-to-region (process-mark (ess-get-process)) (point-max)))
-          (and ess-noweb-mode
-               (ess-noweb-in-code-chunk)
-               (ess-noweb-narrow-to-chunk))
-          (and (fboundp 'pm/narrow-to-span)
-               polymode-mode
-               (pm/narrow-to-span)))
-        (if (or
-             (ess-inside-string-or-comment-p (point))
-             (not (equal ess-language "S")))
-            (insert ess-smart-S-assign-key)
-          (ess-insert-S-assign)))
-    (funcall #'self-insert-command (or N 1))))
+      (let* ((assign (car ess-assign-list))
+             (event (event-basic-type last-input-event))
+             (char (ignore-errors (format "%c" event))))
+        (cond ((and char (ess-inside-string-or-comment-p))
+               (insert char))
+              ((re-search-backward assign (- (point) (length assign)) t)
+               (if (and char (numberp event))
+                   (replace-match char)
+                 (replace-match "")))
+              (t (insert assign))))
+    (funcall #'self-insert-command arg)))
 
-(defun ess-insert-S-assign ()
-  "Insert the assignment operator `ess-S-assign', unless it is already there.
-In that case, it is removed and replaced by `ess-smart-S-assign-key'.
-  `ess-S-assign', typically \" <- \", can be customized."
-  (interactive)
-  ;; one keypress produces ess-S-assign; a second keypress will delete
-  ;; ess-S-assign and instead insert _
-  ;; Rather than trying to count a second _ keypress, just check whether
-  ;; the current point is preceded by ess-S-assign.
-  (let ((assign-len (length ess-S-assign)))
-    (if (and
-         (>= (point) (+ assign-len (point-min))) ;check that we can move back
-         (save-excursion
-           (backward-char assign-len)
-           (looking-at ess-S-assign)))
-        ;; If we are currently looking at ess-S-assign, replace it with _
-        (progn
-          (delete-char (- assign-len))
-          (insert ess-smart-S-assign-key))
-      (if (string= ess-smart-S-assign-key "_")
-          (delete-horizontal-space))
-      (insert ess-S-assign))))
-
-(defalias 'ess--activate-smart-S-assign-key 'ignore "")
-(make-obsolete 'ess--activate-smart-S-assign-key
-               "it does nothing. Set `ess-smart-S-assign-key' instead." "2018-06-08")
 (defun ess-disable-smart-S-assign (&rest _ignore)
-  "Disable `ess-smart-S-assign'."
-  (declare (obsolete ess-smart-S-assign-key "2018-06-08"))
-  (warn "This function is obsolete, use `ess-smart-S-assign-key' instead.")
+  "Disable `ess-insert-assign'."
+  (declare (obsolete "Use ess-smart-S-assign-key instead." "2018-06-08"))
   (setq ess-smart-S-assign-key nil))
 
 (defun ess-add-MM-keys ()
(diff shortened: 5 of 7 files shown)


lisp/ess-custom.el
```diff
@@ -551,19 +551,25 @@ buffer or end chunks etc.")
...
   :group 'ess-S
   :type 'string)
 
-(defcustom ess-smart-S-assign-key "_"
-  "Key used by `ess-smart-S-assign'.
-Should be nil or a \"simple\" key, in other words no key
-modifiers.
+(defcustom ess-assign-list '(" <- " " <<- " " = " " -> " " ->> ")
+  "List of assignment operators.
+`ess-cycle-assignment' uses this list.  These strings must
```
lionel-: Must they be surrounded by spaces? Can it be `'(" <- " "_")`?

jabranham: Kinda. If you have `<-` then `<<-`, the regex messes up. I figured that we should surround them with spaces anyway since `ess-S-assign` is also surrounded by spaces.

jabranham:
There's not much relationship between those two. `ess-cycle-assignment` is new. 

Previously, we had `ess-insert-S-assign` and `ess-smart-S-assign`. `ess-insert-S-assign` inserted `<-` or replaced it with a character (usually `_`). But it was bound to `_` by default, which many (most?) people don't like, since you get ` <- ` when typing `_`.  `ess-smart-S-assign` did mostly the same thing. 

If we pull these changes, `ess-insert-S-assign` now inserts ` <- ` by default and does not replace it (though it can if you customize that new defcustom). It's also not bound to `_`, so typing `_` gets you `_`. 

This meant that we didn't have an easy way to insert ` <- `, so new `ess-cycle-assignment` was created and bound to `C-c C-=` (I wanted to bind it to `M--`, but I think that's against Emacs conventions). Calling it multiple times will cycle through the various assignment operators.

lionel-:
Can we keep original behaviour of ess-insert-S-assign and make it cycle back to self-insert-command when the keybinding is a single letter? Then people can use it with `.` for instance (which would be much better than `"_"` I think, but that's just because I dislike dot.case ;)).

lionel-:
> I don't know if there's an easy/reasonable way to do that?

Maybe `last-input-event`?

> What would happen if someone bound ess-insert-S-assign to M--?

Cycling to nothing (i.e. delete ` <- `) seems reasonable.

lionel-:
Maybe insert-s-assign should use `(car ess-assign-list)`.

lisp/ess-custom.el
lionel-: Is regexp needed? Can we match strings literally instead?

jabranham:
So if `last-input-event` is an ASCII character, replace ` <- ` with it, otherwise delete ` <- `? Seems reasonable to me.

I think if `insert-S-assign` uses `(car ess-assign-list)` then `ess-S-assign` isn't used elsewhere, I could just remove it?

lionel-:
> I could just remove it

I think so, let's keep things as simple as possible.

@vspinu Would it make sense to standardise naming here, given all the changes?

jabranham:
OK, I've pushed a commit removing `ess-S-assign`. Checking whether `last-input-event` is an ascii character is more complicated than I thought it would be.

jabranham:
And now `ess-S-assign` should insert whatever ASCII character was typed if ess-maybe-replace-S-assign is non-nil

lisp/ess-s-lang.el
```diff
@@ -575,68 +575,49 @@ and one that is well formatted in emacs ess-mode."
...
-           (looking-at ess-S-assign)))
-        ;; If we are currently looking at ess-S-assign, replace it with _
-        (progn
-          (delete-char (- assign-len))
-          (insert ess-smart-S-assign-key))
-      (if (string= ess-smart-S-assign-key "_")
-          (delete-horizontal-space))
-      (insert ess-S-assign))))
+  (let ((assign (car ess-assign-list))
+        (event (and (numberp last-input-event)
```
vspinu: Use `simple-event-type` here.

lisp/ess-s-lang.el
```diff
@@ -575,68 +575,49 @@ and one that is well formatted in emacs ess-mode."
...
-          (and (fboundp 'pm/narrow-to-span)
-               polymode-mode
-               (pm/narrow-to-span)))
-        (if (or
-             (ess-inside-string-or-comment-p (point))
-             (not (equal ess-language "S")))
-            (insert ess-smart-S-assign-key)
-          (ess-insert-S-assign)))
-    (funcall #'self-insert-command (or N 1))))
+  (if (eq last-command this-command)
```
vspinu: I think we should have one-key cycling here. People should be able to press `C-c C= = = = =...`.  You can use `set-transient-map` or see tracebug code for a direct approach with event keys.

vspinu:
To be very clear here: we are removing default binding to "_" by this. A big breaking change of the most notorious "feature".

vspinu:
> I could just remove it

I think `M-_` twice should insert `_`.

lisp/ess-s-lang.el
jabranham: `simple-event-type` doesn't appear to be defined on my Emacs. Do you mean `event-basic-type`?

lisp/ess-custom.el
jabranham: Matching literally has the same problem, right? In that `<-` will match  `<<-`?

lisp/ess-s-lang.el
vspinu: Yes. This is what I meant.

jabranham:
>I think M-_ twice should insert _.

OK, done.

lisp/ess-inf.el
```diff
@@ -1959,6 +1959,7 @@ for `ess-eval-region'."
     (define-key map "\M-?"     'ess-list-object-completions)
     (define-key map "\C-c\C-k" 'ess-request-a-process)
     (define-key map ","        'ess-smart-comma)
+    (define-key map (kbd "M-_") 'ess-insert-S-assigns)
```
vspinu: I din't mean this literally. Just as an example.

jabranham: Ah, sorry for the misunderstanding. I'll remove that commit.

lisp/ess-s-lang.el
```diff
@@ -598,8 +598,7 @@ it with the character typed if it is an ASCII character,
 otherwise remove the assignment."
   (interactive)
   (let ((assign (car ess-assign-list))
-        (event (and (numberp last-input-event)
-                    last-input-event)))
+        (event (event-basic-type last-input-event)))
```
vspinu: I guess the check for ascii below is no longer necessary then. Just check for numberp and insert.

lisp/ess-s-lang.el
```diff
@@ -575,68 +575,44 @@ and one that is well formatted in emacs ess-mode."
...
-            (insert ess-smart-S-assign-key)
-          (ess-insert-S-assign)))
-    (funcall #'self-insert-command (or N 1))))
+  (if (eq last-command this-command)
+      (let ((slist ess-assign-list)
+            str)
+        (while (and (setq str (car slist)
+                          slist (cdr slist))
+                    (not (and (re-search-backward str
+                                                  (- (point) (length str)) t)
```
vspinu: BTW, I think whitespace cleanup around point might be useful here.

jabranham:
Indeed, removed.

BTW I'm pushing separate commits so that yall can seem the changes since you last looked, but it'll be easy to squash them once this is ready to merge.

lisp/ess-s-lang.el
jabranham: I was thinking that too, but am not sure how to handle all situations. Some people like to align on the assignment operator, which means that a lot of whitespace might be desirable:

```
this    <- "that"
that    <- "this"
longer  <- "this_or_that"
```

vspinu: Good point. Let's leave it as it is then.

lisp/ess-s-lang.el
jabranham: Done

jabranham:
I think this is ready to be merged, except for the failing test. I don't understand why it's failing and can't reproduce it on my end. The error message `Format specifier doesn’t match argument type` isn't particularly helpful either.

jabranham:
OK, the tests should be fixed now. When looking over what was happening I also noticed that `ess-inside-string-or-comment-p` could be simplified, so I did that. 

Let me know if all looks good and I can squash these commits together.

vspinu:
I am still not sure if we should bring this in before the release (planned for August). After that, the following ESS version (19.04 or so) will bring a lot of user visible changes.  So I would rather pile this change into 19.04. What do others think? @mmaechler ?

jabranham:
Rebased and squashed. Here's the commit summary as it stands:

```
Remove ess-smart-S-assign & new function ess-cycle-assignment  …
* ess-insert-S-assign now inserts the car of ess-assign-list, or
  replaces it with the key typed if already there.

* Removes underscore assignment from {inferior}-ess-mode-map

* New function ess-cycle-assignment cycles through the assignment
  operators in new defcustom ess-assign-list. It is bound to C-c C-=
  in ess-mode-map and inferior-ess-mode-map

Closes #584
```

vspinu:
@jabranham let's keep `_` for the next release. We will rework our keys for 19.04. I think this is the best tradeoff.  If we keep `_` binding there will be no user-level breakage, right?

vspinu:
Sure, but this will affect all users using `_`, and the most annoying "breakage" being when keybindings change.  

Let's then keep this for the following release then. We will have multiple keybinding changes then for sure. In principle we can move the development on the `next` branch and leave `master` in a feature freeze,

jabranham:
I could wrap this in an if statement and keep the key assignment, which I think would keep the current behavior and still allow easy opt-out for those who want it?

jabranham:
@vspinu  - Here's what I was talking about, BTW. Don't merge this in though because we'll need to modify `newfeat` too.

```diff
From 4993ff2e38fd9621a649d96b07cc5bf544e58aca Mon Sep 17 00:00:00 2001
From: Alex Branham <branham@utexas.edu>
Date: Thu, 28 Jun 2018 15:48:37 -0500
Subject: [PATCH] Test

---
 lisp/ess-custom.el | 10 ++++++++++
 lisp/ess-inf.el    |  2 ++
 lisp/ess-mode.el   |  2 ++
 lisp/ess-s-lang.el | 34 +++++++++++++++++++---------------
 4 files changed, 33 insertions(+), 15 deletions(-)

diff --git a/lisp/ess-custom.el b/lisp/ess-custom.el
index 0639af1b..d6edbc3c 100644
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -549,6 +549,16 @@ ess-smart-comma only, but will be enriched in the near future.")
 Used to avoid annoying jumping by ess-eval.*-and-step to end of
 buffer or end chunks etc.")
 
+(defcustom ess-smart-S-assign-key "_"
+  "Key used by `ess-smart-S-assign'.
+Should be nil or a \"simple\" key, in other words no key
+modifiers.
+
+You may change this to nil at any time. However, if you change it
+to another string, it must be set before ESS is loaded."
+  :group 'ess-S
+  :type '(choice (const :tag "Nothing" :value nil) string))
+
 (defcustom ess-assign-list '(" <- " " <<- " " = " " -> " " ->> ")
   "List of assignment operators.
 `ess-cycle-assignment' uses this list.  These strings must
diff --git a/lisp/ess-inf.el b/lisp/ess-inf.el
index ffa91104..6c332a13 100644
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1956,6 +1956,8 @@ for `ess-eval-region'."
     (define-key map "\C-c\C-d"   'ess-doc-map)
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
+    (when ess-smart-S-assign-key
+      (define-key map ess-smart-S-assign-key 'ess-insert-S-assign))
     map)
   "Keymap for `inferior-ess' mode.")
 
diff --git a/lisp/ess-mode.el b/lisp/ess-mode.el
index bd4d6650..67ff1806 100644
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -106,6 +106,8 @@
     (define-key map "\C-c\C-d"   'ess-doc-map)
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
+    (when ess-smart-S-assign-key
+      (define-key map ess-smart-S-assign-key 'ess-insert-S-assign))
     (define-key map (kbd "C-c C-=") 'ess-cycle-assignment)
     map)
   "Keymap for `ess-mode'.")
diff --git a/lisp/ess-s-lang.el b/lisp/ess-s-lang.el
index 517e23f9..bf552d3a 100644
--- a/lisp/ess-s-lang.el
+++ b/lisp/ess-s-lang.el
@@ -597,29 +597,33 @@ keypress to repeat it, so if it is bound to \"C-c C-=\" pressing
      (define-key map (kbd key) #'ess-cycle-assignment)
      map)))
 
-(defun ess-insert-S-assign ()
+(defun ess-insert-S-assign (arg)
   "Insert the first element of `ess-assign-list' unless in string or comment.
 If the character before point is the first element of
-`ess-assign-list', replace it with the last character typed."
-  (interactive)
-  (let* ((assign (car ess-assign-list))
-         (event (event-basic-type last-input-event))
-         (char (ignore-errors (format "%c" event))))
-    (cond ((and char (ess-inside-string-or-comment-p))
-           (insert char))
-          ((re-search-backward assign (- (point) (length assign)) t)
-           (if (and char (numberp event))
-               (replace-match char)
-             (replace-match "")))
-          (t (insert assign)))))
+`ess-assign-list', replace it with the last character typed. If
+`ess-smart-S-assign-key' is nil, do `self-insert-command' using
+ARG as the number of times to insert."
+  (interactive "p")
+  (if ess-smart-S-assign-key
+      (let* ((assign (car ess-assign-list))
+             (event (event-basic-type last-input-event))
+             (char (ignore-errors (format "%c" event))))
+        (cond ((and char (ess-inside-string-or-comment-p))
+               (insert char))
+              ((re-search-backward assign (- (point) (length assign)) t)
+               (if (and char (numberp event))
+                   (replace-match char)
+                 (replace-match "")))
+              (t (insert assign))))
+    (funcall #'self-insert-command arg)))
 
 (defalias 'ess--activate-smart-S-assign-key 'ignore "")
 (make-obsolete 'ess--activate-smart-S-assign-key
                "it does nothing. Use `ess-insert-S-assign' instead." "2018-06-08")
 (defun ess-disable-smart-S-assign (&rest _ignore)
   "Disable `ess-insert-S-assign'."
-  (declare (obsolete "Ignored." "2018-06-08"))
-  (warn "This function is obsolete, Use `ess-insert-S-assign' instead."))
+  (declare (obsolete "Use ess-smart-S-assign-key instead." "2018-06-08"))
+  (setq ess-smart-S-assign-key nil))
 
 (defun ess-add-MM-keys ()
   "Define MM's user keys, currently \\<ess-mode-map>\\[ess-insert-function-outline], and
-- 
2.18.0

```

vspinu:
Cool. Thanks for looking into it. I think this is the right decision for now. Don't forget to place a reminder here and there of how to remove that after the release.

lisp/ess-custom.el
```diff
@@ -549,12 +549,6 @@ ess-smart-comma only, but will be enriched in the near future.")
 Used to avoid annoying jumping by ess-eval.*-and-step to end of
 buffer or end chunks etc.")
 
-(defcustom ess-S-assign " <- "
-  "String used for left assignment in all S dialects.
```
vspinu: This was absoleted in favor of first entry in `ess-assign-list`? If so, the news should be explicit about this. It's extremely unlikely that users changed this variable, but who knows.

lisp/ess-mode.el
```diff
@@ -107,7 +107,8 @@
     (define-key map "\C-c\C-e"   'ess-extra-map)
     (define-key map "\C-c\C-t"   'ess-dev-map)
     (when ess-smart-S-assign-key
-      (define-key map ess-smart-S-assign-key #'ess-smart-S-assign))
+      (define-key map ess-smart-S-assign-key 'ess-insert-S-assign))
+    (define-key map (kbd "C-c C-=") 'ess-cycle-assignment)
```
vspinu: BTW, I am divided if we should use `C-M-=` for this. I don't think that it's a violation of emacs convention as it's generally regarded ok to rebind standard sequences:

> It is also legitimate for a major mode to rebind a standard key sequence whose standard meaning is rarely useful in that mode.

And C-M-= is not even a standard sequence. Thoughts?

doc/newfeat.texi
```diff
@@ -107,15 +107,22 @@ will need to update their customization to
...
-ess-smart-S-assign-key instead: ess-toggle-S-assign,
-ess-toggle-S-assign-key, ess--unset-smart-S-assign-key,
-ess--activate-smart-S-assign-key, ess-disable-smart-S-assign
+@item @code{ess-smart-S-assign} changed name to @code{ess-insert-S-assign}.
+By default, it provides similar functionality. Since many users
+dislike this functionality, it is easier to disable; set
+@code{ess-smart-S-assign-key} to nil. The following functions have
+been made obsolete and will be removed in the next release of ESS:
+ess-smart-S-assign, ess-toggle-S-assign, ess-toggle-S-assign-key,
+ess--unset-smart-S-assign-key, ess--activate-smart-S-assign-key,
```
vspinu: Internal `--` should probably not be listed  here. It's a user facing documentation and should be only about end users care.

doc/newfeat.texi
```diff
@@ -107,15 +107,22 @@ will need to update their customization to
...
-@item Customization of ess-smart-S-assign-key has been reworked.
-Use @code{(setq ess-smart-S-assign-key nil)} to disable smart
-assignment at any time instead of @code{(ess-toggle-underscore nil)}.
-To use another key, you should set the value of
-@code{ess-smart-S-assign-key} before ESS is loaded. The following
-functions have been made obsolete. You should customize
-ess-smart-S-assign-key instead: ess-toggle-S-assign,
-ess-toggle-S-assign-key, ess--unset-smart-S-assign-key,
-ess--activate-smart-S-assign-key, ess-disable-smart-S-assign
+@item @code{ess-smart-S-assign} changed name to @code{ess-insert-S-assign}.
```
vspinu: How about @lionel-'s earlier comment on renaming this one. 

Should we maybe name it `ess-insert<-` or `ess-smart<-`?

lisp/ess-custom.el
jabranham: It's in there already, I believe.

lisp/ess-mode.el
jabranham: I think it's OK to bind `C-M-=`, but I personally find `C-c C-=` easier to type. I don't have a strong opinion though.

doc/newfeat.texi
jabranham: Can we just remove them, then, instead of obsoleting them?

doc/newfeat.texi
jabranham: Sure, easy enough to do. I don't like using `<-` in the name since users can customize what assignment operator gets used (and non-R languages could use something else). 

How about `ess-insert-assign`?

vspinu: > How about ess-insert-assign?

Sounds good.

doc/newfeat.texi
vspinu: Yes.

lisp/ess-custom.el
vspinu: I meant in the news, but if it's in the docs it's also good enough. 

BTW, could we make first element of ess-assing-list picking values from ess-S-assign in case that one was set by the user?

lisp/ess-mode.el
vspinu: Nevermind. I thought C-M-= is much easier to type, but it looks that it's "just me".

lisp/ess-custom.el
jabranham: Is this not what you're talking about? 

```
@item The option ess-S-assign has been removed.
Customize new option @code{ess-assign-list} instead.
```

doc/newfeat.texi
jabranham: Done

doc/newfeat.texi
jabranham: OK, done

jabranham:
I think I replied to all the comments I received, let me know if I missed something. 

@vspinu - I got a comment from you about making `ess-assign-list` contain `ess-S-assign` if defined but I can't find that to reply directly to it, so I'm noting here that that's done too.

lisp/ess-custom.el
vspinu: I thought it should be a bit more explicit about "customizing first element". Nevermind. Let's not complicate things.

vspinu:
Great! Let's merge then and hope for the best :+1:

PR #587, opened 2018-06-23
Remove unsafe local vars

jabranham:
I was getting annoyed by constantly getting asked if I want to set `whitespace-style` and `outline-mode` when opening some ESS files, so this PR removes those variables and consolidates `outline-regex` to .dir-locals.

I'm not sure if anyone is using outline minor mode like it says at the end of some of the files. The fact that it's not set up like that in a lot of the newer ESS files makes me thing that it's unused, perhaps we should not set `outline-regex` in `.dir-locals` either?


vspinu:
Nice change. I don't mind removing the outline dir-locals regexp because I don't use them. But I don't think it harms either, or does it?

jabranham:
No, I don't think it really matters either way.

PR #588, opened 2018-06-26
Set dialect variables in package files for smoother interaction

lionel-:
This sets `ess-dialect` and `ess-setwd-command` in all package files. The goal is for inferior commands to work out of the box in all modes. In the future it might make sense to create a customize-alist for the package mode.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,7 +4,12 @@
 Changes and New Features in development version:
 @itemize @bullet
 
-@item @ESS{[R]}: @code{ess-r-package-use-dir} now works with any mode
+@item @ESS{[R]}: The R package mode now sets @code{ess-dialect} to
+@code{"R"} for all modes. This way interactive commands called from C or
+C++ files automatically select relevant inferior processes.
+
+@item @ESS{[R]}: @code{ess-r-package-use-dir} (which sets the working
+directory to the root of the current package) now works with any mode
 (e.g. in C or C++ files within @code{/src}).
 
 @item This is the last release to support Emacs older than 25.1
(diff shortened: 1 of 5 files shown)


doc/newfeat.texi
```diff
@@ -4,6 +4,18 @@
 Changes and New Features in development version:
 @itemize @bullet
 
+@item @ESS{[R]}: The R package mode now sets @code{ess-dialect} to
```
vspinu: This probably shouldn't be in the news. It's an internal change. Or can be something more generic "Improved support for ess commands outside ess modes".

lisp/ess-r-package.el
```diff
@@ -472,6 +472,16 @@ disable the mode line entirely."
   :lighter ess-r-package-mode-line
   (if ess-r-package-mode
       (progn
+        ;; Forward relevant R settings for interacting with inferior
```
vspinu: Similar functionality is already there. See `ess-local-customize-alist` and `ess-with-current-buffer`. I would say we reuse some of this functionality. It might be worth creating a variable with a list of r-package specific variables used for transfer.

vspinu: Actually, you are not porting them from other buffers, but rather from ess-r-customize-alist directly. That's a bit different.

vspinu:
@lionel- any progress on this and other PRs of yours? It would be good to enter soft lock from now till the release with no major changes.

lionel-:
I won't have much time for a while unfortunately.

vspinu:
No problem. Let's then merge this one in and leave other PRs for the next release. Do you have anything to add here? I will have a sweep over the news in due time and will remove internal and too technical details anyhow, so no need to change that part for now.  

BTW, once sesman is in all commands should work automatically from anywhere, so this part will be in all likelihood removed in the 19.04.

lionel-:
Sounds good. Great to hear sesman will fix this!

vspinu:
Could you please squash this one on merge? It got overly fragmented.

PR #593, opened 2018-06-29
Remove ess-comp

jabranham:
There's almost nothing in the file and it's simpler to set via command
line arguments in the Makefile


vspinu:
I wanted to do it myself one of these days :)

PR #594, opened 2018-06-29
Disable flymake during tests

jabranham:
Flymake complains about lintr not existing, which results in a lot of
unnecessary output in the travis tests. This stops that.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -13,6 +13,9 @@
 
 (require 'ess-site)
 
+;; lintr probably isn't installed in the test suite and flymake will
+;; complain about that. Disable it during tests.
+(setq ess-use-flymake nil)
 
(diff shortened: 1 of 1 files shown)


lionel-:
Should we be less verbose when lintr is not installed?

I think we should have fixed this not by disabling flymake but by installing lintr in the Travis builds, which would slightly increase the test coverage (to catch ess-r-flymake warnings/errors).

vspinu:
I am a bit divided on this one. If something runs in the background every time we create an R buffer, it's no longer a single unit of code which is tested. Plus in most (all?) cases flymake won't even have a chance to run properly because the temporary R buffers in tests are killed immediately.

lionel-:
You're right, sounds good.

PR #597, opened 2018-07-01
Separate R/S runner definition

### LINKED ISSUE #596: ess 20180629.212 is broken : doesn't start.
I follow a few MELPA packages via via `emacs`'s package panager (i.e. in my directory, thus shadowing systemwide packages installed via Debian packages), among which `ess`. A recent update broke it.

Starting `ess` (M-x R) from a "virgin" `emacs` session fails with the following message : 
```
ess-r-s-define-runners: Symbol’s function definition is void: ess-s-define-runners
```

Removing MELPA's `ess` (thus relying on systemwide ess package as packaged in Debian testing, which is currently at 17.11-2), allows to start `ess`.

Let me know what further information you might need.

jabranham:
R runners created when ess-r-mode is loaded, S runners when ess-sp6-d is loaded.

Moving code around I got into the fun land of ESS's requires/provides dance, let me know if something looks wrong/off. 

Closes #596

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -434,11 +434,6 @@ Set this variable to nil to disable searching for other versions of R.
 If you set this variable, you need to restart Emacs (and set this variable
 before ess-site is loaded) for it to take effect.")
 
-;; Create functions for calling different (older or newer than default)
-;; versions of R and S(qpe).
-(defvar ess-versions-created nil
-  "List of strings of all S- and R-versions found on the system.")
-
 (defvar ess-r-created-runners nil
   "List of R-versions found from `ess-r-versions' on the system.")
 (define-obsolete-variable-alias 'ess-r-versions-created 'ess-r-created-runners "2018-05-05")
@@ -665,14 +660,12 @@ as `ess-r-created-runners' upon ESS initialization."
       ;; Iterate over each string in VERSIONS, creating a new defun each time.
       (setq ess-r-created-runners
             (mapc (lambda (v) (ess-define-runner v "R")) versions))
-      (setq ess-versions-created (append ess-versions-created
-                                         ess-r-created-runners))
       ;; Add to menu
-      (when ess-versions-created
+      (when ess-r-created-runners
         ;; new-menu will be a list of 3-vectors, of the form:
         ;; ["R-1.8.1" R-1.8.1 t]
         (let ((new-menu (mapcar (lambda(x) (vector x (intern x) t))
-                                ess-versions-created)))
+                                ess-r-created-runners)))
           (easy-menu-add-item ess-mode-menu '("Start Process")
                               (cons "Other" new-menu)))))))
 (define-obsolete-function-alias
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -443,81 +443,6 @@ before ess-site is loaded) for it to take effect.")
   "List of R-versions found from `ess-r-versions' on the system.")
 (define-obsolete-variable-alias 'ess-r-versions-created 'ess-r-created-runners "2018-05-05")
 
-(defun ess-r-s-define-runners ()
-  "(Re)Create ESS R-X.Y.Z and S-X.Y.Z commands.
-See `ess-r-versions' and `ess-s-versions' for which functions
-will be (re)created."
-  (interactive)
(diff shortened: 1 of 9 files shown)


lisp/ess-r-mode.el
```diff
@@ -739,7 +664,17 @@ as `ess-r-created-runners' upon ESS initialization."
                                      ess-r-versions)))))))
       ;; Iterate over each string in VERSIONS, creating a new defun each time.
       (setq ess-r-created-runners
-            (mapc (lambda (v) (ess-define-runner v "R")) versions)))))
+            (mapc (lambda (v) (ess-define-runner v "R")) versions))
+      (setq ess-versions-created (append ess-versions-created
```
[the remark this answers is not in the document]
jabranham: Yes, that's right. Should it not?

Can we just remove ess-versions-created in favor of the R/S specific ess-r-versions-created?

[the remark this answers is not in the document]
vspinu: > Yes, that's right. Should it not?

What's the use of duplicates? If it's not useful then it should be uniqified here.

> Can we just remove ess-versions-created in favor of the R/S specific ess-r-versions-created?

This sounds right. `ess-versions-created` is an internal variable.  I think we can safely remove it.

lisp/ess-r-mode.el
```diff
@@ -443,81 +443,6 @@ before ess-site is loaded) for it to take effect.")
   "List of R-versions found from `ess-r-versions' on the system.")
 (define-obsolete-variable-alias 'ess-r-versions-created 'ess-r-created-runners "2018-05-05")
 
-(defun ess-r-s-define-runners ()
```
vspinu: You have removed this, but where did you put it back? I don't see it in this diff.

lisp/ess-r-mode.el
```diff
@@ -2219,6 +2154,42 @@ otherwise nil."
...
+                   (ess-find-rterm (concat P-1 "/R/") "bin/x64/Rterm.exe")
+
+                   ;; Always 64 on 64 bit OS, nil on 32 bit OS
+                   (ess-find-rterm (concat P-2 "/R/") "bin/Rterm.exe")
+                   (ess-find-rterm (concat P-2 "/R/") "bin/i386/Rterm.exe")
+                   (ess-find-rterm (concat P-2 "/R/") "bin/x64/Rterm.exe")))
+              (let ((PF (getenv "ProgramFiles")))
+                (nconc
+                 ;; Always 32 on 32 bit OS, depends on 32 or 64 process on 64 bit OS
+                 (ess-find-rterm (concat PF "/R/") "bin/Rterm.exe")
```
vspinu: I think ess-find-rterm should be refactored into handling these 3 cases internally. I think the new logic here doesn't match with old logic in ess-find-rterm, this is why Henric having problems on the mailing list.

lisp/ess-r-mode.el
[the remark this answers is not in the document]
jabranham: OK, I just removed it.

lisp/ess-r-mode.el
jabranham: I didn't put it back. I don't think it's needed now. R runners are created when loading ess-r-mode and S runners are created when loading ess-sp6-d, so there's no need for this function anymore (I think).

lisp/ess-r-mode.el
jabranham: I find the situation on Windows pretty confusing TBH. It probably doesn't help that I don't have a windows machine to test on.

Have you taken a look at this: https://github.com/jabranham/ESS/commit/43501f19167d94dbe8ef5a604bcafe53183ec8e3 ? I think that would solve Henric's problem but I haven't heard back from him yet.

lisp/ess-r-mode.el
vspinu: Arh. Right. I assumed it was renamed into `ess-r-define-runners`, but of course, that's a different function.

lisp/ess-r-mode.el
vspinu: Yeh, it's a bit puzzling but you are probably right with that lexical binding. That builder was the only change in logic as far as I can judge.

 I think all of us have to setup a windows in a virtual box and test it there at least occasionally. 

> Also in my mind that's a separate PR since this one deals with separating R/S runner definition 

Indeed.

PR #601, opened 2018-07-05
Obsolete unpopular languages, ess-mouse, and essd-els

### LINKED ISSUE #595: Obsolete several languages?
Since the current plan AFAIU is to re-write how ESS mode is defined after the next release, we should probably obsolete all the languages we don't want to rewrite. We can put an entry in newfeat along the lines of:

"Support for several languages have been obsoleted and will be removed in the next release.
Please contact us if you still use these, if there is enough demand we may continue to maintain them. These include..."

And here's a list of files we could obsolete (this leaves support for R, Stata, SAS, julia, and bugs/jags):

- ess-dde
- essd-els
- ess-gretl
- ess-omg-d
- ess-omg-l
- ess-lsp-l
- ess-vst-d
- ess-xls-d
- ess-arc-d
- ess-r-gui
- ess-s3-d
- ess-s4-d
- ess-sp3-d
- ess-sp4-d
- ess-sp5-d
- ess-sp6-d
- ess-sp6w-d

Thoughts?

jabranham:
These are mostly languages that have fallen out of use. When we clean
up ESS and use define-derived-mode, these languages won't work
anymore. Thus, we're obsoleting them now for the 18.09ish release, and
will remove them after that is released.

Closes #595


vspinu:
Thanks Alex. This still has to be discussed internally. Martin is traveling at the moment. In any event this could go into the package only after the release in August.

mmaechler:
S-plus is definitely still used in the industry.
Xlispstat is not used productively, but AFAIK you could still get and install it, and the only nice (:-) interface to use it would be emacs with ESS.

Let's not become overzealous here.

Relatedly note that till ca two years ago I had spent many hours to build ESS in a way that you could also use it with R version 1.x.y and even older ones, by consciously replacing all new features in our own R code that is needed at ESS - startup by those that were present in R very early.   Now  "you cool kids" have broken that all around and using such old R versions has with ESS has become basically infeasible.

Not nice to me and the few others out there who occasionally do aRcheology..

jabranham:
The eventual goal is to rewrite the major modes we support using `define-derived-mode`. So `ess-mode` derives from `prog-mode`, and `ess-r-mode` derives from `ess-mode`. This will solve a lot of bugs that keep cropping up. 

However, making `ess-mode` derive from `prog-mode` will break all the major mods ESS supports. So we'll have to invest some effort in fixing the major modes. Which is fine, but I doubt anyone has any inclination to fix S+3 or xlispstat. My idea here was to deprecate them in the next release (18.09?) and remove them after release so that we can get to work on switching to define-derived-mode. 

If someone does want to get and install xlispstat or S+, they can always run ESS 18.09.

> using such old R versions has with ESS has become basically infeasible.

As far as I can tell, there was a decision made in 2009 by ESS Core devs to only support recent R versions: 69de2673fe74ba19609e3d2d0a9924b0eab91ae7

lionel-:
I doubt we still have any users of these modes and I think it's not desirable to distribute unmaintained code. If on top of that this old code hinders efforts of modernisation, then I agree it should be obsoleted. These modes could make a comeback when/if they find a new maintainer.

jabranham:
I'm not sure SPlus is used anywhere anymore. According to wikipedia, the last release was ~7 years ago. I've spent some time looking over TIBCO's (the company that owns S-plus) website looking for how to get Splus and can't find anything. I think it used to be included in their "Spotfire" product, but it has since transitioned to using some proprietary R engine. 

If there are any Splus users left out there, I don't think it's unreasonable to ask them to remain on ESS 18.09. 

OTOH rewriting all the Splus modes to use `define-derive-mode` is a *lot* of work, most of which I can't really test since I don't have a Splus license. On top of that, I'd really rather see #616 merged as soon as possible after releasing 18.09 so that it gets as much testing as possible.

vspinu:
I think we need to preserve one S-plus generic mode, based on the the most recent file (s6 I guess). Testing could be done by making an S link to R executable (ideally an old one). We just provide basic functionality there and be done with it.

mmaechler:
I agree we should keep one S(-PLUS) mode.
It's less than a year that I heard of people in industry actively working with S-PLUS.

jabranham:
> based on the the most recent file (s6 I guess

Well, `ess-site` loads `ess-sp3d` and then either `ess-sp6-d` or `ess-sp6w-d` depending on your platform. It seems like `ess-sp6w-d` should only get loaded on windows, but it ends up getting loaded on my machine anyway (linux). So I've got no idea what's loading what there.

> testing could be done by making an S link to R executable

You get an error if you try that: `S+: The emacs variable ‘inferior-S+-program’ does not point to S-Plus 6 or 7 or 8.  Please add ‘splus[678]?/cmd’ (expand the ‘[678]?’ to match your setup) to your ‘exec-path’ or specify the complete path to ‘Splus.exe’ in the variable ‘inferior-S+-program’ in your ‘.emacs’ file.` Easy to work around, but just one example of the various kinds of unpleasantness you'll encounter mucking about in these old files.

> It's less than a year that I heard of people in industry actively working with S-PLUS.

When's the last time we had a bug report from an S+ user? And I guarantee you it's not because the S+ modes are bug-free :-) 

Again, I don't think it's unreasonable to ask any S+ users to remain on ESS 18.09. S+ hasn't had a release in 7 years. Not a single ESS developer uses S+. As far as I can tell, you can't even purchase a license anymore. We shouldn't be bending over backward trying to maintain support for old, proprietary languages.

I've already spent too much time looking through the ~7 various `ess-s/p` files trying to figure out how to support S+ using `define-derived-mode` for #616. I've got no further interest in doing so, so either A) someone else can, B) we can drop them, or C) I can close #616  because it will break S+ modes. Do let me know which though because rebasing that PR isn't always the easiest :-)

vspinu:
And indeed. In an unlikely event if someone will ever need spNd/w.el they can use ESS-18-9, or even better S+ UI directly.

vspinu:
Ok. I will fix that part.

vspinu:
Why was this one closed? We will pull it in the release straight away.

jabranham:
I must've misunderstood something somewhere. I thought the plan was to merge #616 shortly after releasing ESS 18.09. Since that'll break all these modes (S3-mode and OMG-mode for examples), we'll have to remove those files rather than just obsoleting them (which is what this branch did).

vspinu:
I see.

 Removing is fine. I thought that you have found a way to obsolete those and still make #616 happen.

PR #604, opened 2018-07-09
Fix runner definition on Windows

jabranham:
Someone on the mailing list pointed out that runner definition on Windows is broken. This is an attempt to fix it. 

I don't have any way of checking if this works, though. Does anyone have a windows machine and can test this?


vspinu:
Let's go back on ESS list for this.

PR #606, opened 2018-07-11
Rename ess-mode.el to ess.el

### LINKED ISSUE #592: ess.el should be required
Now that we've moved towards autoloading everything, it would be nice if `ess.el` would be required at some point so that people could do:

```
(with-eval-after-load "ess" 
    <ess config>)
```

which is what `use-package` does (kind of) if you `(use-package ess :defer t)`.

Since there's basically nothing in `ess.el` right now, one easy way of doing this would be to rename a file that everything relies on to `ess.el`. I nominate `ess-custom.el` as a candidate.

Any objections?

(related to #589)

jabranham:
Most of the explanation is in the first commit message, which I'll copy/paste here:

---

Rename ess-mode.el to ess.el
This lets people defer loading ESS and instead rely on autoloads. You
can do something like the following:

(with-eval-after-load "ess" <ess-config-here>)

as well as use `use-package':

(use-package ess
  :bind
  (:map ess-mode-map
        ("=" . ess-cycle-assignment))
  :config
  (setq ess-ask-for-ess-directory nil))

and then ESS will only load what it needs when it needs it, rather
than the traditional (require 'ess-site), which is rather expensive.

Closes #592


doc/newfeat.texi
```diff
@@ -131,6 +131,16 @@ See the docstring for details. New user customization option
 @code{ess-assign-list} controls what assignment operators are cycled
 through.
 
+@item The ESS initialization process has been further streamlined.
+Users now should not need to @code{(require 'ess-site)} at all, as
```
vspinu: This is only valid for packaged ESS from MELPA, right?

lisp/ess.el
```diff
@@ -1,106 +1,1019 @@
...
+                               (or elpa-rev "")))
+    (when (string= ess-revision "")
+      (setq ess-revision "<unknown>"))
+    (concat ess-version " [" ess-revision "]")))
+
 
 (provide 'ess)
 
-;;; ess.el ends here
+;;; ess-mode.el ends here
```
vspinu: Forgot this line?

jabranham: Indeed, fixed & squashed.

doc/newfeat.texi
vspinu: The news entry should be very clear about that. There are a range of other methods to install ESS and for those distributions people would still need to load ess-site.

vspinu:
:+1: Thanks!

lionel-:
I'm not sure I like this change. There's a strong convention for major mode files to end in `-mode.el`.

@jabranham The way you have split the file makes it harder to git blame. It seems it still works with the `-C` flag to detect lines moved or copied between files, but I'm not sure how robust that is:

```lisp
(let ((magit-blame-arguments '("-w" "-C")))
  (magit-blame))
```

You can preserve the full history with this snippet:

```sh
# Set these to the file you'd like to split and a target file:
SRC=ess-r-mode.el
TARGET=ess-r-indent.el

git mv $SRC $TARGET
git commit -n -m 'Save '$TARGET' history'
SAVED_HASH=`git rev-parse HEAD`
git reset --hard HEAD~1

git mv $SRC _temporary.R
git commit -n -m 'Save '$SRC' history'

git merge $SAVED_HASH
git commit -a -n -m 'Split '$TARGET' from '$SRC' 1/3'

git mv _temporary.R $SRC
git commit -n -m 'Split '$TARGET' from '$SRC' 2/3'
```

You will get two identical but separate files with full history. The final step is to manually remove the parts that don't belong to the file (that's the missing 3/3 commit).

vspinu:
Yeh. The moving part we overlooked. 

Regarding ess-mode I think we should have two files ess-mode.el and a very lightweight ess.el with metadata and some super basic generic utilities. 

@jabranham ess-mode is not parent of all modes right?

lionel-:
> Regarding ess-mode I think we should have two files ess-mode.el and a very lightweight ess.el with metadata and some super basic generic utilities.

Yup that'd be better. I'd even move the basic utilities to ess-utils so no file in ESS needs to require `ess.el`, it'd be entirely user-facing.

vspinu:
People might be using "ess-mode" in eval-after-load already (as just occurred on ess mailing list). Can we alias ess-mode to ess for the patch release? Is this even possible?

lionel-:
Perhaps we can `provide` twice?

vspinu:
Doesn't work with eval-after-load unfortunately. It takes a name of the file. In case we decide for another patch release we should probably split the relevant part into ess-mode.el into the master.

lionel-:
Please no split, just a straight rename so we don't complicate the git history.

vspinu:
We need ees.el.

lionel-:
yeah so first rename, then create ess.el.

jabranham:
Why can't we leave `ess.el` as-is? Plenty of other modes are defined without -mode on the end of the filename (python and org just off the top of my head). 

If we want (eval-after-load "ess-mode.el") to work we can create ess-mode.el that just requires ess.el?

lionel-:
I really prefer to follow the majority of prog-modes here.

lionel-:
Also it's not internally consistent with ess-r-mode.el. I keep being suprised not to find ess-mode.el when navigating files.

jabranham:
I don't have a strong opinion but do you have any data that it really is the majority of modes that are in lang-mode.el? Looking through the GNU Emacs source code, looks like plenty aren't in -mode.el: python, js-mode, octave-mode, gud-mode, scheme-mode, inf-lisp, and more. 

It's consistent with `ess-inf` which has `inferior-ess-mode`.

mmaechler:
I also would like to keep the old name.   The pros of change seem to be vastly outweighed by the cons, from my point of view.  (... given that all this is not an important issue at all ...).   
The old saying  "Don't fix it if it ain't broke"  seems to apply strongly here...

lionel-:
I don't have data, no.... But looking at the progmodes folder plus a few others: **ada-mode**, basm-mode, bat-mode, **cc-mode**, **clojure-mode**, cperl-mode, dcl-mode, **elisp-mode**, **haskell-mode**, m4-mode, make-mode, **perl-mode**, **prog-mode**, **ruby-mode**, vera-mode, vhdl-mode, and others.

More to the point, it feels tidier with -mode, you know that you'll find the mode defined there. It wouldn't make sense to rename `ess-r-mode.el` to `ess-r.el` to me.

I'm ok with renaming `ess-inf.el` to `inf-ess.el` for consistency with `inf-lisp.el` and `inf-clojure.el` (though may be best not to touch). Note that inferiors are not progmodes.

jabranham:
Alright, seems like I'm in the minority and like I said I really don't have a strong opinion.

So we'll do:

1. rename ess.el to ess-mode.el
2. create ess.el with just `(require 'ess-mode) (provide 'ess)`

right?

Let's not rename `ess-inf.el` to `inf-ess.el`. We should really be "namespacing" everything with ess- which makes it easier for users to discover options/commands (e.g. M-x apropos-user-option ess-)

jabranham:
Can we also make sure that this doesn't complicate rebasing #616? I *think* git should be fine with renaming files but am not certain...

vspinu:
> right?

The other way around I would say. `ess` should be required from `ess-mode` and `ess-mode` is required from elsewhere. Then people can do with eval after load with both. 

The metadata should stay in ess.el and relevant package specific tooling (versions, path construction) should also be there. Maybe `(require 'ess-custom)` as well.

vspinu:
To avoid complications with rebasing we could create ess-mode.el with one line `(require 'ess)` and replace all `(require 'ess)` with `(require 'ess-mode)`. Then do the actual moving in next.

PR #608, opened 2018-07-16
Implement ess-r-mode-map

### LINKED ISSUE #607: ess-insert-S-assign was turned on in SAS
ess-insert-S-assign is only conditioned on ess-smart-S-assign-key, and is custimized as "_" as default if ess-smart-S-assign-key is non-nil. As they are all global variable and function, so that, ess-insert-S-assign will be binded to "_" for all ess-mode including SAS. 

Could we make ess-smart-S-assign-key as buffer-local variable? Or enable ess-insert-S-assign as hock to R/R inferior mode? I think it is not efficient and necessary to check whether it is in R/R inferior mode in ess-insert-S-assign.

jabranham:
Move ess-insert-assign there so it doesn't get bound in e.g. SAS mode

Closes #607

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -69,6 +69,15 @@
 
 ;;*;; Mode definition
 
+(defvar ess-r-mode-map
+  (let ((map (make-sparse-keymap)))
+    (when ess-smart-S-assign-key
+      (define-key map ess-smart-S-assign-key 'ess-insert-assign))
+    map)
+  "Keymap for `ess-r-mode'.")
+;; TODO: Remove this after using define-derived-mode:
+(set-keymap-parent ess-r-mode-map ess-mode-map)
+
 ;;;*;;; UI (Keymaps / Menus)
 ;;;###autoload
 (defvar ess-dev-map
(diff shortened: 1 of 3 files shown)


vspinu:
Same should be done for inferior mode right?

It's hard to see how this change will affect users. Some folks (myself included) simply bind "_" to nil in ess-mode to get rid of smart underscore. This change will break such config.  This smart underscore refactoring got us in a never ending loop and would be good if we don't introduce more problems with radical changes for next release. 

For the next release I would rather do whatever ess-smart-underscore was doing before. Was it checking for ess-dialect?

jabranham:
> simply bind "_" to nil in ess-mode

I don't understand what you mean here. Do you mean you do something like `(define-key ess-mode-map "_" 'self-insert-command)`? 

Anyway, I'll close this PR and open another implementing the old behavior.

vspinu:
> Do you mean you do something like (define-key ess-mode-map "_" 'self-insert-command)?

Just `(define-key ess-mode-map "_" nil)`. This is how I deactivate keys in all modes, not only ESS.

PR #610, opened 2018-07-18
rename r-mode to R-mode in autoloads

nverno:
This renames r-mode to R-mode in the autoloads since the fset isn't autoloaded, 
I get an error about unknown function when first loading an R file.


lionel-:
Thanks!

PR #611, opened 2018-07-19
Fix imenu in R buffers

jabranham:


vspinu:
I think the idea was to set this in customize alist rather than directly, but I don't think that makes sense in this case because Imenu is really mode specific. Fix is good. Thanks!

PR #614, opened 2018-07-23
Remove function inferior-ess-last-prompt-end

jabranham:
We were setting syntax-begin-function to that in inferior-ess modes,
but syntax-begin-function was obsoleted in Emacs 25.1. The comment
suggests that we were doing that to avoid applying the font-lock
string face to the prompt, but Emacs is smart enough not to do that

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -279,9 +279,6 @@ This may be useful for debugging."
             (unless (and proc (eq (process-status proc) 'run))
               (error "Process %s failed to start" procname))
 
-            ;; don't font-lock strings over process prompt
-            (set (make-local-variable 'syntax-begin-function) ;; fixme: obsolete in emacs 25.1
-                 #'inferior-ess-last-prompt-end)
             (set (make-local-variable 'font-lock-fontify-region-function)
                  #'inferior-ess-fontify-region)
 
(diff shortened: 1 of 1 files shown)


jabranham:
> What made you think so?

comint-mode uses field properties to help it find and colorize the prompt so I figured that would take care of problems like this.

vspinu:
It does not. First, those field properties are taking effect only for the first prompt when you enter commands at the repl. All other prompts are missed. I played with those properties quite a bit and font-lock doesn't honor those for some unclear reason. 

Also, the inline comment is not about prompts per se, but about spilling font-lock over/across prompts. 

In any case our fontify-region-function does a good job there and the syntax-begin-function should not be necessary any longer. Some other tweaks are needed for external tools like smartparents to get the beggining of syntax right, but I have to still figure it out in emacs 25+

jabranham:
> First, those field properties are taking effect only for the first prompt when you enter commands at the repl. 

I guess this explains why `comint-prompt-read-only` doesn't work on all prompts in inferior buffers. 

> Also, the inline comment is not about prompts per se, but about spilling font-lock over/across prompts.

Do you have an example of how/when that can happen? 

> In any case our fontify-region-function does a good job there and the syntax-begin-function should not be necessary any longer.

:+1:

vspinu:
Historically we used to treat `'` as string delimiter in the inferior. This resulted in a lot of unmatched delimiters with R messages. We no longer highlight `'` strings in the inferior but I believe unmatched `"` could still occur occasionally with interrupted printing of long strings.

lionel-:
Ideally we should detect printing of data structure and fontify accordingly. And print lengthy data structures with configurable folding/unfolding. All other outputs would be unfontified. This way only the visible parts of printed data would need to be fontified. I plan to work on customisation of printing/autoprinting in base R at some point, I'll keep the IDE needs in mind.

vspinu:
This is actually relatively easy (at least in the basic form) on the emacs side. As we have a temporary cache buffer with all the output, we could print only first and last X number of lines and fold the inner parts.

Issue #562, opened 2018-06-09
ess fails to start when emacs instance started with git mergetool.


krivit:
This is under Debian testing with the following relevant versions:

* GNU Emacs 25.2.2 (x86_64-pc-linux-gnu, GTK+ Version 3.22.30)
* ESS version 17.11

I had submitted it to Debian BTS, and was asked to forward it here.

# To reproduce #

1) Set up a git merge conflict:

```
# Create the repository.
git init test
cd test

# Create and commit a file containing '0' (a syntactically correct R file).
echo '0' > test.R
git add .
git commit -m 'Zero.'

# Create a branch in which '0' is replaced by '1'.
git branch one
git checkout one
echo '1' > test.R
git add .
git commit -m 'One.'

# Replace '0' with '2' in the mater branch.
git checkout master
echo '2' > test.R
git add .
git commit -m 'Two.'

# Attempt a merge, creating a conflict.
git checkout one
git merge master
```

2) Try running mergetool with emacs as editor:

```
git mergetool -t emerge
```

# Expected result: #

A three-way merge interface comes up with files open in ESS R mode.

# Actual result: #

Empty emacs buffer appears, with the following message in the minibuffer:
```
ess-mode: Customise alist is not specified, nor  ess-local-customize-alist is set.
```

eddelbuettel:
Thanks for bringing it here.  For the record the initial bug report was [here](https://bugs.debian.org/901122).

jabranham:
How did you install ESS, and can you post the value of `auto-mode-alist`? 

It would be helpful to see a backtrace too. Can you put `(setq debug-on-error t)` in your Emacs config file, produce this bug, and post what appears in the `*Backtrace*` buffer?

lionel-:
I think the cause of this kind of failure is that `major-mode` for all ESS modes is `ess-mode` and running `(ess-mode)` doesn't work.

jabranham:
Right, I'm trying to understand how `(ess-mode)` gets called. Shouldn't `.R` files call `R-mode` from `auto-mode-alist`?

krivit:
@jabranham, I installed ESS by installing the Debian package `ess`. Here's the rest of it:

## `auto-mode-alist` value ##
```
(("\\.sp\\'" . S-mode)
 ("/R/.*\\.q\\'" . R-mode)
 ("\\.[qsS]\\'" . S-mode)
 ("\\.ssc\\'" . S-mode)
[...]

PR #616, opened 2018-07-23
Use define-derived-mode for ESS major modes

jabranham:
Here's a list of all the major modes ESS currently defines, sorted by which mode I intend to derive them from. I'm also implementing `ess-inferior-<lang>-mode` for the inferior buffers. 

Current status by language:
- R: everything should work
- julia: everything should work... probably?
- SAS: Likely needs work
- Stata: Likely needs work
- S+: broken
- gretl: works

# prog-mode
- [x] ess-mode
# ess-mode
- [x] ess-bugs-mode
- [x] ess-jags-mode
- [x] ess-transcript-mode
- [x] R-mode
- [ ] S+-mode
- [x] SAS-mode
- [x] STA-mode
- [X] gretl-mode
# comint-mode
- [x] inferior-ess-mode
# inferior-ess-mode
- [x] ess-inferior-r-mode
- [x] ess-inferior-julia-mode
- [ ] ess-inferior-sas-mode (I can't check if this works)
- [x] ess-inferior-stata-mode (I can't check if this works)
- [ ] ess-inferior-s-mode
# ess-transcript-mode
- [x] R-transcript-mode
- [ ] S+-transcript-mode
- [x] STA-transcript-mode
# text-mode
- [x] Rd-mode (perhaps tex-mode would be better though?)
# SAS-mode
- [x] SAS-log-mode
# special-mode
- [x] ess-help-mode
- [x] ess-rdired-mode
- [x] ess-rutils-mode
- [x] ess-watch-mode
- [x] sas-dir-mode
- [x] ess-stata-help-mode
- [x] SAS-listing-mode


vspinu:
I agree. We should derive from comint and have inferior-ess-r-mode, or maybe ess-inf-r-mode. And yes, this will have to wait till the next release.

jabranham:
I've also changed `ess-mode` to derive from `prog-mode` and `ess-r-mode` to derive from `ess-mode`. This seem to be working for R and inferior R stuff. 

I think stata mode is working, but inferior stata is likely broken. Hard to check though since I don't have a stata license. 

Everything else is probably broken at the moment :-)

jabranham:
Closes #562

vspinu:
I am afraid I will have to push a change to font-lock system (as it's currently broken) which will affect quite a bit of language files. Most of the changes are deletions but will surely affect this branch. Hopefully this will be the last major change before the release.

jabranham:
Rebased off master.

jabranham:
Rebased off the `next` branch

jabranham:
@vspinu I'm trying to fix a failing test: `ess-inf-send-cat-some.text-test` but I don't understand why `output` has a leading newline. `cat` doesn't output text starting with a newline. Can I apply this diff or am I missing something?

```diff
diff --git a/test/ess-inf-tests.el b/test/ess-inf-tests.el
index 441e22f6..4d32cfaf 100644
--- a/test/ess-inf-tests.el
+++ b/test/ess-inf-tests.el
@@ -115,8 +115,7 @@ cleaned-prompts >
   (let ((input "cat(\"some. text\\n\")
 head(cars, 2)
 ")
-        (output "
-some. text
+        (output "some. text
 > 
   speed dist
 1     4    2

```

the other test currently failing `test-org-ob-R-data-frame` complains about selecting a deleted buffer. I'm not quite sure why it's failing right now. It passes locally for me, both interactively and in a batch run.

vspinu:
> Once I fix the final failure on travis I'll merge this into next assuming that's OK with everyone.

I would like to have a cursory look before the merge. Maybe @lionel- would like too. Let's wait by the end of the week. People are probably ignoring this thread by now due to the huge number of updates.

jabranham:
Works for me

jabranham:
Great, rebased off `next`.

lisp/ess-inferior-r.el
```diff
@@ -0,0 +1,250 @@
...
+    (define-key map "ss" 'ess-skip-to-help-section)
+    (define-key map "su" 'ess-skip-to-help-section)
+    (define-key map "sv" 'ess-skip-to-help-section)
+    map)
+  "Keymap for `ess-r-help-mode'.")
+
+;; TOTHINK: Prevent string delimiting characters from messing up output in the
+;; inferior buffer
+(defvar ess-inferior-r-mode-syntax-table
+  (let ((tab (copy-syntax-table ess-r-mode-syntax-table)))
```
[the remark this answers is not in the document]
lionel-: In the progmodes folder I seet lots of `st` and `table`, and one `tbl`. I'd suggest using one of the first two to follow convention. My preference is `table`.

lisp/ess.el
```diff
@@ -333,41 +318,17 @@ Variables controlling indentation style:
...
-  (unless is-derived
-    (setq major-mode 'ess-mode)
-    (setq mode-name (concat "ESS[" (or ess-dialect ess-language) "]")))
-  ;; The following line does the next 20 or so :-).
-  (ess-write-to-dribble-buffer
-   (format "(ess-mode-1.6): editing-alist=%s \n"
-           ess-mode-editing-alist))
-  (ess-setq-vars-local ess-mode-editing-alist)
+indentation style. See `ess-style-alist' for predefined styles."
+  ;; TODO: get rid of these and rely on modes to set variables properly
```
lionel-: Yes! I'm not sure what's the best way to do this. Do we need mode-specific variables? Do we need inheritance across the mode hierarchy? Do we need a different approach for user-facing variables and private variables?

test/ess-tests.el
```diff
@@ -32,4 +29,4 @@
   (should (string= 'ess-julia-mode (with-temp-buffer (ess-julia-mode) major-mode))))
 
 (ert-deftest sas-mode ()
-  (should (string= 'ess-mode (with-temp-buffer (SAS-mode) major-mode))))
+  (should (string= 'SAS-mode (with-temp-buffer (SAS-mode) major-mode))))
```
lionel-: Probably should be `sas-mode`?

lisp/ess-inferior-r.el
```diff
@@ -0,0 +1,250 @@
+;;; ess-inferior-r.el --- inferior-r-process         -*- lexical-binding: t; -*-
```
lionel-: I like the new file but it is very important to me that we keep the full git history by splitting files the hard way as outlined in https://github.com/emacs-ess/ESS/pull/606#issuecomment-431657961.

I have been guilty of butchering the history in past refactorings, which I regret.

lisp/ess-inferior-r.el
```diff
@@ -0,0 +1,250 @@
...
+;; Support for starting, running, and otherwise interacting with R
+;; inside Emacs buffers. See the ESS Info manual for more information.
+
+;;; Code:
+
+(require 'ess-help)
+(require 'ess-r-mode)
+(require 'ess-inf)
+
+(defvar ess-inferior-r-mode-map
```
lionel-: This is not a conventional namespace. If we're going to create a new ns, it should probably be `inf-ess-` to follow convention. There are several cases where the namespace is not a prefix, e.g. you don't write `ess-with-` but `with-ess-`.

So we should use `inf-` like in `inf-haskell-mode`, `inf-clojure-`, . It's also fine to keep `inferior-ess-` as in `inferior-python-`, `inferior-tcl-`, `inferior-octave-`, though I like the conciseness of `inf`.

One interesting case is python where they define `inferior-python-mode`, with accompanying map and hook variables, but otherwise namespace most functions and variables with `python-shell-`. I would also consider this solution.

lisp/ess-inferior-r.el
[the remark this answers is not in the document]
lionel-: In fact I see we already use `table` elsewhere in the code so we should go with that.

lisp/ess-utils.el
```diff
@@ -491,6 +492,11 @@ etc.")
     ["no" (lambda () (interactive) (setq ess-eval-visibly nil))
      :style radio :enable t :selected (eq ess-eval-visibly nil) ]))
 
+(defun ess-derived-mode-p ()
```
lionel-: This will go away once ess-julia-mode derives properly right? If so, should be a private `ess--` function.

lisp/ess.el
jabranham: We should be setting variables in the "highest" mode and using inheritance for most of this.

lisp/ess-inferior-r.el
jabranham: The more I work with it the less I'm sure about splitting this into a new file. The inferior stuff is pretty tightly integrated with the "normal" major mode stuff. It forces us to put stuff like:

```elisp
;; ensure `R' is defined if the user didn't load the autoloads file
(autoload 'R "ess-inferior-r")
(autoload 'run-ess-r- "ess-inferior-r")
```

in ess-r-mode.el, which is a little odd. What do you think about keeping this all in ess-r-mode.el?

lionel-: Sounds good, we can always split later on if needed.

lisp/ess-inferior-r.el
jabranham: I'm not a huge fan of abbreviating long words to shorter words, so how about `inferior-ess-r-mode`?

lionel-: ok let's do that, it has the clear advantage of being the status quo. We already have functions in that namespace, e.g. `inferior-ess-r-force`.

lisp/ess-utils.el
jabranham: There's two ways to go about this:

1. (what we're doing now) let ess-julia-mode derive from julia-mode. This has the benefit of using their syntax table, etc but we have to override the keymap with our own
2. derive ess-julia-mode from ess-mode. This basically flips the first situation and now we're using our keymaps by default but would need to override "our" syntax table with theirs.

I hadn't thought about number 2 until you made this comment, but it probably makes sense to do that. In that case, yes this function will eventually go away.

lionel-: Perhaps a crazy thought, but should ESS be a minor mode?

lisp/ess-inferior-r.el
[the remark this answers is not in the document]
jabranham: OK, `table` it is.

lisp/ess.el
lionel-: Have you looked into CEDET by the way? They implement mode-local functions and variables. I think this was one of the inspiration for the mode generics that @vspinu implemented.

test/ess-tests.el
jabranham: There are weird functions that convert the string ess-dialect into a symbol and call that symbol in order to start the inferior mode so we'll at least have to provide SAS-mode as an alias.

lisp/ess-utils.el
jabranham: I don't think it's totally crazy but am certain it's outside the scope of this PR :-)

test/ess-tests.el
lionel-: I see. Do as easiest, it's not that important.

lisp/ess-utils.el
lionel-: Agreed, we can keep thinking about it.

jabranham:
@lionel- thanks for the review. I'm going to make the changes we discussed (anything I didn't comment on I agree with you) and force-push here later if that's OK with you.

jabranham:
OK, updated and force pushed. Took the opportunity to clean up a few commits along the way where I accidentally fixed whitespace. Sometimes `aggressive-indent-mode` can be too aggressive. I wish there was an equivalent to `ws-butler` for the indent world where it would only aggressively indent lines that I've modified.

jabranham:
@vspinu would be nice to merge this this week. Do you think you'll have time to look it over? No rush, of course.

vspinu:
Sorry for being a bottle neck on this. Have been clogged these two weeks with work. Will go through asap.

jabranham:
No problem at all :-)

vspinu:
I have never given serious thoughts on the derived-mode design, but are we completely sure that prog-mode -> ess-mode -> ess-xyz-mode is the design we want to go with? My main concern is that it creates hard dependency and if we want to split r-mode from ESS eventually it will be much harder this way. 

A more forward looking design might be to implement ess as a minor mode which would overlay on top of existing modes even if they are not part of ESS ecosystem (julia-mode.el for example). MInor modes have their disadvantages though. 

At this stage there is no turn around on this patch for sure, but I still wonder what are your thoughts about this.

lisp/ess-gretl.el
```diff
@@ -545,31 +516,31 @@ been created using the variable `ess-r-versions'."
   :type 'string)
 
 ;;;###autoload
-(defun gretl-mode  (&optional proc-name)
+(define-derived-mode gretl-mode ess-mode "ESS[gretl]"
```
vspinu: This most surely should be ess-gretl-mode.

lisp/ess-help.el
```diff
@@ -296,7 +308,7 @@ if necessary.  It is bound to RET and C-m in R-index pages."
             ess-local-process-name pname)
       (setq buffer-read-only nil)
       (delete-region (point-min) (point-max))
-      (ess-help-mode)
+      (ess--set-help-major-mode)
```
vspinu: Is this ess--set-help-major-mode extra step necessary? It was clear before that ess-help-mode is installed. 

If it's really needed then a better name could be `ess--install-help-mode`.

lisp/ess-inf.el
```diff
@@ -53,15 +52,47 @@
...
 (defvar proc)
 (defvar start)
 (defvar end)
 
+(defvar inferior-ess-mode-syntax-table
+  (let ((tab (copy-syntax-table comint-mode-syntax-table)))
+    tab)
+  "Syntax table for `inferior-ess-mode'.")
+
+(defun ess--set-inferior-major-mode (dialect)
```
vspinu: Same here, "install" would ring better in my ears.

lisp/ess-r-mode.el
```diff
@@ -355,11 +348,9 @@ To be used as part of `font-lock-defaults' keywords."
      (ess-make-source-refd-command-function . #'ess-r-make-source-refd-command)
      (ess-format-eval-message-function      . #'ess-r-format-eval-message)
      (ess-install-library-function          . #'ess-r-install-library)
-     (ess-eldoc-function                    . #'ess-r-eldoc-function)
```
vspinu: Why these variables are removed? There are two main functions of customize alist. First is that we could move move variables across buffers. Second is that we avoid repetition when installing modes. I think both functions are not entirely obliterated with the derived mode system.

lisp/ess-r-mode.el
```diff
@@ -355,11 +348,9 @@ To be used as part of `font-lock-defaults' keywords."
...
      (ess-help-web-search-command           . #'ess-r-sos)
      (ess-build-help-command-function       . #'ess-r-build-help-command)
      (ess-dump-filename-template            . ess-r-dump-filename-template)
-     (ess-mode-editing-alist                . ess-r-editing-alist)
```
vspinu: So editing-alist is done with? Merged into custom alist?

lisp/ess-r-mode.el
```diff
@@ -2203,6 +2202,71 @@ state.")
 
 
 
+;;;*;;; Inferior R mode
+
+(defvar inferior-ess-r-mode-map
```
vspinu: I don't think this is necessary with the derived mode map. This map is automatically created and is derived from inferior-ess-map.

lisp/ess-r-mode.el
```diff
@@ -2203,6 +2202,71 @@ state.")
...
+  (let ((map (make-sparse-keymap)))
+    (define-key map "\M-\r" 'ess-dirs)
+    map)
+  "Keymap for `inferior-ess-r-mode'.")
+
+;; TOTHINK: Prevent string delimiting characters from messing up output in the
+;; inferior buffer
+(defvar inferior-ess-r-mode-syntax-table
```
vspinu: I forgot how syntax tables work with derived modes. Is there a notion of derived syntax table? Don't we have to change our strategy of installing syntax tables now?

lisp/ess-r-mode.el
```diff
@@ -2203,6 +2202,71 @@ state.")
...
+  (ess--maybe-setup-company ess-r-company-backends t)
+  (setq comint-get-old-input #'inferior-ess-get-old-input)
+  (add-hook 'comint-input-filter-functions 'ess-search-path-tracker nil 'local))
+
+

+
+;;;*;;; R Help mode
+
+(defvar ess-r-help-mode-map
```
vspinu: Idem.

lisp/ess-tracebug.el
```diff
@@ -122,8 +122,6 @@ Indicates that ess-tracebug-mode is turned on.")
 
 (defvar ess-watch-mode-map
   (let ((map (make-sparse-keymap)))
-    (set-keymap-parent map special-mode-map)
-    (define-key map "?" 'ess-watch-help)
```
vspinu: Why ? was removed?

lionel-:
I've been thinking about making ESS a minor-mode as well. This seems to be the proper way with independent prog-modes like julia-mode, and I agree ess-r-mode should split at some point. In the end the main point of ESS is to interface with the REPL in a standard way. We could also think about implementing it for python and IELM (emacs-lisp repl), that might give a better sense of what the minor mode would (or should) be about.

lisp/ess-help.el
jabranham: I /really/ don't like "install" here. If you don't like set, maybe one of these instead?: 

- switch
- pick
- choose

alternatively, we could just eliminate that word and do: `ess--help-major-mode`.

lisp/ess-r-mode.el
jabranham: I can't tell which variables you're talking about exactly. `ess-eldoc-function`? It's now set in the appropriate major modes. For R that's ess-r-mode and inferior-ess-r-mode. We repeat ourselves a bit but I think the code is much simpler: (If we're trying to separate ess-r-mode from ess-mode this will also have to be done at some point) 

this is what we remove:

```elisp
    ;; eldoc)
    (require 'eldoc)
    (when (and ess-eldoc-function ;; if mode provide this, it suports eldoc
               (or (and (not inferior) ess-use-eldoc)
                   (and inferior (eq ess-use-eldoc t))))
      (when (> eldoc-idle-delay 0.4) ;; default is too slow for paren help
        (set (make-local-variable 'eldoc-idle-delay) 0.1))
      (set (make-local-variable 'eldoc-documentation-function) ess-eldoc-function)
      (eldoc-mode 1))
``` 

in favor of the much simpler:

```elisp
  ;; eldoc
  (add-function :before-until (local 'eldoc-documentation-function)
                #'ess-r-eldoc-function)
  (when ess-use-eldoc (eldoc-mode))
```

lisp/ess-r-mode.el
jabranham: Yes, editing-alist is gone. The variables editing-alist set are now mostly (all?) set from their major modes.

lisp/ess-r-mode.el
jabranham: Right, but if we want R specific commands in the inferior we need this. Right now that's just `ess-dirs`, which used to get set by `inferior-ess-mode` with this:

```
  (when (string= ess-language "S") ;; todo: what is this doing here?
    (local-set-key "\M-\r"    'ess-dirs))
```

lisp/ess-tracebug.el
jabranham: `?` is bound by default in modes derived from special-mode to `describe-mode` which does basically the same thing.

lisp/ess-r-mode.el
jabranham: > derived syntax table

Not really. You can still do `copy-syntax-table` to sort-of derive from a specific syntax table. 

> Don't we have to change our strategy of installing syntax tables now?

Yes. We no longer rely on the various -customize-alists (I thought I had removed all these but I see now that I missed a few. Will fix that now). Instead the syntax tables are activated by the major modes.

jabranham:
> but are we completely sure that prog-mode -> ess-mode -> ess-xyz-mode is the design we want to go with?

It has some benefits like being able to add to `ess-mode-hook` that will be run in all the various ess language modes. I do see what you're saying though.

So what would ESS-as-a-minor-mode look like? A set of keybindings (and functions) to evaluate code? That might make sense, though rewriting everything to function like that would entail quite a bit of work I think.

> I think the best is to just merge and start testing and fixing if it's broken

Yes, this is what I think too. I've been using this branch off-and-on for the last few months. I think it's safe to say that (for R at least) nothing is super-horribly broken. That said, I'm sure there are use cases I haven't run into yet where things are broken or at least don't work as expected. And I've got no idea about SAS/stata.

lisp/ess-help.el
vspinu: Let's just do `ess--help-mode`. But again, what's the point of having that extra step?

lisp/ess-r-mode.el
vspinu: I don't think we strictly need this. One can simply do `(define-key 'inferior-ess-r-mode-map ...)` after `inferior-ess-r-mode` has been defined. Or, if I recall correctly, it could be done within the :map argumen in `define-derived-mode`. 

My main reason to worry about this is that `inferior-ess-r-mode-map` is not derived from `inferior-ess-map` if `defvar` like that, but I might be wrong. `define-derived-mode` might be doing some side-effecting here. Even if it does, it's better if things are explicit.

lisp/ess-r-mode.el
vspinu: > I can't tell which variables you're talking about exactly. 

Everything that you have removed from `ess-r-customize-alist` -  `ess-ac-sources`, `ess-company-backends` 

> We repeat ourselves a bit but I think the code is much simpler:

Ok, looks like some variables are now bypassing ess-customize. This is inconsistent behavior. Either all or none.

>  ;; eldoc
  (add-function :before-until (local 'eldoc-documentation-function)

Why do we need to advice eldoc-documentation-function if we can just install our own eldoc-documetnation-function? 

Do you use advice in other contexts as well?

lisp/ess-r-mode.el
vspinu: > We no longer rely on the various -customize-alists (I thought I had removed all these but I see now that I missed a few. 

Well, this is a very big step which will ruin quite a lot of user configs out there. But most importantly this might not be the right approach. As I said earlier, one important function of that customize-alist is to clone modes to external buffers like help,  preview etc. which are generic across dialects and for which derivation is not an option. I don't think we ever discuss this change.

lisp/ess-help.el
```diff
@@ -56,6 +60,14 @@
 ;;;; * The major mode ess-help-mode
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(defun ess--set-help-major-mode (&optional dialect)
+  "Determine which help major mode to call, and call it.
+DIALECT defaults to `ess-dialect'."
+  (let ((ess-dialect (or dialect ess-dialect)))
+    (cond ((string= ess-dialect "R") (require 'ess-r-mode) (ess-r-help-mode))
```
vspinu: After this one is merged we should refactor such conds into cl-generics dispatching on ess-dialect. Coresponding functionality of cl-generic was introduced in e25.

lisp/ess-r-mode.el
lionel-: Do you think many people change the syntax table of alist variables via their user config? Or is it something else?

lisp/ess-r-mode.el
jabranham: We had to bypass some of the -customize-alist stuff because of the timing of when variables get set. If we've got `ess-r-mode` turned on in .R files from auto-mode-alist, the first thing it does is call `(ess-r-mode)` (well, technically not the *first*, but close enough). Using define-derived-mode means that they syntax table is set immediately from ess-r-mode-syntax-table, which bypasses the customize alists since they aren't set yet. We could of course go back and reset it but when I tried that I ran into problems (I forget exactly what). It seems simpler and more consistent with what other major modes do to set this up properly in the first place.

As complicated as this PR is overall, the resulting code is /less/ complicated, I think. It's at least shorter --- about  300 lines shorter.

lisp/ess-inf.el
```diff
@@ -214,23 +233,11 @@ This may be useful for debugging."
...
-            (inferior-ess-mode)
+            (ess--set-inferior-major-mode ess-dialect)
             (ess-write-to-dribble-buffer
              (format "(inf-ess 3.0): prog=%s, start-args=%s, echoes=%s\n"
                      inferior-ess-program infargs comint-process-echoes))
             (setq ess-local-process-name procname)
-            (goto-char (point-max))
-
-            (when ess-history-file
-              ;; Load past history
```
vspinu: Why this history-file treatment was removed?

lisp/ess-help.el
jabranham: OK, I'll change it. 

The point is because we have ess-help-mode and also ess-r-help-mode (and ess-stata-help-mode but it's the same as ess-help-mode at the moment). We have to decide which help mode to use and this was the simplest way I could think of when I wrote it. Do you have a different idea?

lisp/ess-inf.el
vspinu: Nevermind. You moved it to ess-inferior-mode. I think it's a good idea.

jabranham: It's not gone - it's in `inferior-ess-mode` now. I tried to make when variables are set a little more standard (it's still not perfect) by setting as many as possible in the major mode setup. It makes more sense (to me) to have it in inferior-ess-mode rather than inferior-ess since it has nothing to do with the actual process (i.e. we never query the process or send anything to it).

lisp/ess-r-mode.el
vspinu: > add-function is a documented way of modifying -function variables.

Hmm. This is something new to me :| 

> Isn't it ok to do it progressively? 

I don't think this could work, but I cannot judge the impact of these changes. 

Let's define the goal. Do we want to get rid of the customize alist in the first place?

Good news is that we don't use customize alist for porting variables from one buffer to another that often. One has to go through all those `ess-setq-vars-local` and see case by case if it could be replaced with something else. If variable porting is no longer a concern it all comes to whether we want to copy paste same setq code across different non-derived modes.

lisp/ess-help.el
jabranham: Yes. Depending on how attached you are, we could actually port from ess-generics and use cl-generics instead. I've played around a bit with porting the functions we have right now and it's not terribly difficult.

lisp/ess-r-mode.el
jabranham: > Do we want to get rid of the customize alist in the first place?

I'd prefer to, eventually. I find them a little weird/confusing. We've got ess-customize-alist which isn't buffer local and gets changed and ess-local-customize-alist which is buffer-local. And then another customize-alist for each language. It's hard to keep track of what's happening when.

Most of the ess-setq-vars-local calls can be replaced with just calling the appropriate major mode, I think.

That said, I don't think we have to get rid of the customize-alists in this PR.

lisp/ess-utils.el
```diff
@@ -584,49 +590,35 @@ See also `ess-use-ido'."
...
-                   (and inferior (eq ess-use-eldoc t))))
-      (when (> eldoc-idle-delay 0.4) ;; default is too slow for paren help
-        (set (make-local-variable 'eldoc-idle-delay) 0.1))
-      (set (make-local-variable 'eldoc-documentation-function) ess-eldoc-function)
-      (eldoc-mode 1))
-
-    ;; tracebug
-    (when (and ess-use-tracebug inferior isR (fboundp 'ess-tracebug))
-      (ess-tracebug 1))))
+(defun ess--maybe-setup-auto-complete (sources &optional inferior)
```
vspinu: Can we please get rid of "maybe"? "Setup" can mean "do nothing" if the settings is so.

jabranham:
@vspinu thanks for the review! Can you let me know there's any reason I shouldn't change the test I mentioned here: https://github.com/emacs-ess/ESS/pull/616#issuecomment-427883931 it's the only failing test and it would be nice not to get an email every time I push to this branch :-)

lisp/ess-utils.el
jabranham: Sure, I'll go back and change it. I wrote this before I knew of your hatred of that word :-)

lisp/ess-help.el
vspinu: Ok. I found the definition of that function :) I still think "install" is the best mnemonic but never-mind. That kind of dispatch we should get rid anyhow in a short run. 

>  Do you have a different idea?

Yes. We could use cl-generic and dispatch on ess-dialect, but that's orthogonal to this PR.

jabranham: When I wrote this I was worried that we'd get to a place where we'd need to set the proper help major mode before ess-dialect was set but that is apparently not the case so should be easy enough to do after this gets merged.

lisp/ess-r-mode.el
vspinu: Ok. I have the same feeling. It's better to use consistent and standard tooling, even if one has to write a few more lines of setqs here and there. We can also resort to some simpler tricks than ess-customize-alist for variable porting and for reducing copy-paste burden. 

So let's aim at removing customize alist by next release and also sticking to only standard inheritance mechanisms (ess-derived-mode and cl-generic). I would also like to remove ess-defgeneric. The purpose of the internal :overwrite could be achieved by an extra cl-generic per function, and it's not really a big deal given the extra benefits of the standard tooling like xref which works with it.

lisp/ess-help.el
vspinu: Looks like there is some telepathy going on ;). Yes, no attachment.  cl-generic it is.

lisp/ess-r-mode.el
vspinu: Not syntax table, but other variables for sure. I don't know why that variable was called `-customize-` alist and if it was ever intended to be `customized` or was documented as being `customizable`. To me it was always part of internals and I don't think non-lisp-aware people ever touched it. So I think if we break and remove it in the next release those guys who do will find a quick way to fix their configs.

lisp/ess-r-mode.el
jabranham: OK, great, sounds like a plan to me. I'll open some issues so we don't lose track of what we talked about here. 

As for this PR, I *think* I've responded to everything, let me know if I missed something.

vspinu:
>  shouldn't change the test I mentioned here: #616 (comment) 

Arh,  sorry. It slipped my mind (again). 

I don't recall why there is that new line there. If that patch fixes the test then for sure, go for it. That test is flimsy anyhow. It will probably fall for me locally after once I am on the next branch. Will dig into it then.

jabranham:
OK, done. I think we're ready to merge then, right? Any objections @vspinu @lionel- ?

vspinu:
All good on my side. Let's roll the toughest ESS PR ever!!

A proper merge rather than rebase+merge would work better here for easier back-reference in case we want to test stuff.

jabranham:
> A proper merge rather than rebase+merge would work better here for easier back-reference in case we want to test stuff.

Yes, I was planning on doing the "semi-linear" merge we discussed a while ago. Basically rebase off of next (done already) and then merge with a merge commit.

lisp/ess-r-mode.el
lionel-: I agree with everything said here. A main goal for the ESS rewrite is to switch towards standard Emacs tooling and idioms, where feasible.

jabranham:
OK, merged. Thanks for the thorough reviews @vspinu and @lionel-!

PR #617, opened 2018-07-24
Fix many compiler warnings

### LINKED ISSUE #612: Compiling *.el now *breaks* functionality
I don't know for how long... just noticed it now.  As I always automatically compile the lisp/*.el files after `git pull`..

To reproduce, type `make`  in the toplevel ess directory.
Then   `emacs -f R`  and within *R* buffer, use

`C-c C-v`  (which aims opening a help page in a buffer)

---> I get an error.

The `*Messages*` buffer then ends with

>  inferior-ess-mode-hook: Increasing ess-loop-timeout..
> Type C-h m for help on ESS version 17.11
> ess-tracebug mode enabled
> Forming completions for .GlobalEnv...done
> Forming completions for ESSR...done
> Forming completions for package:graphics...done
> Forming completions for package:grDevices...done
......
> Forming completions for Autoloads...done
> Forming completions for package:base...done
> Retrieving RDS aliases...done
> ess-s-get-help-topics-function: Invalid function: with-ess-process-buffer

and indeed,  `with-ess-process-buffer` is not a function but a macro (defined in `ess-inf.el`).

jabranham:
This PR fixes up a lot of the ESS files to fix byte compiler warnings (#506) 

On current `master`, M-x compile in the `lisp` subdir results in 124 warnings (this actually undercounts by quite a bit because it groups some together). 

After this PR, I get 11 (one of which is in julia-mode, which we can't do anything about directly). 

Closes #612

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -908,9 +908,11 @@ See `ess-noweb-mode' and `R-mode' for more help."
 ;;;###autoload
 (add-to-list 'interpreter-mode-alist '("r" . R-mode))
 
-(defun R-fix-T-F (&optional from quietly)
-  "Fix T/F into TRUE and FALSE *cautiously*, i.e. not in comments and strings;
- starting from the current position (point)."
+(defun ess-r-fix-T-F (&optional from quietly)
+  "Change T/F into TRUE and FALSE cautiously.
+Do not change in comments and strings. Start at FROM, which
+defaults to point, and change to end of buffer. When QUIETLY, do
+not issue messages."
   (interactive "d\nP"); point and prefix (C-u)
   (save-excursion
     (goto-char from)
@@ -919,6 +921,8 @@ See `ess-noweb-mode' and `R-mode' for more help."
     (goto-char from)
     (ess-rep-regexp "\\(\\([][=,()]\\|<-\\) *\\)F\\>" "\\1FALSE"
                     'fixcase nil (not quietly))))
+(define-obsolete-function-alias 'R-fix-T-F 'ess-r-fix-T-F
+  "2018-07-25")
 
 (defvar ess--packages-cache nil
   "Cache var to store package names. Used by
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -45,7 +45,10 @@
 (require 'ess-r-completion)
 (require 'ess-r-syntax)
 (require 'ess-r-package)
+(require 'ess-trns)
 (when (>= emacs-major-version 25) (require 'ess-r-xref)) ;; Xref API was added in Emacs 25.1
+;; TODO: Remove when we drop support for Emacs 24:
+(declare-function ess-r-xref-backend "exx-r-xref")
 (when (>= emacs-major-version 26) (require 'ess-r-flymake)) ; Flymake rewrite in Emacs 26
 
 ;; TODO: Refactor so as to not rely on dynamic scoping.  After that
@@ -249,6 +252,13 @@
     table)
   "Syntax table used for completion and help symbol lookup.
 It makes underscores and dots word constituent chars.")
+
+(defvar ess-r-namespaced-load-verbose t
+  "Whether to display information on namespaced loading.
+
+When t, loading a file into a namespaced will output information
+about which objects are exported and which stay hidden in the
+namespace.")
 
 (defun ess-r-font-lock-syntactic-face-function (state)
   (let ((string-end (save-excursion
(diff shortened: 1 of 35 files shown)


lisp/ess-tracebug.el
```diff
@@ -56,11 +56,46 @@
...
   (require 'cl-lib))
+(require 'ess-custom)
 (require 'ess-utils)
 
 (defvar text-scale-mode-amount)
 (autoload 'text-scale-mode              "face-remap" "[autoload]" nil)
 
+;; Silence the byte compiler. This is OK here because this file is
+;; only loaded from ess-inf and has no autoloads.
+;; TODO: This is a LOT. Can we move some of this around?
```
vspinu: I think it should be the other way around. Here `(require 'ess-inf)` and `declare-function` in ess-inf.

lisp/ess-tracebug.el
```diff
@@ -94,6 +155,44 @@ The postfix keys are defined in `ess-tracebug-map':
...
+Elements should be directory names, not file names of directories.
+"
+  :type '(choice (const :tag "Unset" nil)
+                 (repeat :tag "Directory list" (string :tag "Directory")))
+  :group 'ess-debug)
+
+(defvar ess-watch-buffer "*R watch*"
+  "Name of the watch buffer.")
+
+(defcustom ess-watch-height-threshold nil
```
vspinu: Is there a technical reason to move these defcustoms around? Tracebug file contains a bunch of isolated functionality, each section (like "watch") is self contained. It's more modular in my eyes this way.

lisp/ess-tracebug.el
```diff
@@ -2355,50 +2445,12 @@ Arguments IGNORE and NOCONFIRM currently not used."
...
   :group 'ess-debug)
 
-(defvar  ess-watch-current-block-overlay nil
-  "The overlay for currently selected block in the R watch buffer .")
-(make-variable-buffer-local 'ess-watch-current-block-overlay)
-
-(defvar ess-watch-buffer "*R watch*"
-  "Name of the watch buffer.")
-
 (defvar  ess-watch-start-block "@----"  ;; fixme: make defcustom and modify the injected command correspondingly
```
vspinu: Why is this one not moved?

lisp/ess-tracebug.el
jabranham: They were defined after getting referenced so the byte compiler would issue a warning about referencing undefined global variables. My personal preference for elisp files is to have requires, defvars/customs, and defuns in that order, but I can move them further down (but still above their original location) if you want.

lisp/ess-tracebug.el
jabranham: It is - it's on line 166 now. Or have I misunderstood you?

lisp/ess-tracebug.el
jabranham: That makes sense to me, I'll fix it up.

lisp/ess-r-mode.el
```diff
@@ -2185,6 +2184,74 @@ otherwise nil."
              (ess-find-rterm (concat PF "/R/") "bin/x64/Rterm.exe")))))))
 (ess-r-define-runners)
 
+(defun ess-fix-miscellaneous (&optional from verbose)
```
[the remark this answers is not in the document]
jabranham: OK, I moved it back and fixed the byte compiler warning for checking if `ess-r-fix-T-F` is defined.

lisp/ess-tracebug.el
jabranham: OK, done

lisp/ess-tracebug.el
vspinu: I meant that you moved all other defvars around it but didn't move this one. Nevermind.

lisp/ess-tracebug.el
vspinu: There are some differences in this regard betwen emacs 27 and others I think. Ok.

lisp/ess-tracebug.el
vspinu: That's strange. I would guess sit's not much. Maybe it's worth creating a dedicated ess-r-inf and move tracebug there. In any case, we will have a clear picture of what is what once we remove all old files. let's not do any massive rearrangements till next release.

jabranham: OK, so I'll leave this as-is for now.

lisp/ess-tracebug.el
jabranham: Oh, I see. You're talking about `ess-watch-start-block`. I only moved the variables that the byte compiler was complaining about in order to make as few changes as possible. But if you want I can move this up with the others.

vspinu: It's fine. We will have to reorganize that code anyhow at some later stage.

jabranham: OK, I think this is ready to be merged then. Let me know if I missed something.

PR #618, opened 2018-07-25
Search for .lintr files for flymake

### LINKED ISSUE #494: Flymake does not find lintr config files
Currently flymake does not respect `.lintr` config files in the same directory or project root.

jabranham:
Currently we don't tell lintr::lint() about .lintr config files. This
adds functionality to find .lintr files:

1. In the current directory
2. In a project root (currently trying ess package root then
   projectile root)
3. In $HOME

Closes #494 

You'll have to update lintr in order to take advantage of this: https://github.com/jimhester/lintr/pull/338


vspinu:
I think, instead of projectile we should use `(project-roots (project-current))` which returns a list of projects. The author of projectile said he will add support for project.el.

lisp/ess-r-flymake.el
```diff
@@ -62,24 +62,44 @@ each element is passed as argument to 'lintr::with_defaults'."
 
 (defvar-local ess-r--flymake-proc nil)
 
+(defvar-local ess-r--lintr-file nil
+  "Location of the .lintr file for this buffer.")
+
 (defvar ess-r--flymake-def-linter
   (replace-regexp-in-string
    "[\n\t ]+" " "
-   "esslint <- function(str, linters, cache) {
```
vspinu: Shall we move this to etc/ESSR on this occasion?

jabranham:
OK, I pushed a commit removing projectile and adding project.el.

Is there a reason you changed `(eval-when-compile (require 'cl-lib))` to just `(require 'cl-lib)`? We're just using `cl-loop` from it, which is a macro.

lisp/ess-r-flymake.el
jabranham: If you want, sure. It's convenient to have it here, though.

I really don't have a strong preference though.

vspinu:
`eval-when-compile` is not needed on requires. The [emacs doc](http://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html) says `Effectively require is automatically eval-and-compile, the package is loaded both when compiling and executing.`

vspinu:
Arh. Nevermind. That's for eval-and-compile. `eval-when-compile` doesn't harm indeed.

lisp/ess-r-flymake.el
vspinu: I don't think it's that convenient, it's just a string not a real code. We will eventually be making ESSR a package and would be nice to track all the relevant R code there.  Let's keep an uniform convention throughout ESS.

vspinu:
I recalled why I removed it. It's an ugly piece which has only marginal benefits. Most likely some other package (ESS itself?) already loaded `cl-lib` and that require is just a noop. Plus, it's annoying to always think and check if we use macros only.

lisp/ess-r-flymake.el
jabranham: As it is the R process we're spawning for flymake doesn't load ESSR though. Is there anywhere else in ESS that spawns background R processes quickly then exits? 

I don't think we even need to define `esslint`, since flymake will log to the *Flymake-log* buffer if anything goes wrong anyway.

vspinu: I missed this part. Indeed, it's a different beast in here.

lisp/ess-r-flymake.el
```diff
@@ -62,24 +64,45 @@ each element is passed as argument to 'lintr::with_defaults'."
...
+the user's home directory.  Return nil if we couldn't find a .lintr file."
+  (cond (;; current directory
+         (file-readable-p (expand-file-name ".lintr" default-directory))
+         (expand-file-name ".lintr" default-directory))
+        ;; Project root
+        ((and (fboundp 'ess-r-package-project)
+              (ess-r-package-project)) (expand-file-name ".lintr" (cdr (ess-r-package-project))))
+        ((and (fboundp 'project-roots)  ; checking that project-roots is defined can be removed when drop Emacs 24
+              (project-current)
+              (project-roots (project-current)))
```
vspinu: I just discovered that `project-roots` throws on `nil` but `(project-current)` returns nil when three is no project.

@dgutov should this be fixed in project.el?

lisp/ess-r-flymake.el
```diff
@@ -62,24 +64,45 @@ each element is passed as argument to 'lintr::with_defaults'."
...
+         (expand-file-name ".lintr" (car (project-roots (project-current)))))
```
vspinu: It should loop over roots and check if file exists I believe. Apparently there are systems in the wild with multiple roots.

jabranham: That's strange. Anyway, I just pushed a new commit that does that.

vspinu: There is another issue with project-roots. It will not work on our projects of the form `(R-project . "path"). We need to define our own `project-roots` method. 

I am having doubts about project.el :/ There are still too many unpolished edges.

lisp/ess-r-flymake.el
vspinu: @dgutov `project-roots` doesn't work on custom `('foo . "path/to/project")`. I think the built in method should be defined on arbitrary lists rather on `(head vc)` and `(head transient)`. It would enforce uniform convention.

dgutov: Fixed how?

When there's no current project, your code should decide what to do. You can call `(project-current t)`, of course.

vspinu: Just return nil on nil. If `(project-root (project-current))` is idiomatic way to retrieve roots, then `project-root` should handle the obvious possibility that `project-current` will return nil.

lisp/ess-r-flymake.el
vspinu: OOP doesn't automatically mean that I have to copy-paste methods every time I define a new "class". There should be sane default methods.

dgutov: What reason is there to assume that the cdr of a project instance is a list of strings?

jabranham: @vspinu FWIW I don't think that affects this since we check if `ess-r-package-project` returns non-nil before checking project.el stuff.

dgutov: The problem with such a loose-typed approach is that an author of any backend that doesn't conform to this scheme will have a harder time debugging why `project-roots` doesn't work properly.

Right now, the answer to that question is simple: "you haven't defined it for your project".

> Wouldn't (proj-type root1 root2 ...) as project instance be enough for pretty much everything?

Well, there are also "external roots", for one thing.

jabranham:
Rebased off master. @vspinu can this be merged or are we waiting for something?

PR #619, opened 2018-07-25
Support rectangle-mark-mode in ess-eval-region

### LINKED ISSUE #45: evaluate rectangular-block as region
I use ESS in conjunction with [evil](http://gitorious.org/evil).

Sometimes I would like to evaluate a selected rectangular block. Eg. in 

```
funkyFun(
  x = 5
 ,y = 10
 ,c = 25
)
```

I visually select block

```
x = 5
y = 10
c = 25
```

If I then do `ess-eval-region`, I get

```
R> x = 5
R> , y = 10
Error: unexpected ',' in " ,"
...
```

jabranham:
If we're in rectangle-mark-mode, loop over each line of the rectangle
and send it to the inferior process.

Closes #45

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1611,11 +1611,13 @@ they might throw off the debugger."
     (setq end (point))))
 
 (defun ess-eval-region (start end toggle &optional message type)
-  "Send the current region to the inferior ESS process.
-With prefix argument toggle the meaning of `ess-eval-visibly';
-this does not apply when using the S-plus GUI, see
-`ess-dde-send-region'. TYPE is a symbol indicating what type of
-region this is."
+  "Send the region from START to END to the inferior ESS process.
+TOGGLE switches the meaning of `ess-eval-visibly'. If given,
+MESSAGE is `message'ed. TYPE is a symbol indicating what type of
+region this is.
+
+If command `rectangle-mark-mode' is active, send the lines of the
+rectangle separately to the inferior process."
   (interactive "r\nP")
 
   (ess-force-buffer-current "Process to use: ")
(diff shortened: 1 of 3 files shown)


lisp/ess-inf.el
```diff
@@ -1625,18 +1624,27 @@ region this is."
     ;; External applications might call ess-eval-* functions; make it
     ;; easier for them
     (ess-setq-vars-local (symbol-value (ess-get-process-variable 'ess-local-customize-alist))))
-
+  (if (and (bound-and-true-p rectangle-mark-mode)
+           (fboundp 'extract-rectangle-bounds))
```
vspinu: I guess this should be removed with emacs 25 right?

vspinu:
Nice, but it's definitely worth a new entry and documentation in `ess-eval-region` and other use facing `eval` functions.

vspinu:
What's the main use case for this BTW?

lisp/ess-inf.el
jabranham: Actually now that I think about it we can get rid of it now since we're already checking rectangle-mark-mode. Just need to tell the byte compiler it's OK with a `declare-function`. Will update and push soon.

jabranham: Argh, turns out we can't because rectangle-mark-mode can be bound in Emacs 24.5 but extract-rectangle-bounds wasn't added until Emacs 25.1. I added a TODO about this.

jabranham:
> new entry and documentation in ess-eval-region

Done

> What's the main use case for this BTW?

Mainly it's one more bug report we can close I had hopes once of getting the number of open issues down to double digits. :-)

I could see it being useful for playing around with code where you don't want to evaluate some things:

```R
x <- 3; x <- 42
y <- 4
x + y
```

If you put everything but x<-42 in a rectangle, you'll get 7. If you did that with a rectangle before this commit, you'd get 46 even though x<-42 wasn't in the active region (well, at least not the "active region" that users would think of).

jabranham:
> Would this be helpful for evaluating commented code?

Yes, absolutely:

```
## x <- 3
## x + 50
```

with a rectangle it's easy to eval that but before this the easiest way would be would be something like `M-h, M-; C-M-x`

NB that rectangles can  go past the physical end of the line so even if you have

```
## x<-1
## x
```

You can still put the whole thing in a rectangle

vspinu:
This sounds like a good use case. Let's merge then :+1:

Issue #555, opened 2018-06-01
flymake: filename as args for ess-r--flymake-proc and wrong rx for flymake-parse-output


ShuguangSun:
Windows 7
Emacs 27
R: 3.5.0

In the ess-r--flymake-proc of ess-r-flymake, the lintr::lint requires filename, e.g., buffer-file-name, as argument, but not the buffer-string. 

In the ess-r--flymake-parse-output, the "<text>:" of the rx will not be matched, as the outputs from lintr::lint begins as filename:row:col: .... 
It works to replace as (minimal-match (one-or-more not-newline)) ":"

vspinu:
I think you still need the newest github version of lintr. I am a bit surprised it's still not on CRAN.

jabranham:
They are working towards a new CRAN release.

mmaechler:
I see that  `flymake` seems active by default in ESS now?  At least it is on one of my Linux computers, and there it is a pain:  I use a large "emacs-desktop" (which re-loads a few hundred of my recently used emacs buffers (dired and files)).   Many of these are related to R,  and with Emacs 26 (new on that computer) my desktop loading takes ages (not least because it produces a warning for every *.R file that it does not find something like 'flymake-init'...

I'm tending to ask this to be made optional (which it probably is) and off by default but maybe you have discussed this before?  Is "lintr" or similar nowadays "on" by default in Rstudio ?

jabranham:
> I see that flymake seems active by default in ESS now?

Yes it is

> my desktop loading takes ages

Yes, that does sound annoying. Ideally this would be fixed by Emacs upstream. I could see flymake using the new threads in Emacs 25 (or was it 26?) to check things asynchronously. But obviously you're looking for a solution now, so that won't help.

> I'm tending to ask this to be made optional (which it probably is) and off by default but maybe you have discussed this before? Is "lintr" or similar nowadays "on" by default in Rstudio ?

I believe lintr is active in rstudio by default. 

If you don't like the linting, you can disable it with `(setq ess-use-flymake nil)`, which you'd want to do before letting Emacs reload your desktop files. 

I'd prefer to keep flymake on by default, since I think it's easier for people to turn off features they don't like than discover potentially useful features they don't know about. I'll look around to see if there's some way to either speed up spawning the subprocess(es) and/or delaying their start.

lionel-:
[...]

PR #634, opened 2018-08-20
New option for ess-use-flymake to only run with live inferior

jabranham:
Flymake can slow down Emacs if you have many R buffers open. This adds
an option to ess-use-flymake to only run make-process if there's an
alive inferior process associated with the current buffer.

See related discussion in #555


mmaechler:
I'd be happy about this change... but have no clue about flymake... so if someone else has a look and merge it...

vspinu:
This is a good change in the right direction. But let's make the default nil for now and mark this feature experimental in the news. Especially that this will likely be the last old-school ESS version I think it's important to keep surprises to the minimum.

mmaechler:
Well, I may be wrong, but I think that introducing flymake and turn it on at all is all ready a surprise to previous ESS users.   So, turning it on less frequently is less of a surprise, isn't it ?

PR #637, opened 2018-08-23
Simplify ess-inside-{comment|string}-p

jabranham:
Removes ess-within-{comment|string}-p since these are functionally equivalent.

Also redefines ess-inside-{comment|string}-p to be a little simpler and removes the `(interactive)` call.

This is just cleaning up code - it shouldn't have any user visible effects (other than removing the interactive bit, which I assume no one was using?)


lionel-:
This created a naming inconsistency in ess-r-syntax.el. Renamed the other predicates to use the `inside` particle: https://github.com/emacs-ess/ESS/commit/ec1730617f648f085bd257644422c23264cc54b4

PR #638, opened 2018-08-23
Modify how ESSR gets loaded on a remote machine

### LINKED ISSUE #429: Potential security issue in ESSR
Hi!
While using ESS remotely (which worked great), I noticed that ESS instructs the remote server to load and execute a .rda file from a 3rd party server over HTTP (see `etc/ESSR/LOADREMOTE`) Since this file can contain arbitrary code, this could be an issue. Would it be possible to either load the .rda from a more secure address (e.g. https://raw.github...), or even better, just send the R source code over ssh? On my machine it takes  <100 ms to parse, and I think this would increase reliability too, for example in cases where the remote server is behind a firewall.

jabranham:
Instead of downloading ESSR.rda from a website, we'll transfer it from
the local machine to the remote machine. This will probably be faster
anyway if the two computers are on the same network.

Closes #429


jabranham:
I've tested this with my laptop and desktop over ssh on the same network and things work well. It would be nice if other people could test it with different setups.

vspinu:
This should go into 18.09 for sure. That archive fetching was an awful hack. I will play with as soon as I can.

jabranham:
> 18.09 

/me looks at calendar

Are we planning a release in the next ~7 days? There are some misc typos and docstrings I've been meaning to fix and/or make checkdoc happy with.

vspinu:
For us 18.09 means by the end of September and sometimes later :)

jabranham:
Oh, hey, look at that. August is month number 8. Who would've guessed. :100: :1st_place_medal:

vspinu:
I recalled the real reason for the hack - it's ess-remote. The proposed implementation will break with all those common use patterns based on ess-remote (tmux,screen,docker). We need to find a way to transfer rds archive through stdin. Looking at how tramp does it might help.

lisp/ess-r-mode.el
```diff
@@ -1265,44 +1265,58 @@ we flush the cache.")
...
 (defun inferior-ess-r-load-ESSR--local (pkg-dir src-dir)
+  "Load ESSR.
+PKG-DIR is unused, SRC-DIR should be the location of the R
+directory inside ESSR."
   (let ((cmd (format "local({
                           source('%s/.load.R', local=TRUE) #define load.ESSR
                           load.ESSR('%s')
                       })\n"
                      src-dir src-dir)))
-    (ess-write-to-dribble-buffer (format "load-ESSR cmd:\n%s\n" cmd))
```
vspinu: Why did you remove this? Errors during load will not be propagated to the user.

jabranham:
I haven't used ess-remote myself, so mind explaining how this will break that? My understanding is that users call `ess-remote` to activate inferior-ess-mode in a remote buffer. So long as Emacs knows that buffer is remote and that $HOME/.config is writable, this should work, right?

lisp/ess-r-mode.el
jabranham: That was a mistake; I'll add it back in.

vspinu:
Nope. `ess-remote` is in the local buffer. You start a shell, then do `ssh some-host` then do `R`, then `ess-remote`. The idea is that the procedure of starting R could be quite complex and involve intermediary steps like screen or docker initialization. But once you have R running in the emacs shell you call ess-remote and it installs all the ESS functionality given that shell.

jabranham:
Could we maybe load it through `buffer-substring` or something like that? Or is that crazy?

vspinu:
Not sure what you mean by that. Sending the whole buffer through stdin? If so then it didn't work when I wrote that code. Anything longer than 500 lines would be corrupted on the way. I was pretty sure that it was an emacs bug, but didn't investigate. And it might have been fixed in the meanwhile. 

In any case I think we should first compress all files (or just ship binary with ESS) and send binary (possibly in batches) with a fingerprit check on the other side.

jabranham:
Yes, that's what I was thinking. Shame it doesn't work. Possible that it's been fixed though, tramp is pretty actively developed. 

> In any case I think we should first compress all files ... and send binary

What do you mean here? How can we send a binary from the local machine to a remote machine when Emacs doesn't know where the remote buffer is? (or am I still misunderstanding ess-remote?)

vspinu:
Through `stdin` ;)

jabranham:
OK I've looked into this a bit more. It doesn't appear to be possible to reliably send a binary file to a process over stdin. Or, if it is, I can't figure it out.

So if that's not possible, it leaves us with a few options:

1) Keep the current situation
2) Remove `ess-remote` and tell users to use tramp for remote buffers
3) Implement something like the current PR for tramp buffers and keep the current situation for `ess-remote` buffers. 

If it were up to me, I'd pick option number 2. From our perspective, it's easier to maintain. There's also currently no guarantee that wherever the remote buffer is has internet access (we only know that the local machine can connect to it, not that it can access the website that ESSR.rda is at), and if it doesn't we probably hang Emacs currently (though I haven't checked that). Using tramp at least guarantees us that we can connect from the local machine to the remote buffer. 

I'm also _guessing_ that `ess-remote` is not widely used since we don't get many bug reports that mention it and we state in the manual that tramp is the recommended way of connecting to a remote process.

WDYT?

vspinu:
> Or, if it is, I can't figure it out.

Ok, looks like nothing changed in the meanwhile. What issued do you see?


> Keep the current situation

We really  need to do something about it. 


> Remove ess-remote and tell users to use tramp for remote buffers

Cannot be (at least not at the moment). Tramp cannot handle a range of situation which I listed above. 

Long time I was thinking that ess-remote should go, but now I am more inclined to actually fix it and make it work in even more context than it already does (ansi-term for instance). There will be always people running atypical setups. We cannot provision a specialized runner for every possible scenario. 

> Implement something like the current PR for tramp buffers and keep the current situation for ess-remote buffers.

Yeh. Surrey ess-remote should be treated differently unless TRAMP case could be handled similarly. I am pretty sure that sending a binary over stdin will be much faster than dir copy with TRAMP.

> we only know that the local machine can connect to it, not that it can access the website that ESSR.rda is at), 

True, but that's rarely a problem. Close to 100% of remotes have internet connection. On the rare occasions when that's not the case the administrator has to copy that rda manually. 

> I'm also guessing that ess-remote is not widely used since we don't get many bug reports that mention it 

That's because it's a brilliant piece of software ;)

> we state in the manual that tramp is the recommended way of connecting to a remote process.

That's what I wrote and I don't necessarily think that any longer.

jabranham:
> Ok, looks like nothing changed in the meanwhile. What issued do you see?

As far as I can tell, there's no way to transfer a binary file over stdin in elisp.

> We really need to do something about it.

Agreed, just wanted to list all the possibilities :-) 

> Tramp cannot handle a range of situation which I listed above.

Out of the box, perhaps not. But tramp can be extended. For example https://github.com/emacs-pe/docker-tramp.el gives tramp a method for docker. I've never used tmux/screen, but there's probably ways to deal with them as well.

>  I am pretty sure that sending a binary over stdin will be much faster than dir copy with TRAMP.

That's probably true but testing it out on my local network the tramp transfer was so quick I didn't even notice it.

If we want to keep `ess-remote`, I can email emacs-devel about this issue and see if they have any insight.

vspinu:
> testing it out on my local network t

Local networks are irrelevant in this case. It's a few hundreds kbs and even on remotes it should be bearable these days, but why bother if one can do it directly, much faster and considerably simplify the code?

> I can email emacs-devel about this issue and see if they have any insight.

Not now. We need to have a reproducible example. I will give it a stab after the release or likely sooner.

For the 18.09 I will just place that rda on github somewhere so that it won't get stuck forever on my site.

> As far as I can tell, there's no way to transfer a binary file over stdin in elisp.

:) You are funny. Did you miss `process-send-string` ?

So here is the algorithm. If you are so eager to work on it (which is very good!):

  - keep an rda archive as part of the repo (it's 12k only)
 - durin initialization split the content of the rda into 250 chars long batches) 
 - send to R in the form list(string = "..batch-content..", hash = "..hash-of-batch..")
 - verify on R side that hash matches the string and return an "..ok.." marker

The hash should probably be md5 because both emacs and core R can compute that (tools package). 

The reason for the check is the potential corruption which I was seeing in my tests 4 years ago. We might be able to safely send the whole rda in one blow. But the hash check should be there.

vspinu:
I think we should load ESSR this way in all cases. Then we don't have to think at all about cases when the program is run locally but it's actually a remote session.

jabranham:
> ) You are funny. Did you miss process-send-string ?

Not at all. I mean to say "reliably". The manual says that elisp will also put in random EOF's into the string that's sent if it's long, which I assumed would mess this up. Here's the section I was looking at:

```
   Some operating systems have limited space for buffered input in a
pty.  On these systems, Emacs sends an EOF periodically amidst the other
characters, to force them through.  For most programs, these EOFs do no
harm.
```

vspinu:
How exactly this EOFs are sent? Is it ^D? In any case we can sort that out by replacing such characters or removing them from the input if they are interfering. I doubt there are many such systems out there. Sending 12k of data should go smoothly. Another question is that the pty could interpret non-ascii characters in a wrong way, but we can always re-encode the binary in base64 or something. Remains to be seen.In the worse case we send text directly, or even function by function.

jabranham:
I'm closing this since whatever we do will need to be very different from the approach here.

PR #642, opened 2018-08-25
Simplify font-lock keyword UI

vspinu:
- merge signal-keywords into ess-R-modifiers
  - merge control-flow-keywords into ess keywords
  - make ess-R-font-lock-keywords of (symbol . boolean) type
  - rename -modifyiers -> -modifiers
  - remove ess-f-t-face in favor of ess-constant-face
  - rename ess-r--bare-keywords -> ess-r--non-fn-kwds


lisp/ess-custom.el
```diff
@@ -2742,82 +2731,82 @@ for an example.")
...
 
 (defvar ess-R-fl-keyword:constants
   '(eval . (cons (regexp-opt ess-R-constants 'words) 'ess-constant-face))
-  "Font-lock constants keyword.")
+  "Font-lock constants keyword.
+See `ess-R-constants' for the list of constants.")
 
 (defvar ess-R-fl-keyword:F&T
-  (cons "\\b[FT]\\b" 'ess-f-t-face)
+  '("\\b[FT]\\b" . ess-constant-face)
```
lionel-: 👍

lisp/ess-custom.el
```diff
@@ -2742,82 +2731,82 @@ for an example.")
...
   '("in" "else" "break" "next" "repeat"))
 
-(defvar ess-R-fl-keyword:bare-keywords
-  '(eval . (let ((bare-keywords (delq nil (mapcar (lambda (keyword)
-                                                    (car (member keyword ess-r--bare-keywords)))
-                                                  ess-R-keywords))))
-             (cons (regexp-opt bare-keywords 'words) 'ess-keyword-face)))
-  "Font-lock keywords that do not precede an opening parenthesis.")
+(defvar-local ess-r--keyword-regexp nil)
+(defun ess-r-find-fl-keyword (limit)
```
lionel-: I think this should be a private `ess-r--` function.

vspinu: It should indeed.

lionel-:
With the merge of signal and modifiers keywords I was thinking that this group can be thought of as side effecty functions in the broad sense.

* The original set was about modifiers to the search path (library, attach, etc) or the global env (source).

* The signal variables are about modifiers to control flow (conditions) and I/O (message, warning).

Would it make sense to include `install.packages()` in that list, on the grounds that it has side effects on the user library? Then maybe also `cat()` for its effect on stdout?

lisp/ess-custom.el
lionel-: IIUC, changing the keywords variables now takes effect only once per session rather than each time a new buffer is opened? I think we should aim for an effectful configuration UI everywhere we can. Maybe make `ess-r--keyword-regexp` a buffer-local variable?

lionel-: oh it is buffer-local!

lionel-:
hmm I think we should add `setwd()` to the list. But agreed we shouldn't try to be exhaustive here, just add the most relevant/important functions, i.e. the ones that should raise eyebrows when reading code. So maybe not `cat()`. That leaves options, setwd, and install.packages.


> There is one way to avoid the "salad" if we add more modifiers. Just inherit modifiers face from call-name face and add bold face.

One issue with this is that in the future it'd make sense to have a special face for NSE and data-masking functions like `quote()`, `library()`, `with()`, `subset()`, `transform()`, etc. I find the special faces for macros in lisp modes help a lot. Hopefully R will gain some way of declaring those functions explicitly to help static analysis and allow IDEs to inform users about the special semantics of functions like `ggplot2::aes()`. In that case I think the face should not stand out too much because of the sheer number of NSE functions out there. The bold face would be ideal.

By the way, since we've made the configuration of keywords effectful and we've added quite a few new words to highlight, I think we can expect ESS users to start tweaking those variables more. Should we take this chance to make these variable names more consistent? I sugest:

* `ess-r-keywords-words`
* `ess-r-modifiers-words`
* `ess-r-constants-words`

vspinu:
>  I find the special faces for macros in lisp modes help a lot. 

Agreed but it's unlikely to happen. All functions are essentially macros, so going into the rabbit hole of classifying which one is which is lost battle from the start. I also think that italics is a bit more subtle than bold and would work better for this case. 

> ess-r-keywords-words
> ess-r-modifiers-words
> ess-r-constants-words

I don't like double plural and I don't think `words` is correct. Those are symbols or names and can be camelCase or snake_case so I wouldn't bother. 

 Regarding the ess-r prefix, I guess it would be ok, but that means changing all those keywords, and introducing hundreds of aliases (also changing ess-s-xyz and who knows what). So I would rather delay this till the major refactoring for the next release. 

To be frank after so much time trying to get used to `r` being `R` I am still struggling. It's kinda okeish for `ess-r-` prefix but when R comes in the middle of a symbol like in `company-r-args` it just strikes me as unnatural. In polymode I get a lot of symbols of `poly-html+R-mode` kind and the new naming convention gives me quite some stress. But it might be "just me".

There is also Rd mode which would need to be renamed ess-rd-xyz.

lionel-:
> All functions are essentially macros, so going into the rabbit hole of classifying which one is which is lost battle from the start.

I think it should be declarative, then it's up to package authors to let R and IDEs know about the semantics of their functions.

> Regarding the ess-r prefix, it would be ok, but that means changing all those keywords, and introducing hundreds of aliases (also changing ess-s-xyz and who knows what). So I would rather delay this till the major refactoring for the next release.

It's not only about the downcase prefix it's also about having a common descriptive suffix `-words`. It indicates they are easy-to-modify font-lock variables.

I was only suggesting to rename these 3 variables. There is a large amount of font-lock variables in ess-custom.el but I don't really consider them as part of the UI. The three `-words` variables and `ess-R-font-lock-keywords` are the only ones R users should care about. That's part of why I argued before for hiding the regexps variables, they unnecessarily increase the surface of configuration variables exposed to the user.

vspinu:
> I think it should be declarative, 

Makes sense. Let's reserve italics for those or a subtle shade of `ess-call-face`. Bold is too aggressive for this purpose ending up boldifying the whole buffer. 

> aving a common descriptive suffix -words

Double plural in symbol names is ugly and those are not just "words", but symbols potentially consisting of multiple words. Let's not do it.

vspinu:
I am adding `setwd` and `options` to modifiers, what about `par` and `load`?

lionel-:
> Double plural in symbol names is ugly

This statement could have been toned down a bit.

> and those are not just "words", but symbols potentially consisting of multiple words

They are words in the syntactic sense but I guess that might be too technical for UI variables.

> I am adding setwd and options to modifiers, what about par and load?

Makes sense. Also `rm()` I think.

vspinu:
>> Double plural in symbol names is ugly
> This statement could have been toned down a bit.

This statement might not have been there.

> Also rm() I think.

Indeed. The elephant in the room.

lionel-:
> They are words in the syntactic sense

hmm not really, I think I was confused by our regexp-opt calls:

```lisp
(regexp-opt ess-R-modifiers 'words)
```

vspinu:
I have tried bold ess-call-face and it didn't look good at all. So I went into the least surprise direction and just added `bold` weight to current `ess-modifiers-face`. I am happy with how it looks in my theme but let's try it out for a few weeks and see if people are ok with it.

font-lock-bold

lionel-:
That seems fine for most of the functions in modifiers but I think it's not good that `message()` and `warn()` have bold face and `stop()` does not. Same for `withCallingHandlers()` having bold face but not `tryCatch()`.

vspinu:
It does not have in your theme? What theme is that? Keywords are bold in all themes I have tried, but we could add :bold to just make sure.

lionel-:
I use spacegray-eighties.

> we could add :bold to just make sure.

Yeah, in the trial period at least. Otherwise it's not consistent to have heavier weight for modifiers.

vspinu:
Actually I was wrong. A lot of themes don't have bold keywords. When themes are re-loaded the weight property is not reset which confused me. 

Messing with themes like that is a bad idea. Will have to revert.

PR #647, opened 2018-08-28
update ESS julia package to work with newer versions

nverno:
This updates the ess-julia.jl to work with newer versions of Julia. I tested the
completion/eldoc with versions 0.6, 0.7 and 1.0. I hadn't looked at Julia in a long
time, so, it could probably be written better, but at least it seems to work.

### DIFF
--- a/etc/ess-julia.jl
+++ b/etc/ess-julia.jl
@@ -1,32 +1,45 @@
 module ESS
 
+# These methods have been deprecated / moved
+macro current_module()
+    return VERSION >= v"0.7-" ? :(@__MODULE__) : :(current_module())
+end
+
+function parse(str)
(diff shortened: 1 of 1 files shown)


vspinu:
Great! Thanks a bunch for this. Does help work as well? In the sense that C-c C-d C-d gives you completions and displays the help page?

nverno:
Yea, I just checked for versions 1.0 and 0.6. I don't know about earlier versions since 
things seem to have changed / been moved around quite a bit. It might not be worth 
trying to continue support for the really beta versions.

vspinu:
I don't think that supporting old versions makes much sense. They promoted julia to v1.0 which presumably means that core API is stable. 

Thanks for the effort!

PR #651, opened 2018-08-29
Only load history file if it's readable

### LINKED ISSUE #162: Windows 7: ESS hangs when starting julia on "Cannot read history file c:/.../.juliahistory"
Emacs version:  24.4.1
ESS version:     15.03
Julia version:     0.3.7

As described in issue #27, upon typing "M-x julia RET", emacs prints the message "Cannot read history file c:/.../.juliahistory" and hangs. 

Typing "C-g" unhangs emacs, and then one can type "C-x b _julia_ RET" to switch to the now running Julia buffer. However, that buffer does not display the julia prompt. (there is no cursor, no ">" sign, nothing to indicate when julia has finished running a command, or indeed that julia is actually running. 
screenshot 2015-04-20 10 59 57-c

jabranham:
Closes #162

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -217,16 +217,13 @@ This may be useful for debugging."
 
             (when ess-history-file
               ;; Load past history
-              (if (eq t ess-history-file)
-                  (set (make-local-variable 'ess-history-file)
-                       (concat "." ess-dialect "history"))
-                ;; otherwise must be a string "..."
-                (unless (stringp ess-history-file)
-                  (error "`ess-history-file' must be nil, t, or a string")))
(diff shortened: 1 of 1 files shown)


jabranham:
It would be nice if someone with windows 7/10 who experiences #162 to check to see if this fixes it, but either even if it doesn't this is nicer than what we have now. If ess-history-file doesn't exist or isn't readable, we currently hang Emacs for a while. This will just disable loading the history if ess-history-file isn't readable.

vspinu:
Great!

habamax:
Yep it fixes the history file issue for me.

But introduces the next "hang point":

image

I have to press C-g to "unhang" it.

PR #652, opened 2018-08-29
Don't interpret single quote as delimiter in inferior buffer

lionel-:
Otherwise error messages with phrases like "Can't do this" mess up the output quite a bit.


vspinu:
Fine with me, but those highlighting issue never bothered me personally because they don't spill over prompts anyhow. 

So it's a trade-off between seeing 'strings' as stings and having those occasional messages badly highlighted.

lionel-:
The R deparser prints strings with double-quotes so I think the trade-off weighs in favour of disabling single quotes which are a common character in natural languages.

PR #653, opened 2018-08-29
Fix a range of new byte compiler warnings

jabranham:
Emacs built from git master adds a few more byte compilation checks,
this addresses them in preparation for the ESS 18.09 release:

- Aliases should be defined _before_ their referent. I moved them up,
  except for those that have been obsolete for five years or more. Those got
  removed. (there were only two or three of these)

- The byte compiler now checks whether cl-lib is needed at runtime,
  not just compile time so added where needed

- set-buffer-process-coding-system is now marked for interactive use
  only, so prefer set-process-coding-system

### DIFF
--- a/lisp/ess-trns.el
+++ b/lisp/ess-trns.el
@@ -162,7 +162,7 @@ in the region, leaving only the S commands.  Other keybindings are:
   (setq comint-use-prompt-regexp t)
   (make-local-variable 'comint-prompt-regexp)
   (setq comint-prompt-regexp (concat "^" inferior-ess-prompt))
-  (setq inferior-font-lock-keywords 'inferior-R-font-lock-keywords)
+  (setq inferior-ess-font-lock-keywords 'inferior-R-font-lock-keywords)
   (setq font-lock-defaults '(ess-build-font-lock-keywords
                              nil nil ((?\. . "w") (?\_ . "w") (?' . "."))))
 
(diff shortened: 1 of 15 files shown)


lisp/ess-trns.el
```diff
@@ -162,7 +162,7 @@ in the region, leaving only the S commands.  Other keybindings are:
   (setq comint-use-prompt-regexp t)
   (make-local-variable 'comint-prompt-regexp)
   (setq comint-prompt-regexp (concat "^" inferior-ess-prompt))
-  (setq inferior-font-lock-keywords 'inferior-R-font-lock-keywords)
+  (setq inferior-ess-font-lock-keywords 'inferior-R-font-lock-keywords)
```
jabranham: @vspinu  - why did you change this recently? Won't that mess up font lock for all non-R inferior modes started from transcript buffers?

vspinu: I didn't change anything substantially here (except of introducing that typo which you fixed). See f51df5dcfc9bf Transcript is in R output only and `inferior-ess-font-lock` is buffer local. Or do you have other issues in mind here?

vspinu:
> Emacs built from git master adds a few more byte compilation checks,

Some of those might not stay till release. The referal one is a special pain in the ass because it affect functions only but not variables or other obsoletes.

In any case, sharp eye and good job! Thanks.

lisp/ess-trns.el
vspinu: I don't know, but that was the old behavior and I am reasonably sure almost no one is using transcript even for R, let alone stata and sas.

PR #654, opened 2018-08-29
Fix dribble buffer writing

jabranham:
This fixes ESS creating and writing to the dribble buffer even when
ess-write-to-dribble-buffer was nil.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -3138,8 +3138,7 @@ See also `ess-verbose'."
   :type 'boolean)
 
 (defvar ess-dribble-buffer "*ESS*"
-  "Name of buffer for temporary use for setting default variable values.
-Used for recording status of the program, mainly for debugging.")
+  "Buffer or name of buffer for printing debugging information.")
 
 (defvar ess-customize-alist nil
   "Variable settings to use for proper behavior.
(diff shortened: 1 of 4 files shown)


lionel-:
Thanks!

vspinu:
I would propose that we clean these dribble messages out of ESS after the release. There is no benefit for debugging purposes and it takes code space, output space and speed.

mmaechler:
> I would propose that we clean these dribble messages out
    > of ESS after the release. There is no benefit for
    > debugging purposes and it takes code space, output space
    > and speed.

Possibly.  In the old times, there were not many emacs lisp
debugging facilities AFAIK, and I have often heavily made
use of the information  in that buffer.

Also, most of our users would not know about emacs lisp
debugging but could easily read or send along (parts of) the
dribble buffer.

Martin

vspinu:
> could easily read or send along (parts of) the dribble buffer.

Theoretically that's a good idea. Practically, I don't recall a single instance where the content of dribble buffer helped on the occasions when it was sent over. In order to be practical we have to log every single detail in all our function and as most of the code doesn't do any reporting we end up with partial information anyhow.

I would propose to replace it with automatic tracing. We could have a command `ess-trace-ess` which would trace all ess functions and generate a call graph with a complete report of what happened. Also we could have `ess-print-state` which would print the values of all ess variables whose values differ from the default ones.

PR #662, opened 2018-09-06
More fixes for the byte compiler

jabranham:
- Move inferior-ess-r--adjust-startup-directory to ess-r-mode.
- Declare a variable for Rd-mode
- Move ess-function-arguments to ess-inf since it requires several
  functions from there
- Remove option 'ensure-mode' from ess-replace-regexp-dump-to-src so
  byte compiler doesn't complain about ess-mode being undefined
- Fixes a type in declare-function (not that the byte compiler
  complains about this)

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -567,29 +567,15 @@ This marks the process with a message, at a particular time point."
                  default-directory)))
     (directory-file-name dir)))
 
-;; FIXME: Move all that R stuff elsewhere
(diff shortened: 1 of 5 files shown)


vspinu:
Thanks. But lets' postpone further cleaning till after the release. With such cleanups it's easy to introduce new bugs. We should try to stick to only bug an doc fixes before the release.

jabranham:
Fair enough, but TBH I see byte compiler warnings as bugs. 

I'll hold off on fixing the rest though.

PR #663, opened 2018-09-06
lisp/Makefile: Add ess-autoloads.el target

jabranham:


jabranham:
I added it to the default "all" target too, let me know if there are any questions.

Maybe we should update the installation/loading docs in the manual as well to mention this way...

vspinu:
So what's the impact on the user more concretely. No more needs for `(load ess-site.el)` even with distro distributions or `make install`? Maybe we check it first if it indeed works and then document.

jabranham:
> No more needs for (load ess-site.el) even with distro distributions or make install

Yes, that's the idea, at least in theory. They should be able to do `(load "ess-autoloads.el")`, assuming it's on their load-path. 

It would be great if someone who uses `make install` or packages ESS could test this out.

lisp/Makefile
```diff
@@ -53,8 +54,18 @@ uninstall:
...
+	@printf	"\nGenerating $@\n"
+	$(EMACSBATCH) --eval "(progn\
+	(setq make-backup-files nil)\
+	(setq generated-autoload-file (expand-file-name \"$@\"))\
+	(setq find-file-visit-truename t)\
+	(update-directory-autoloads default-directory))"
+
 clean:
 	rm -f $(ELC)
+	rm -f ess-autoloads.el
```
vspinu: I am working from source and this cleanning of .el files interferes with this pattern. I just removed the cleaning of julia-mode.el for this reason. I wonder if we should not clean ess-autoloads.el here either. 

I am actually thinking about adding an extra target. Either cleanall, or cleanelc. Probably the latter. WDYT?

jabranham: Fine with me. `cleanelc` is probably clearest. If you add that I can rebase this PR.

lisp/Makefile
```diff
@@ -24,16 +24,17 @@ NOWEAVE=noweave
...
 # ESSR_VER =`cat ../etc/ESSR-VERSION`
 ##ESSR_VER = $(shell cat ../etc/ESSR-VERSION)
 
 ### Targets
 
-all: julia-mode.el $(ELC) ess-custom.el
+all: julia-mode.el $(ELC) ess-custom.el ess-autoloads.el
 
-.PHONY: dist
+.PHONY: dist ess-autoloads.el
```
vspinu: Why did you make it a  .PHONY target? It's a real file.

vspinu: It also probably should not depend on ELCs either because els are scanned for cookies.

jabranham: It started out as a phony target and I forgot to change it. I've deleted that and the elcs dependency.

jabranham: Oh, you did that. Nevermind :-)

Issue #670, opened 2018-09-10
[review / help] Debian packaging using elpa framework


eddelbuettel:
I emailed about parts of this off-list with Alex; it had become time to turn Debian's ess package into elpa-ess.  The reasons are mostly internal:  we had a set of scripts to place and byte-compile elisp package source in place, we now do this via the elpa and package framework relying more on "upstream" which is generally the right thing.  I am pretty ignorant of the details for Emacs but this is standard for Debian in as much as getting some infrastructure together for a set of packages.  And Elisp is one such set of packages.  Many now exist as `elpa-*` allowing fluid interplay between our repos and Emacs's.

Now, when doing this I needed to make two simple patches and I was wondering if you would want to look at these / fold them / make changes, or maybe these have to remain local.  The first is more important as without it, the generated path of ..../package-version/  lacked the version:

```diff
--- ess-17.11.orig/lisp/ess-pkg.el
+++ ess-17.11/lisp/ess-pkg.el
@@ -1,2 +1,2 @@
-(define-package "ess" "0" "Emacs Speaks Statistics" '((julia-mode "0.3"))
+(define-package "ess" "17.11" "Emacs Speaks Statistics" nil
   :url "http://ess.r-project.org" :keywords nil)
```

The second may be due to me not (then) knowing about `(package-initialize)`.  Should this be needed?

```diff
--- ess-17.11.orig/lisp/ess-bugs-d.el
+++ ess-17.11/lisp/ess-bugs-d.el
@@ -23,6 +23,7 @@

 ;;; Code:

+(require 'ess-custom)
 (require 'ess-bugs-l)
 (require 'ess-utils)
 (require 'ess-inf)
```

I don't read all you post here (it is a very active repo) but I noticed a lot of autoload changes.  Should this be better now?

jabranham:
As for the first, `ess-pkg.el` won't exist for the upcoming 18.09 release. Does Debian rely on this file somehow?

We've already added ess-custom requirement to `ess-bugs-d.el`. 

Thanks for the feedback!

eddelbuettel:
The issue was not so much `ess-pkg,el` but the `"0"` for the version.   AFAICT we'd need the current version set somewhere so that `${PACKAGE}-${VERSION}` path can be built.   That may be better now in (M)ELPA current; I only (re-)used the older 17.11 tarball.

jabranham:
Does the contents of the file `./VERSION` work?

eddelbuettel:
[...]

PR #671, opened 2018-09-10
no longer need debian/ with the salsa.debian.org git repo

eddelbuettel:
Per discussion in #670


vspinu:
I think we should remove texlive dependency. It's a crazy overhead to be able to compile pdf docs only. The debian/base-doc doesn't even list pdfs. If someone really, really wants pdf docs we could consider shipping them as part of the repo.

vspinu:
After spending a few hours on this it's pretty clear to me that there is no way we can split it for now. There have been too many structural changes to our files (docs, lisp, copyright, authors etc) which conflict with the hard-coded values in the debian config files. There is no way we can ask you to figure those things out each time we change them upstream. We need an automatic way of propagating such changes.

So I propose that we keep the debian as part of the repo for a few more years till the new-ESS settles down. I will create a branch with minimal elpa-ess debian config based on what I have read so far and we will work from there till we make it right.

vspinu:
> Maybe I misunderstand, and if so, maybe you can explain your intent differently.

Indeed. By streamlining `dh_elpa` I meant that we streamline the transition from the old `ess` -> `elpa-ess`.

vspinu:
> And building manuals is very, very common for Debian packages so I am not sure it is a concern here.

I might be missing the point here. Build-Depends are not the dependencies which user has to have to install the package, it's only that you, as a deb maintainer, need those, right? Then it would be perfectly ok.

eddelbuettel:
>  streamline the transition from the old ess -> elpa-ess

I do not think you need to get too involved in deeper details; the package has been in pretty good shape for 15+ years.  Debian packaging tends to happen on the Debian side, independent of the upstream package.  I simply needed to / did initiate the switch to `elpa-ess` as the (internal) id if emacs binaries changed inside Debian (dropping the version ie the 24 or 25 in `emacs24` and `emacs25` when we had multiple) which broke some package -- including ess (but also `auc-tex` as I learned from KH who often emails me when he has a technical issue).

My work would be helped if the package was simply conforming to the normal standards -- which it does as it is `M-x package-install` installable.  Then again, some files or variable appear to be expected / required and that really is matter _for the `dh_elpa` package_ so if you want to get involved there this is the wrong thread.  I am sure you found its salsa.debian.org repo by now.

In the short to medium term we can fudge.  That is what we did for the first 17.11 build as elpa-ess when I got a very helpful assist from David Bremner (who is quite involved there).

eddelbuettel:
> Build-Depends 

They have nothing to do with this repo.   Build-Depends are just that: what you need to _create_ a binary, as opposed to what you must have ("Depends:") when installing a package, or what you may find useful ("Recommends:"; "Suggests:").  

Can we take this discussion elsewhere?  I do not think this is going anywhere useful.

vspinu:
> I do not think you need to get too involved in deeper details; the package has been in pretty good shape for 15+ years. 

I disagree. Stuff gets easily desynchronized. For  example [copyright](https://salsa.debian.org/edd/ess/blob/master/debian/copyright) hasn't been updated since 2010.,ess-pckg.el is no longer there, the debian-autoloads.el will do more harm than help. So my proposal of having those things pulled automatically has some merits. You can of course go by your own and try to figure upstream changes by yourself, but then you should probably not expect to get much help from here. Or we could get involved and have some basic understanding of dh_elpa packaging and keep things in sync for you. 

> My work would be helped if the package was simply conforming to the normal standards 

It does already because it's no different than other packages regarding the metadata. 

In any case, with my best of intentions to help, I have put the stub of [changes](https://github.com/emacs-ess/ESS/commit/d67b6e5496d811d1a98a2b4b425f3291bb41eef4) that I consider necessary on the [debian](https://github.com/emacs-ess/ESS/tree/debian) branch. Please have a look at least for the `debian/copyright` sake.

eddelbuettel:
It was suggested that Debian issue be dealt with at the Debian repo for ess (at salsa.debian.org).  There are pros and cons for splitting, but we decided to split, and proceeded with that move.  

As such, a debian branch is no longer relevant as it is no longer part of upstream.  We can discuss Debian issues over there.

jabranham:
Big +1 from me; there's no reason for ESS to package itself for all the 100's of linux distros.

PR #673, opened 2018-09-11
Fix insert assign

### LINKED ISSUE #659: ess smart assing key no longer works
I have this in my .emacs file. It used to work fine but no longer works. Pressing ";" just inserts ";".

(require 'ess-site)
(require 'ess-rutils)
....
;; smart assign
(setq ess-smart-S-assign-key ";")  ;if ";" needed, press ";" key twice
(ess-toggle-S-assign nil)
(ess-toggle-underscore nil) ;leave underscore key alone!

Thanks.

lionel-:
Closes #660 and closes #659.

*   Fixes this sequence of commands:

    ```lisp
    (setq ess-smart-S-assign-key nil)
    (define-key ";" ess-mode-map #'ess-insert-assign)
    ```

    The flip side is that the smart key cannot be set to nil at any time to be disabled, it has to be set before loading ESS. But since we agreed to deprecate the smart key at the next release, it's important to avoid any interference between the smart key and `ess-insert-assign`.

*   Announces soft-deprecation of the smart key.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -123,12 +123,19 @@ will need to update their customization to
 @code{inferior-ess-R-program}. These variables are treated as risky
 variables.
 
-@item @code{ess-smart-S-assign} changed name to @code{ess-insert-S-assign}.
-By default, it provides similar functionality. Since many users
-dislike this functionality, it is easier to disable; set
-@code{ess-smart-S-assign-key} to nil. The following functions have
-been made obsolete and will be removed in the next release of ESS:
-ess-smart-S-assign, ess-toggle-S-assign, ess-toggle-S-assign-key,
+@item @code{ess-smart-S-assign} was renamed to @code{ess-insert-assign}.
+It provides similar functionality but for any keybinding, not just `_`.
+For instance if you bind it to `;`, repeated invokations cycle through
+between assignment and inserting `;`.
+
+The variable @code{ess-smart-S-assign-key} is now deprecated. If you
+would like to continue using `_` for insterting assign in future
+releases, please bind @code{ess-insert-assign} in @code{ess-mode-map}
+the normal way.
+
+In addition the following functions have been made obsolete and will be
+removed in the next release of ESS: ess-smart-S-assign,
+ess-toggle-S-assign, ess-toggle-S-assign-key,
 ess-disable-smart-S-assign.
 
 @item The option ess-S-assign has been removed.
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -557,12 +557,14 @@ buffer or end chunks etc.")
 (defcustom ess-smart-S-assign-key "_"
   "Key used by `ess-smart-S-assign'.
 Should be nil or a \"simple\" key, in other words no key
-modifiers.
+modifiers. Can only be changed before ESS is loaded.
 
-You may change this to nil at any time. However, if you change it
-to another string, it must be set before ESS is loaded."
+This variable is deprecated and will be removed in the future.
+Please bind `ess-insert-assign' in `ess-mode-map' to your key of
+choice with `define-key' or similar."
   :group 'ess-S
   :type '(choice (const :tag "Nothing" :value nil) string))
+(make-obsolete-variable 'ess-smart-S-assign-key nil "ESS 18.09")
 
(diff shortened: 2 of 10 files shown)


vspinu:
You are really not letting this one go ;) I like the change. 

> The flip side is that the smart key cannot be set to nil at any time to be disabled,

If it's nil in the begging we don't assign it. I think it's indeed consistent and insulation is worthwhile.

doc/newfeat.texi
```diff
@@ -123,12 +123,19 @@ will need to update their customization to
...
-dislike this functionality, it is easier to disable; set
-@code{ess-smart-S-assign-key} to nil. The following functions have
-been made obsolete and will be removed in the next release of ESS:
-ess-smart-S-assign, ess-toggle-S-assign, ess-toggle-S-assign-key,
+@item @code{ess-smart-S-assign} was renamed to @code{ess-insert-assign}.
+It provides similar functionality but for any keybinding, not just `_`.
+For instance if you bind it to `;`, repeated invokations cycle through
+between assignment and inserting `;`.
+
+The variable @code{ess-smart-S-assign-key} is now soft-deprecated. If
```
vspinu: Do we plan to set it to nil or remove in the next release? Then we should say `deprecated`.

lisp/ess-custom.el
```diff
@@ -557,10 +557,11 @@ buffer or end chunks etc.")
 (defcustom ess-smart-S-assign-key "_"
   "Key used by `ess-smart-S-assign'.
 Should be nil or a \"simple\" key, in other words no key
-modifiers.
+modifiers. Can only be changed before ESS is loaded.
 
-You may change this to nil at any time. However, if you change it
-to another string, it must be set before ESS is loaded."
+This variable is soft-deprecated and will be removed in the
```
vspinu: idem. `soft-deprecated` is too _soft_.

jabranham:
Github is still not loading for me, so I guess we'll do this mailing
list style. Which means I haven't seen any edits (if any) made to any
posts, so bear that in mind.

The biggest thing that stands out to me is that (setq
ess-smart-S-assign-key nil) will no longer work to disable the smart
assignment stuff after ESS is loaded (which means the docstring of
ess-insert-assign is wrong).

I agree that "soft-deprecated" doesn't really make sense for elisp code;
just go ahead and mark it as obsolete and mention in newfeat that it'll
be removed next release.

vspinu:
> The biggest thing that stands out to me is that (setq ess-smart-S-assign-key nil) will no longer work to disable the smart

Right. This is not backward compatible. Setting it to nil always deactivated smart assign. 

For the sake of the documentation we have to decide now on the actual deprecation strategy:

  1)  setting the S-assing-key to nil in 19.04 but removing in a few years 
  2)  removing in 19.04

To me preserving maximal compatibility for 18.09 and cleaning all up in 19.04 is the most appealing.

lionel-:
Perhaps we should bind to `_` a variant of `ess-insert-assign` that self-inserts if smart key is nil. This way we can prepare the transition by allowing `(define-key ";" ess-mode-map #'ess-insert-assign)` without breaking compatiblity.

Agreed to removing the key in the next release.

jabranham:
I guess the only downside of removing the variable is that it'll be harder for users who want that functionality to discover this function and bind it themselves. 

That said, since I don't think most people like the smart-assign behavior, I don't think that's a huge deal.

It's also worth pointing out that this doesn't actually fix the original reason for this whole rewrite since #659  (the issue that kicked off this recent discussion) was setting the variable /after/ ESS was loaded :-)

lionel-:
Good point. I just tried cycle assign, looks nice! But why is it `-assign` for insert and `-assignment` for cycle? These two functions were just added, we should standardise. Should it be `ess-cycle-assign`?

jabranham:
Good question. I really should know the answer since I wrote it but I've got no clue.

I guess we can standardize now. Do you want to simply rename or alias so that we don't break people's config who're already using the new function from e.g. MELPA?

lionel-:
hmm I think we should just rename. I would guess not many users have discovered that function and even less have binded it to a non-default key. And if they did it means they closely follow the development and so they'll be able to figure out what happened.

jabranham:
OK, done

PR #676, opened 2018-09-11
Test major mode

jabranham:
Adds some tests to make sure we don't error when invoking a major mode & fixes these travis failures by requiring bugs/jags in ess-site


jabranham:
Picked to master

Issue #672, opened 2018-09-11
`M-x R-3.4.3` no longer creates a buffer named `*R-3.4.3*`


mmaechler:
This happens only for a couple of days (upto 10?) .. but is a PITA for me.
Yes, this was a feature that not so many people used, but believe me,  R core members have been very happy with it for ca. 20 years ....

Please re-instantiate (the more general case):  `M-x R-3.5.1`  --> buffer named `*R-3.5.1*` etc etc.

jabranham:
Can you include what does happen (an error, message, hangs Emacs, etc)?

Sorry if you've made any edits since the original report; github.com is
down for me (and https://status.github.com/messages suggests I'm not
alone).

mmaechler:
Yes, indeed @vspinu.  It's `*R:2*` or `*R:3*` for me, as of course I'm already running one other R version.... but it is crucial that the buffers are named more usefully and self-explainably.

mmaechler:
yes (of course)  .. I wanted to say.  But now I see the reason is that you *CHANGED* its name, and so my previous setting was no longer used...
Do we really want to frustrate all users who have set the old name to 't' ?

mmaechler:
Thank you, Alex.  The first change (ion May 23) was no problem doe me, at all so the aliasing then worked fine.
The problem for me started when you did your "fix" two weeks ago.  For me it has been the contrary of a fix.  I do have the very regular  
```
(custom-set variables
   ....
   ((long list of a "feeled" 100 lines))
)

where the line that worked till recently was

 '(ess-use-inferior-program-in-buffer-name t)

even more crazily,  `C-h v` that, I'm told that that is  the *correct* variable name and that the version with the extra `-name` in it is deprecated.  But the behavior I now see is the contrary:  The version without the `-name` is no longer working and I need to use the one *with* a `-name` .. which is said to be deprecated ????

((Yes, I'm very very regularly `git pulling` the master branch, and subsequently calling `make` and work with that all day on both of my regularly used (Linux Redhat) environments.

vspinu:
A wild guess here. It might have something to do with the order in which that defalias is declared. It's only very recently that emacs 27 requires the defalias before the declaration. Could it be that it has different effect on older emacses? Would be pretty crazy though, but what Martin describes looks crazy.

jabranham:
@vspinu - No, nothing has changed except the new byte compile warning since it's possible to overwrite user customizations in specific circumstances when you alias stuff.

[...]

PR #678, opened 2018-09-12
Add tests for runners

jabranham:
Please don't merge this yet. I'm not 100% happy with how we setup a fake R executable on travis. It would be nice to be able to run these tests locally somehow but I haven't gotten around to doing that yet. If you've got any ideas, let me know.

-----

Adding tests for runners as a part of #672

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,15 @@ before_install:
   - evm config path /tmp
   - evm install $EVM_EMACS --use --skip
 
+before_script:
+  # link R executable to test defining runners based on other R
+  # versions found on PATH. We're lying here that it is in fact a
+  # different R version but that doesn't matter.
+  - mkdir ess-r
+  - ln -s $(which R) $PWD/ess-r/R-3.2.1
+  - export PATH=$PATH:$PWD/ess-r/
+
+
 script:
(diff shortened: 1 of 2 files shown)


jabranham:
In the spirit of "something is better than nothing," I'm going to go ahead and merge this.

lionel-:
Vaguely related, but eventually it might make sense to install R-devel and rerun all our inferior-based tests on R-devel to catch regressions before they make their way into a release. I'm thinking about the dump-to-buffer regression for instance.

jabranham:
I've looked into that a bit and the biggest hurdle is getting an R-devel
binary that's updated.

lionel-:
We should probably use the same binary as r-travis uses.

jabranham:
I looked into that a while ago but forgot why I didn't like it. :-/

vspinu:
I am not sure if it's worth it. These ess-runners constitute a small piece of functionality and pulling R-devel on travis would increase test times considerably. 

Having said that, devtools has just changed user API in devel and we almost missed that for next release. So having some long term solution for R-side testing would be nice in principle. On the other hand, if a few of us sit on R-devel and devtools-devel then we should be fine as well.

Issue #682, opened 2018-09-14
Add company-R-library to default company backends?


jabranham:
Can we add `company-R-library` to the default company backends?

vspinu:
We probably should, but does it work reliably? I forgot to try it out.

jabranham:
Seems to work well for me. 

It would be nice if inside library statements the packages were listed before other completions. Right now they're at the bottom, which isn't ideal. Is this easy/possible to do? I don't know much about how company determines which match to show first.

vspinu:
In 99.9% of cases users should be fine. In all other cases people will have to disable this backend manually.

vspinu:
@jabranham, the order of backeds is automatically taken care of in grouped backhands as ours with special marker `:separate`.

nverno:
@vspinu I think you could use
`'(company-R-args :with company-R-library company-R-objects :separate)`
to have packages filter to the top?

vspinu:
They are currently filtered on the top after the closing commit. We could give priority to R args though.

mmaechler:
I will be one of  the 0.1% of users (and so will be remaining few colleagues at our institution who still work with Emacs because they also get a `.libPaths()` entry with > 10'000 packages installed....

 is this  -- manual disable -- prominently explained in news.texi and/or other docs that are easily found?

Rstudio has learned in the mean time to start up relatively quickly even withour bombastic .libPaths(), is still slow to deliver a prompt (because it is collecting all the info about all the packages ....).

I really would not want to see this automatic feature..., or then it should only happen in the  background (as e.g., emacs desktop loading)...

nverno:
Yea, maybe this one as more like a `company-yasnippet`, in that it is sometimes useful, but maybe better bound to another command that could be run purposely instead of by default, especially if it causes slowdowns. Also, there is no help docs associated with the packages until they are loaded.

vspinu:
> I will be one of the 0.1% of users 

This is a support feature for a third party package company. If you don't use company then you won't be affected. If you do use, then it would be really helpful if you could assess the slowdown of the company completion inside `library()`. 

I think we should fix this once and for all by bypassing R's `installed.packages`. It's insanely slow because it reads pkg metadata but we don't need that. So I propose that we just list directories in .libPaths and be done with that. As a matter of fact it would be even faster to fetch from the web as we do in `C-c C-e C-i`.

vspinu:
> I think we should fix this once and for all by bypassing R's installed.packages

[...]

PR #683, opened 2018-09-16
add require to library completion prefixes

nverno:
I had forgot to add `require` to the prefixes allowing for 
library completion. So, there will now be completions for packages
in library or require statements.


nverno:
#682

jabranham:
Thanks!

PR #685, opened 2018-09-18
Julia: Remove unused basic-offset variables

### LINKED ISSUE #130: julia-basic-offset should be customizable
In ess-custom.el, julia-basic-offset is defined with

```
(defvar julia-basic-offset 4
  "Offset for julia code editing")
```

This should probably be a customizable variable, similar to how other modes treat their indentation, like so:

```
(defcustom julia-basic-offset 4
  "Offset of a single indentation step in julia code."
  :group 'ess-Julia
  :type '(integer))
```

jabranham:
They don't appear anywhere else in the codebase and aren't in
julia-mode.el either.

Closes #130


vspinu:
Looks right.

Issue #645, opened 2018-08-27
ess-julia-mode autoloads


nverno:
When julia is installed after ess, the autoloads aren't generated. Is there a way around wrapping the entire file in a condition clause?

jabranham:
I also really don't like how ess-julia is handled. I suggest we either:

* Make julia-moded a dependency and just assume it's available or 
* Spin off ess-julia into its own package, which would have ESS as a dependency and live under emacs-ess github organization.

nverno:
I wouldn't mind either of those options. The first seems like it would require much less work though, which is probably a major benefit. Then, in the future if needed it could be separated out?

vspinu:
I completely missed that we don't have julia-mode dependency. Of course we should add that. 

Which autoloads are not generated? For julia-mode? The ess autoloads should all be there.

nverno:
I might have tested incorrectly, let me try from fresh install. Also, I just installed Julia 1.0.0 and there appear to be issues with the ESS julia completion. Anyone else seen those issues?

jabranham:
@vspinu  It's sort-of kind-of a dependency. Installing from MELPA, it's a dependency. But then we guard against this in the beginning of ess-julia (which makes for some strangeness when byte compiling) and it's not mentioned as a dependency anywhere in the docs. And the Makefile actually goes out and downloads julia-mode.el to include in the tarball.

vspinu:
What do you mean by sort-of? I don't see it declared anywhere. It should be declared in ess.el. 

The fetching is for tarball and cannot be removed unless we drop other distribution channels beside ELPA.

@nverno Breakage are quite likely. If julia folks change the internals then things will break. We are entirely relying on users to fix those because none of the core ESS developers is using julia at the moment. See ess-julia.jl in the etc folder. It's usually trivial to fix with even basic knowledge of julia.

nverno:
Ok, I don't get any autoloaded `ess-julia-mode` from a clean install from MELPA even when I install `julia-mode` prior to installing `ess`

jabranham:
It used to be in ess-pkg.el, which got deleted about 2 months ago 36d4e2a319bc61fefe5b58046e66ea6db29dbc4b.

I think we still need ess-pkg.el, it's how multi-file packages should declare themselves `(info "(elisp) Multi-file Packages")`. 

> The fetching is for tarball and cannot be removed unless we drop other distribution channels beside ELPA.

[...]

PR #689, opened 2018-09-22
Do not open jog and jmd files in ess-jags-mode

jabranham:
See discussion in #645


vspinu:
News? We get this in 18.09, right?

jabranham:
> News

Will add once the other PR cleaning up the news file gets merged. 

> We get this in 18.09, right?

Yes, I think this is fine for 18.09. Will merge after adding news.

PR #693, opened 2018-09-23
Improve Makefile

### LINKED ISSUE #146: Installation with el-get fails when style files for documentation not present
Hi, I don't know if that is an issue you want  to care of. If not, just close it. :-)

I tried to install ESS via el-get on an Emacs 24.3. The installaiton just hung somewhere without any notice what happened. After looking into the running processes I saw a "pdflatex refcard" that seemed to hang. I turned out it couldn't find a style file and expected some interaction.

Perhaps it would be an idea to check if all files/software needed to build everything is available or somehow catch those errors.

jabranham:
- Remove "version" target as it fails when git isn't present

- New variables BUILD_PDF_DOCS and BUILD_HTML_DOCS control whether
  pdf/html documentation gets built (off by default), which enables
  "make all" without texlive

Closes #146

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ version:
 	@echo "********************* VERSIONS **************************"
 	@echo $(shell $(EMACS) --version | sed -n 1p)
 	@echo ESS $(ESSVERSION)
-	@echo git HEAD $(shell git rev-parse --short HEAD)
 	@echo "*********************************************************"
 
 .PHONY: lisp
@@ -51,6 +50,8 @@ tarballs: $(ESSDIR)
 	@echo "**********************************************************"
 	@echo "** Making distribution of ESS for (pre)release $(ESSVERSION) from $(ESSDIR)/"
 	@echo "** Creating .tgz file **"
+	cd doc/ ; $(MAKE) pdf
+	cd doc/ ; $(MAKE) html
 	test -f $(ESSDIR).tgz && rm -rf $(ESSDIR).tgz || true
 	$(GNUTAR) hcvofz $(ESSDIR).tgz $(ESSDIR)
 	@echo "** Creating .zip file **"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,28 +6,20 @@ include ./Makeconf
 .PHONY: all install uninstall
 all install uninstall: lisp doc etc
 
-.PHONY: version
```
vspinu: No need to remove the entire version, just the git within it. The most useful part of that vesion is the Emacs version which people are often confused about.

doc/Makefile
```diff
@@ -29,18 +29,26 @@ ESSINFONODE1='* ESS: (ess).'
...
 ESSINFONODE3='                        (S/S+/R, SAS, BUGS, Stata, XLisp-Stat).'
 
-### Targets --
 PDFs = ess.pdf readme.pdf refcard/refcard.pdf
 TXTs = ../README ../ANNOUNCE ../NEWS ../ONEWS
 
-all  : info text html pdf
-docs : info text html
+### Targets --
+ALL_TARGETS = info text
```
vspinu: I think those BUILD_XYZ_DOCS config are not particularly useful? IIt would be simpler to have those html and pdf targets run on travis and have them in the tarball dependencies.

Makefile
jabranham: OK, done

doc/Makefile
jabranham: OK, done

jabranham:
OK to merge now, @vspinu ?

vspinu:
Yes. Thanks!

mmaechler:
I have an academic visitor since Saturday and did not have time to follow you.. sorry..

But I think I am very unhappy about removing the 'version' target. Because when I do the release I definitely want the "full" and correct git version in that string in `lisp/*.el` !!

jabranham:
Isn't that done by this line in the `$(ESSDIR)` target?

```
	cut -c 1-12 $(ESSDIR)-git/.git/refs/heads/master > $(ESSDIR)/etc/git-ref
```

We should probably get rid of this though. The only place it's used is in M-x ess-version, which currently fails in all sorts of ways (e.g. installation from MELPA, if ess is a git submodule, if the directory ess is in is a git directory but not ESS's git directory, etc).

PR #694, opened 2018-09-26
Makefile: Create GPG signatures for tarballs and git tags

### LINKED ISSUE #625: GPG signed releases
ESS should probably provide GPG signatures for releases so that people installing from the tarball can be sure the tarball they received is the one we actually released. Given that many of our users are high-value targets (people with access to all sorts of data --- health info, financial information, etc), it would be a *bad thing* if attackers were to modify ESS since it can run arbitrary code.

As far as I know, there are no downsides to providing GPG signatures with release tarballs.

jabranham:
Closes #625

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,13 @@ tarballs: $(ESSDIR)
 	cd doc/ ; $(MAKE) html
 	test -f $(ESSDIR).tgz && rm -rf $(ESSDIR).tgz || true
 	$(GNUTAR) hcvofz $(ESSDIR).tgz $(ESSDIR)
+	@echo "Signing tgz file"
+	gpg -ba -o $(ESSDIR).tgz.sig $(ESSDIR).tgz
 	@echo "** Creating .zip file **"
 	test -f $(ESSDIR).zip && rm -rf $(ESSDIR).zip || true
 	zip -r $(ESSDIR).zip $(ESSDIR)
+	@echo "Signing zip file"
+	gpg -ba -o $(ESSDIR).zip.sig $(ESSDIR).zip
 	touch $@
 
 # Create the "release" directory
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -122,7 +126,8 @@ ChangeLog: VERSION
 tag:
 	@echo "** Tagging the release **  1) pull existing;  2) tag  3) push it"
 	git pull --tags
-	git tag -a -m'release tagging' v$(ESSVERSION)
+	@echo "Creating tag and signing using gpg"
+	git tag -s -m'release tagging' v$(ESSVERSION)
```
vspinu: Don't we have to push the tags explicitly or the `git push` from below does the job?

jabranham: I believe the git push does that.

jabranham:
@vspinu OK to merge?

mmaechler:
Thank you @jabranham .  I've switched to this branch and tried  `make tarballs`  which ended with an error,  I can also reproduce directly :

       gpg -ba -o ess-18.10.tgz.sig ess-18.10.tgz

which gives
```
gpg: no default secret key: secret key not available
gpg: signing failed: secret key not available
```

jabranham:
Sounds like you don't have gpg set up on that computer correctly. I'm assuming `gpg --list-secret-keys` doesn't return anything?

You can teach gpg about a new secret key with the `--import` option.

mmaechler:
No, there *is* exactly one key listed (the newly created one with 4096 bit) by  `gpg --list-secret-keys`.
But there are many more listed if I list  --list-keys ...
I see I can add    `-u maechler@ethz.ch` to the argument list and then it seems it *would* work
if I remembered the my secret key ;-\(   -- I'll look later today on my home computer I may have "remembered" the key there...

mmaechler:
Yes, of course, --list-keys contains many, but also the one from `--list-secret-keys`

Fedora Linux 28 is quite recent, and the system very well supported by our IT. I've decided a few years ago to go with that and devote my time to other things than OS + system tools maintenance.
Hence I use its gpg, a version 1.4.23 and really don't want to manually maintain a newer version..

mmaechler:
The current master `Makefile` works with current `Makeconf`  and so this branch becomes obsolete.
After all the  comments thread was the important thing here...

PR #695, opened 2018-09-27
Remove unused ess-.*-loop-timeout variables

jabranham:
Does this need a mention in news @vspinu ?


vspinu:
News not needed. Those have not been used for ages.

Issue #413, opened 2017-03-29
ess-roxy-preview-HTML complains about unexpected END_OF_INPUT


johndoe28:
Consider the following code:

##' .. description
##'
##' .. content
##' @title foo
##' @param x 
##' @param y 
##' @param z 
##' @return baz
foo = function(x,y,z = TRUE) {
    baz();
}

If you put point in emacs in the comment and execute the function ess-roxy-preview-HTML, you will get the following error:

foo

Description:

     .. description

Usage:

     foo(x, y, z = TRUE)
     
Arguments:

       x:

       y:

       z:

Details:

     .. content

Value:

     baz

Warning message:
In parse_Rd("c:/Users/xxx/AppData/Local/Temp/RD_13108V6l.Rd",  :
  c:/Users/xxx/AppData/Local/Temp/RD_13108V6l.Rd:20: unexpected END_OF_INPUT '.. content
'
If you execute instead the command ess-roxy-preview-Rd, you will see the generated Rd code is missing a closing brace '}'.

% Generated by roxygen2: do not edit by hand
% Please edit documentation in RtmpMpNQI8/file22343b5692
\name{foo}
\alias{foo}
\title{foo}
\usage{
foo(x, y, z = TRUE)
}
\arguments{
  \item{x}{} \item{y}{} \item{z}{}
}
\value{
baz
}
\description{
.. description
}
\details{
.. content

mmaechler:
I''m also seeing this bug more often than I want.
The bug must be inside  `ess-roxy-preview`  .. I have just pushed a313565f97badfefb289672af45df6311060eef3, 
a slightly nicer formatted version of that, for easier bug fixing.

PR #698, opened 2018-09-30
ess-roxy-preview-Rd: Fix

### LINKED ISSUE #697: missing closing curly bracket in ess-roxy-preview-Rd
A simple example to reproduce this issue is given as follows:

```R
##' Add together two numbers.
##'
##' @param x A number.
##' @param y A number.
##' @return The sum of \code{x} and \code{y}.
##' @examples
##' add(1, 1)
##' add(10, 1)
add <- function(x, y) {
    x + y
}
```
Move the cursor inside of the roxygen documentations, hit `C-c C-o C-r` and I got something like

```
% Generated by roxygen2: do not edit by hand
% Please edit documentation in RtmpaSAqBo/file11865e98c031
\name{add}
\alias{add}
\title{Add together two numbers.}
\usage{
add(x, y)
}
\arguments{
  \item{x}{A number.} \item{y}{A number.}
}
\value{
The sum of \code{x} and \code{y}.
}
\description{
Add together two numbers.
}
\examples{
add(1, 1)
add(10, 1)
```
where the closing curly bracket is missing.

As a result, `ess-roxy-preview-text` (`C-c C-o C-t`) threw a warning complaining `unexpected END_OF_INPUT`.

It would be great if it could be fixed.  Thanks!

jabranham:
- add a newline in the call to cat() so that the prompt gets stripped
  out properly. Test that we caught all the + and > at the beginning
  of the buffer too.

- Reindent let properly and use ess-boolean-command to check for
  roxygen package

- Add tests on travis

Closes #697


jabranham:
Closes #413 as well

jabranham:
cc'ing @lionel-  since this modifies the travis setup. It's now possible to add R packages from the ubuntu PPA. We're using the updated R from there as well as a consequence.

lisp/ess-roxy.el
```diff
@@ -607,15 +606,17 @@ in a temporary buffer and return that buffer."
...
       (with-current-buffer roxy-buf
-        (goto-char 1)
-        (if (search-forward-regexp "FALSE" nil t)
-            (error (concat "Failed to load the " ess-roxy-package " package; "
-                           "in R, try  install.packages(\"" ess-roxy-package "\")"))))
-      (ess-command (concat out-rd-roclet "(\"" tmpf "\")\n") roxy-buf))
+        ;; Kill characters up to % in case we missed stripping prompts
+        ;; or +'s:
+        (goto-char (point-min))
+        (while (not (looking-at-p "%"))
```
vspinu: Why % is special? I think prompts are always on first line before new line and that is cleaned in ess-command.

Also a `(when (re-search-forward "%") (delete-region ...)))` logic would be cleaner and more efficient here I believe.

jabranham: The % is the first character in the roxygen output. Without this we miss a `+` so the preview buffer looks like:

```
+ % <output>
```

I'll change the looping logic later.

PR #699, opened 2018-10-01
Cleanup ess-custom

jabranham:
- Remove several options that we don't use anymore and mark them as
  obsolete: `ess-mode-load-hook', `ess-speedbar-use-p',
  `ess-synchronize-evals', `ess-eval-visibly-at-end',
  `ess-font-lock-mode'

- Remove defining variables which aren't used: `no-doc',
  `comint-highlight-prompt', `ess-source-modes'

- Make inferior-ess-exit-command a defvar rather than defcustom since
  it gets overridden by the language customize alists

### DIFF
--- a/doc/ess.texi
+++ b/doc/ess.texi
@@ -2971,14 +2971,6 @@ and R Transcript Mode buffers.
 @cindex font-lock mode
 @cindex highlighting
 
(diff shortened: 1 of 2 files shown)


jabranham:
Since this doesn't change any behavior, I'll merge later today unless someone objects

PR #706, opened 2018-10-04
lisp/ess-custom.el (ess-highlight-prompt-face): New face

jabranham:


vspinu:
Why do you think we need this change? 

With two faces some prompts will be highlighted in comint face some others in ess face.

jabranham:
Ah, right, I forgot we share that work with comint. This probably doesn't make sense then.

vspinu:
Yeh. It's a bit of a mess but we have to put up with it if we want to build on top of comint.

PR #707, opened 2018-10-04
Remove RPM build

### LINKED ISSUE #306: Problem building RPM from supplied spec
I see the following error when trying to build an RPM from the release tarball:

`error: line 9: Illegal char '-' in: Version: 15.09-2`

jabranham:
RPM and debian packages are built, signed, and distributed by
downstream. Debian build files have already been removed, this removes
the RPM build files.

Closes #306


mmaechler:
Ehmm,  not really!   We have that file in there  provided for people who want/need to make an RPM -- sometimes even outside proper distributions..

Why should we not have that there?
(and ditto for the debian  parts) ?

jabranham:
I'm not sure why ESS should provide build files for specific distros. Does any other project do that?

Downstream is better at packaging for their specific distro. ESS shouldn't be involved in packaging itself for all the various linux package managers.

mmaechler:
I'm not following the advice  _"we are not nice .. because they are not nice either"_   
I come to realize that such thinking becomes prevailing .. notably in the government of the country where GH and its owners reside..  {sorry, it's Friday ;-)}

jabranham:
I'm not sure what github or government has to do with this?

In any case, this has nothing do to with being nice or not nice. The RPM build didn't work for the releases with a hyphen and no one responded when someone asked for help with this. ESS shouldn't be providing an RPM build file if we aren't willing to assist when it errors.

Downstream maintainers will understand their build systems much better than we do anyway.

lionel-:
This makes sense to me, it's more about clear separation of responsibilities than being nice.

jabranham:
> clear separation of responsibilities

Yes, that's a good way of putting it. Thanks.

vspinu:
>> (and ditto for the debian parts) ?

> Ask Dr Spinu. I am not getting involved.

Not sure what you are trying to imply here, but as far as I recall you deleted debian without me agreeing on that (#671) 

> clear separation of responsibilities

Indeed, ideally that should be the case. Unfortunately maintaining a debian repo requires an in-depth understanding of ESS, package.el, emacs, deb packaging and elpa-debian. A daunting task for one person. The elpa-debian packages are maintained in a [big team](https://wiki.debian.org/EmacsenTeam) and I believe that's the right approach in the long run, but who I am to judge.

eddelbuettel:
@vspinu  Feel free to revert #671 if you think that is the better road to take.

You threw one changeset at me -- that I did not ask for -- which deleted *all* customization accrued to the Debain ess package reverting to the cookie-cuttter.  If you think that is better, then you would indeed be well at home in one of those teams.  I happen to differ in my views on that matter.

jabranham:
Now that 18.10-1 is out, the RPM file will again apparently fail: #306 

We should either acknowledge the current situation and leave it to the distros to package ESS for themselves or someone should step up and fix the RPM build files.

Since I know nothing about building RPM files I suggest we just merge this PR.

vspinu:
I am afraid we will probably have to pass on this one. The [comment](https://github.com/emacs-ess/ESS/pull/732#issuecomment-439958230) by @dsteuer does provide a sensible account for why we have to keep RPM in. 

#752 simplifies RPM build and adds a travis test. We should be in a reasonably healthy state after that. Someone with good RPM knowledge has to approve the PR though. @dsteuer would you be so kind to have a look?

dsteuer:
The build process for SuSE packages does not use any rpm information the source provides. It was very useful to me to have the spec file by Tom Moertel as an example to get up and running quickly, back then. Most upstream packages don't provide such an example and that is not really a problem. Therefore I consider it a nice to have, but nothing more. As a packager the complicated part is hunting for dependencies. Only (rpm-)packaged libraries or tools can be used in the build process, or they must be included in the source ball. (For example Julia included tons of stuff in the beginning) These external dependencies tend to have different names for different distributions or have different ideas where site-lisp is found etc, so the spec fils is distribution dependent. Of course one can, better could, package missing libs additionally to ESS itself, but there is only so much time(tm).

To make my point: It is not important to distribute a spec file file with the sources, only a nice to have for new packagers. To build rpms easily it is much more important not to rely on not-packaged-as-rpm external dependencies. You can find i.e. the opensuse spec file there:
 https://build.opensuse.org/package/show/devel:languages:R:released/emacs-ess

Hope that helps

vspinu:
This is really helpful. To summarize our rpm targets are not used (at least in SuSE) and we have to chose between "nice-to-have" for some potential lonely user or remove and provide alternative means (like home-dir install, or package.el friendly tar.gz) for users. Plus, one can easily download rmps from the location you provided, and probably many others. 

No hesitation in my mind that we have to proceed with this PR then.

dickmao:
It's true I've gone back and forth, and back again in #752, switching the title several times to reflect both my ongoing experimentation and the strong admonishments of mmaechler and eddelbuettel.

In past projects I've been on vspinu's side of disruption and mmaechler's side of nondisruption.  My vote for now is to keep RPM.  In the end, I don't think it matter *too* much, and by golly, as mmaechler said, there are so many more interesting things to be done with the elisp.

PR #712, opened 2018-10-12
Remove smart assign key features

jabranham:


doc/newfeat.texi
```diff
@@ -4,6 +4,14 @@
 Changes and New Features in 19.04 (unreleased):
 @itemize @bullet
 
+@item By default, pressing underscore now inserts an underscore.
+Users who liked the previous behavior (i.e. underscore inserting
```
vspinu: I would say loud and clear on the first line "smart-underscore and related functionality (`ess-smart-S-assign-key`, ...),  have been removed". People search for specific variable names when they have problem.

jabranham: Yes, that's better. Done.

PR #716, opened 2018-10-15
Drop Emacs 24 support

jabranham:
I'll leave this for about a day or so in case anyone wants to look it over and then merge into `next`.


lisp/ess-r-mode.el
```diff
@@ -934,22 +929,22 @@ use \"bin/Rterm.exe\"."
...
-  (setq ess--local-handy-commands
-        (append '(("weave"      . ess-swv-weave)
-                  ("tangle"     . ess-swv-tangle))
-                ess-handy-commands)
-        ess-dialect "R"
-        ess-language "S")
-  (put 'ess--local-handy-commands 'permanent-local t)
-  (run-mode-hooks 'Rnw-mode-hook))
+(interactive)
+(require 'ess-noweb);; << probably someplace else
```
vspinu: Looks like something is wrong with indentation here.

jabranham: Indeed. Not sure what happened there but it's fixed now. Thanks.

PR #717, opened 2018-10-15
Use project instead of projectile for inferior buffer names

jabranham:
project.el is included with Emacs 25.1 and newer.

This will fail travis until I merge #716 since it uses some Emacs 25+ lisp. I'll rebase this after merging that, just dropping the PR here so I don't forget.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -17,6 +17,14 @@ Users can change it to 'frame, 'window, or an integer. See the
 docstring for details. @code{ess-auto-width-visible} controls
 visibility.
 
+@item Options for 'ess-gen-proc-buffer-name-function' have been renamed.
+ess-gen-proc-buffer-name:projectile-or-simple was renamed to
+ess-gen-proc-buffer-name:project-or-simple and
+ess-gen-proc-buffer-name:projectile-or-directory was renamed to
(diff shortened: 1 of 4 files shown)


lionel-:
Yes I've been meaning to fix this. I don't think this will be a very disruptive change. All this stuff has been in flux from the start anyway.

lionel-:
By the way we should also define ESS projects for non-packages. They should return either R packages or folders containing something like a `.ess` file (which can contain further config). Then we can start having project-wise config and behaviour, such as setting the startup directory to the current project.

We can't use the full project.el list for this because this would cause R to start in surrounding git repos. This doesn't work well for lots of users who version control their homedirs.

mmaechler:
Hmm, or should this go into master for 18.10 ?  We are officially supporting versions of Emacs 24.x in that "last pre-`next`" release, so we shouldn't use  25++-only functionality?
Or did I completely misunderstand something..

lionel-:
This will go in the `next` branch.

jabranham:
I just rebased off of `next` since #716 was merged and will merge this when the tests pass.

@mmaechler project.el was only added in 25.1 so this can't go into master since we need to support 24 still there.

PR #719, opened 2018-10-16
ess-utils: Enable lexical binding

jabranham:
I'm not sure how familiar you are with lexical-binding but it has some nice benefits over the default dynamic bindings. This is a nice [blog post](https://nullprogram.com/blog/2016/12/11/) that summarizes the advantages as:

- It allows the compiler to catch more mistakes.
- It eliminates a class of bugs related to dynamic scope: Local variables are exposed to manipulation by callees.
- Lexical scope has better performance.

I'd like to eventually turn it on for all ESS files. This gets the ball rolling for using it with `ess-utils.el`.

### DIFF
--- a/lisp/ess-r-completion.el
+++ b/lisp/ess-r-completion.el
@@ -434,7 +434,7 @@ packages are installed."
   (interactive (list 'interactive))
   (cl-case command
     (interactive (company-begin-backend 'company-R-library))
-    (prefix (and (member (car (ess--fn-name-start 'symbol))
+    (prefix (and (member (car (ess--fn-name-start))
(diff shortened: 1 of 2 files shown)


vspinu:
So the idea now to move to lex binding one file by one file? I guess you could do that of PRs as it's non-controversial.  Just let alone  @lionel- indentation code :)

jabranham:
Yes, that's the tentative plan. And no, I'm not touching the indentation stuff.

lionel-:
One Saturday I will wake up and think "time to move the syntax and indentation code to lexical binding".

Issue #720, opened 2018-10-18
Why different mechanism for `ess-eval-line` and `ess-eval-line-and-step`?


ashiklom:
I'm having trouble coming up with a MRE, but I occasionally and unpredictably hit a weird problem where `ess-eval-line` and `ess-eval-line-and-step` seem to be running different R processes, even though they send code to the same buffer. The exact behavior is that objects that I create via `ess-eval-line` are unavailable to code sent via `ess-eval-line-and-step`. E.g. In the code below, if I run `ess-eval-line` on Line 1 and then run `ess-eval-line-and-step` on Line 2, I get `Error: object 'x' not found`. However, if I run Line 2 with `ess-eval-line` instead, it evaluates just fine.

```r
x <- 10     # Line 1
y <- x + 2  # Line 2
```

I looked at the source code for both of these functions and was surprised to see how different they are. I was expecting `ess-eval-line-and-step` to just be something like `(ess-eval-line) (ess-next-code-line)`. However, they seem to use very different mechanisms for sending code to the process buffer. [`ess-eval-line`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-inf.el#L1736-L1745) uses [`ess-eval-region`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-inf.el#L1566-L1590). Meanwhile, [`ess-eval-line-and-step`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-inf.el#L1779-L1798) uses [`ess-eval-linewise`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-inf.el#L1480-L1545), and also includes a call for `ess-force-buffer-current`, which I suspect may be causing the bug I describe above in some edge cases. Why are these two functions so different? Why not use the same mechanism for both functions?

vspinu:
This difference is not good I agree. Do you evaluate inside a package by any chance?

ashiklom:
Thanks for the quick reply! Yes, I'm working inside a package. My development workflow often involves setting the function arguments as variables in a scratch buffer and then running function code line-by-line from within the package code files. If R processes are somehow (re-)assigned based on buffer names part of the way through, then me jumping between buffers in completely different directories (or, for a `*scratch*` buffer, not in a file at all) could be part of the problem.

ashiklom:
Testing more carefully, I can confirm that the problem specifically occurs when evaluating inside a package. To reproduce:

[...]

PR #721, opened 2018-10-18
Simplify ess-eval-* logic

jabranham:
Now all the ess-eval-* functions will have the same behavior.

This is inspired by, but does not fix, bug report #720.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -54,8 +54,6 @@
 (declare-function tramp-dissect-file-name "tramp")
 (declare-function tramp-tramp-file-p "tramp")
 
-(declare-function ess-skip-blanks-backward "ess-r-syntax")
-(declare-function ess-skip-blanks-forward "ess-r-syntax")
 (declare-function ess-mode "ess")
 
 ;; TODO: refactor and remove file-local variable
@@ -1471,9 +1469,9 @@ they might throw off the debugger."
     (skip-chars-backward "\n\t ")
     (setq end (point))))
 
-(defun ess-eval-region (start end toggle &optional message type)
+(defun ess-eval-region (start end vis &optional message type)
   "Send the region from START to END to the inferior ESS process.
-TOGGLE switches the meaning of `ess-eval-visibly'. If given,
+VIS switches the meaning of `ess-eval-visibly'. If given,
 MESSAGE is `message'ed. TYPE is a symbol indicating what type of
 region this is. If command `rectangle-mark-mode' is active, send
 the lines of the rectangle separately to the inferior process."
@@ -1493,15 +1491,15 @@ the lines of the rectangle separately to the inferior process."
       ;; rectangle. Send them separately.
       (let ((reclines (extract-rectangle-bounds (min (mark) (point)) (max (mark) (point)))))
         (mapc (lambda (l)
-                (ess--eval-region (car l) (cdr l) toggle message type))
+                (ess--eval-region (car l) (cdr l) vis message type))
               reclines))
-    (ess--eval-region start end toggle message type)))
+    (ess--eval-region start end vis message type)))
 
-(defun ess--eval-region (start end toggle &optional message type)
+(defun ess--eval-region (start end vis &optional message type)
   "Helper function for `ess-eval-region', which see.
-START, END, TOGGLE, MESSAGE, and TYPE described there."
+START, END, VIS, MESSAGE, and TYPE described there."
   (ess-eval-region--normalise-region)
-  (let ((visibly (if toggle (not ess-eval-visibly) ess-eval-visibly))
+  (let ((visibly (if vis (not ess-eval-visibly) ess-eval-visibly))
         (message (or message "Eval region"))
         (proc (ess-get-process)))
     (save-excursion
@@ -1510,19 +1508,19 @@ START, END, TOGGLE, MESSAGE, and TYPE described there."
     (ess-deactivate-mark))
   (list start end))
 
-(defun ess-eval-buffer (vis)
+(defun ess-eval-buffer (&optional vis)
   "Send the current buffer to the inferior ESS process.
 VIS has same meaning as for `ess-eval-region'."
   (interactive "P")
   (ess-eval-region (point-min) (point-max) vis "Eval buffer" 'buffer))
 
-(defun ess-eval-buffer-from-beg-to-here (vis)
+(defun ess-eval-buffer-from-beg-to-here (&optional vis)
   "Send region from beginning to point to the inferior ESS process.
 VIS has same meaning as for `ess-eval-region'."
   (interactive "P")
   (ess-eval-region (point-min) (point) vis "Eval buffer till point"))
 
-(defun ess-eval-buffer-from-here-to-end (vis)
+(defun ess-eval-buffer-from-here-to-end (&optional vis)
   "Send region from point to end of buffer to the inferior ESS process.
 VIS has same meaning as for `ess-eval-region'."
   (interactive "P")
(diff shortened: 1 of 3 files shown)


jabranham:
This drops our instructions to the compiler for `ess-eval-line` from 25 to 7 :-)

jabranham:
This seems to be working well and the changes are pretty uncontroversial so I'll merge this later today.

vspinu:
There are overlapping changes with master. Things need to be rebated first. I could alos wait till the patch release because there might be other issues.

jabranham:
master got merged into next and this is rebased off next, so should be good, right?

lisp/ess-inf.el
```diff
@@ -1781,14 +1750,11 @@ true.
 Note that when inside a package and namespaced evaluation is in
 place (see `ess-r-set-evaluation-env') evaluation of multiline
 input will fail."
-  (interactive "P\nP"); prefix sets BOTH!
-  (ess-force-buffer-current)
```
vspinu: This overwrites the changes which solved Martin's bug reports.

vspinu: If EVEN-EMPTY is removed, it should be removed from the docs as well.

lisp/ess-inf.el
```diff
@@ -1830,10 +1796,22 @@ far as possible and return -1."
     (goto-char pos)
     n))
 
+(defun ess-step (&optional skip)
+  "Move point to the end of SKIP (if given) and then step.
+Step forward one line or to the next code line depending on
+`ess-eval-empty'. SKIP can be 'function, 'paragraph, or 'line."
+  (cond
```
vspinu: And if it is not given?

lisp/ess-inf.el
```diff
@@ -1830,10 +1796,22 @@ far as possible and return -1."
     (goto-char pos)
     n))
 
+(defun ess-step (&optional skip)
+  "Move point to the end of SKIP (if given) and then step.
```
vspinu: How bout `ess-step-thing` instead? `ess-step` is too generic. Also `ess-skip-thing` sounds a bit more meaningful to my ears.

lisp/ess-inf.el
```diff
@@ -1759,18 +1727,19 @@ input will fail."
   "Send the current line to the inferior ESS process.
 VIS has same meaning as for `ess-eval-region'."
   (interactive "P")
-  (let* ((beg (point-at-bol))
-         (end (point-at-eol))
+  (let* ((beg (line-beginning-position))
+         (end (line-end-position))
```
vspinu: Shall we use `point-at-eol` instead? I find it more consistent with `point`, `point-min`, `point-max`.

lisp/ess-inf.el
jabranham: I chose `step` to be consistent with all the functions that use it (e.g. ess-eval-*-and-step).

How about `ess-skip-and-step`? or `ess-step-maybe-skip`?

lisp/ess-inf.el
jabranham: Then don't skip, just step. I think that's clear from the code but if you want me to add it to the docstring I can.

jabranham: I modified the docstring slightly to make this clearer.

vspinu: Splitting hairs and introducing semantic difference between skip and step for this small context is not worthy IMO. `maybe` is mostly used when the author is lazy enough not to think about better and more concrete name.  The argument `skip` is commonly used as a verb or a boolean variable and is a bit confusing as a noun here. 

I think `(defun ess-skip-thing (&optional thing)` where `thing` is either 'function 'paragraph or 'line captures the idea very clearly. People are used to `thing-at-point` function. The doc could clarify that after the skipping we go to the next code line.

jabranham: Since it really does two things I broke it into two functions `ess-skip-thing` and `ess-step-line`. Since they aren't inferior-specific I also moved them to `ess-utils` along with `ess-next-line` which `ess-step-line` calls.

lisp/ess-utils.el
```diff
@@ -92,12 +92,61 @@ Used in noweb modes.")
 
 ;;*;; elisp tools
 
+(defun ess-next-code-line (&optional arg skip-to-eob)
```
lionel-: Should we namespace these functions in `ess--`? It feels like we should.

jabranham: > It feels like we should.

Why?

lionel-: The question should always go in the opposite direction: why should be export this function? Exporting makes it harder to refactor and change things later.

jabranham: > The question should always go in the opposite direction:

I'm not sure I agree. If we've written useful code, others should be able to use it too if they want.

lionel-: They should, but if we haven't exported the function, they should also be prepared to change their code on update.

I think these functions need some more consideration especially regarding ess-r-syntax and the likely changes coming for this release. We don't have to make them public right away.

jabranham: > but if we haven't exported the function, they should also be prepared to change their code on update.

they're already non-interactive, so I'm guessing not very many people will make direct use of them.

> these functions need some more consideration

Well, you can always say that :-) 

It's worth pointing out that ess-next-code-line isn't a new function in this PR. It just got moved to ess-utils.

lionel-: Oh good, then forget about it.

Most syntax functions are namespaced in `ess-` instead of `ess-r--` because I didn't know better at the time. I'm going to clean that up. I think generic syntax stuff would be very hard to achieve so has to be language-specific.

lionel-: (but the next code line stuff does look generic)

lisp/ess-utils.el
```diff
@@ -92,12 +92,61 @@ Used in noweb modes.")
...
 
+(defun ess-skip-thing (thing)
+  "Leave point at the end of THING.
+THING can be 'function, 'paragraph, or 'line."
+  (cond
+   ((eql thing 'line) (goto-char (line-end-position)))
+   ((eql thing 'paragraph) (forward-paragraph))
+   ((eql thing 'function) (goto-char (cadr (ess-end-of-function nil 'no-error))))))
+
+(defun ess-step-line (&optional skip)
```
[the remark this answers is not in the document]
jabranham: I like the name since it's called from all the `eval-and-step` functions. The naming is consistent with those.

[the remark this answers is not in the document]
jabranham: > That's not a good reason

Well I think it is.

> This function does much more than those -step functions require

It skips something then steps to the next line. I suggested `ess-step-and-skip` or something like that but you didn't like that.

What would you like to name it?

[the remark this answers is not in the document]
vspinu: >  I suggested ess-step-and-skip

It's the other way around `ess-skip-and-step`. You see, even you are confused. Those skip and step are close relatives and we don't want people to deal with the marginal difference in semantics, but I am repeating myself. 

> What would you like to name it?

I would like to merge it into `ess-skip-thing` as proposed earlier.  It would make the code using the function more readable. 

Proliferation of new functions with marginal value is something that we want to avoid in the future. The mess that we are in is mostly because different people (myself including) were adding over the years whatever they thought is useful without consideration for the big picture and overall design.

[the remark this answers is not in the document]
jabranham: I'm not sure how that's any different than the current situation, but sure. How do you want to merge it in? Like this?

```diff
diff --git a/lisp/ess-utils.el b/lisp/ess-utils.el
index 2f480ae5..e0e27c50 100644
--- a/lisp/ess-utils.el
+++ b/lisp/ess-utils.el
@@ -132,18 +132,12 @@ far as possible and return -1."
     (forward-line (1- line))))
 
 (defun ess-skip-thing (thing)
-  "Leave point at the end of THING.
+  "Leave point at the end of THING then go to the \"next\" line.
 THING can be 'function, 'paragraph, or 'line."
   (cond
    ((eql thing 'line) (goto-char (line-end-position)))
    ((eql thing 'paragraph) (forward-paragraph))
-   ((eql thing 'function) (goto-char (cadr (ess-end-of-function nil 'no-error))))))
-
-(defun ess-step-line (&optional skip)
-  "Step forward to the \"next\" line.
-Depending on `ess-eval-empty', step one line or to the next code
-line. When given, SKIP gets passed to `ess-skip-thing'."
-  (when skip (ess-skip-thing skip))
+   ((eql thing 'function) (goto-char (cadr (ess-end-of-function nil 'no-error)))))
   (if (or ess-eval-empty
           (and (fboundp 'ess-roxy-entry-p)
                (ess-roxy-entry-p)))
```

PR #722, opened 2018-10-18
Remove little-to-no used languages

jabranham:
Removes old versions of S+, ARC, gretl, OMG, VST, and XLS since they won't work after #616 (`define-derived-mode`)


vspinu:
So what's left from S?

 Also, gretl is a new tool. i would rather keep it for now. It's true that we didn't commit to it since it was pulled in. I am not even sure it works, but I guess maintenance burden should be minimal.

jabranham:
> So what's left from S?

ess-s-lang.el and ess-sp6-d.el. 

Someone would have to fix gretl-mode to derive from ess-mode. I know nothing about gretl.

vspinu:
Ok. I will have a look once the derived PR is in.  Will brush my econometrics on this occasion :)

jabranham:
OK I will un-delete it from this branch. Once I do that OK to merge?

vspinu:
Sure but you shouldn't be deleting them I guess. Simply move to old dir. 

There are other files which should be moved. Ess-noweb particularly, but I think it should go to obsolete rather than old for now.

vspinu:
What about other antics ess-mouse, ess-gui, make-regexp ...? I have never looked into those, must be really dusty. regexp should be moved on regex-opt.

jabranham:
I was planning on removing them rather than putting them in old since the code won't work. Doesn't make sense to ship broken code. Plus anyone who really wants them can either grab ESS 18.10 or dig through the git history.

As far as noweb and the other oddities, I'm totally fine removing those as well. I was just starting with the languages.

I haven't looked into noweb at all but I imagine the define-derived-mode rewrite will break it as well.

> regexp should be moved on regex-opt.

I don't understand what you mean here.

vspinu:
> the code won't work.

I see. Makes sense then.

> I haven't looked into noweb at all but I imagine the define-derived-mode rewrite will break it as well.

Probably not. It just calls the mode function and doesn't care if it's derived or not. 

> I don't understand what you mean here.

We have make-regexp.el in our repo. I want to get rid of it and replace its use by the build-in regexp-opt package.

jabranham:
OK, I will un-delete gretl and merge this later today or tomorrow.

+1 for removing make-regexp. I haven't looked at all at how similar make-regexp and regexp-opt are, but if it's difficult to transition I think the only file that uses make-regexp is stata mode so we could just ess- namespace the make-regexp stuff and move it into stata mode.

PR #728, opened 2018-10-23
Make ESS respect 'display-buffer-alist'

### LINKED ISSUE #687: ess-switch-to-inferior-or-script-buffer activates frame on other desktop
I use emacs on a linux pc with multiple desktop (most linux desktops have this)

Using emacs server, if I have an emacs frame with an *R* buffer on another desktop, and this happens to be used by an R script buffer that Im working with on my current desktop, then switch, all that happens is that it defocuses my current emacs window and probably focuses that on the other desktop. But I have to go there myself to see it. Which is not at all what I want.

I would want to see that frame in my current window, I don't mind that its viewed in multiple frames on multiple desktops.

Also I use ido which is probably related.

jabranham:
Now users can configure 'display-buffer-alist' to customize where and
how ESS's various buffers are displayed. This also deprecates the
existing ESS-specific framework we have in favor of offloading that
responsibility to Emacs. The docstrings of inferior-ess-same-window and
inferior-ess-own-frame were modified to include instructions on how to
take advantage of 'display-buffer-alist'.

This doesn't touch the `ess-help` display stuff, but we might also think about deprecating those variables (e.g. `ess-help-own-frame` and friends) and relying on `display-buffer` framework.

I think the only user-visible change this has if they haven't customized anything is that M-x R will now open a new window if the frame only has one window, just like we already do if you do e.g. C-c C-c in an R-mode buffer.

Closes #687

### DIFF
--- a/lisp/ess-dde.el
+++ b/lisp/ess-dde.el
@@ -164,7 +164,7 @@ file into an emacs buffer and displays it."
     (sleep-for sleep)
     (if (not bufname)
         (find-file filename)
-      (switch-to-buffer bufname))
+      (pop-to-buffer-same-window bufname))
     (revert-buffer t t) ;; this allows the user to reuse the BUF name
     ))
 
--- a/doc/ess.texi
+++ b/doc/ess.texi
@@ -573,11 +573,28 @@ It is not normally necessary to pass arguments to the iESS program; in
 particular do not pass the @samp{-e} option to @code{Splus}, since ESS
 provides its own command history mechanism.
 
-By default, the new process will be displayed in the same window in the
-current frame.  If you wish your iESS process to appear in a separate
-frame, customize the variable @code{inferior-ess-own-frame}.
-Alternatively, change @code{inferior-ess-same-window} if you wish the
-process to appear within another window of the current frame.
+To control where buffer containing the process is displayed, customize
+@code{display-buffer-alist}, @xref{Window Choice,,, emacs}. For
+example, to ensure Help buffers always appear at the right side of the
+current frame and inferior R buffers always appear at the bottom of
+the current frame, you could add this to your Emacs initialization
+file:
+
+@example
+(add-to-list 'display-buffer-alist
+             '("*Help"
+               (display-buffer-reuse-window display-buffer-in-side-window)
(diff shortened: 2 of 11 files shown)


jabranham:
@lionel- this is what I was referencing this morning. I can't seem to find that conversation now though to link it here.

lisp/ess-dde.el
```diff
@@ -164,7 +164,7 @@ file into an emacs buffer and displays it."
     (sleep-for sleep)
     (if (not bufname)
         (find-file filename)
-      (switch-to-buffer bufname))
+      (pop-to-buffer-same-window bufname))
```
vspinu: What is the semantic difference between these? I cannot figure it out from the docs.

jabranham: The docs are horrible. As far as I can tell, `pop-to-buffer-same-window` is nearly the same as `switch-to-buffer`, but relies on `display-buffer` framework, which `switch-to-buffer` seems to ignore.

jabranham:
One misfeature of this branch in its current state is that if you have `(setq inferior-ess-same-window t)` in your config, open an R script buffer, and do something like C-RET to start the inferior, then the inferior buffer is on the left and the R script is on the right.

I'm not sure how common that situation is though since `t` was the default value. People would have had to explicitly set it to the default value.

jabranham:
`inferior-ess-same-window` controls whether inferior processes start in the current, selected window or a an "other" window. However it only has an effect from M-x R and friends, not from things like C-RET in an R-mode buffer. I could just remove it, I doubt anyone would be terribly unhappy if M-x R pops up a new window with the process. That's at least consistent with the behavior we have already in an R-mode buffer.

I'll think about it some more if it's easy to preserve the current behavior but if not that's probably what I'll do unless someone objects.

vspinu:
Ok, let's remove. If someone complains we will see how to put it back. Now what I understand that it means I am personally quite annoyed by the current full window startup behavior. Making it work consistently with C-c C-z  is a good feature.

jabranham:
Sounds good to me. Pushed a new commit with `inferior-ess-same-window` removed. I'll merge this soon then.

PR #729, opened 2018-10-24
Use defvaralias for variables

zhangpj:
`ess-r-customize-alist` and `ess-r-post-run-hook` are variables, so use `defvaralias` instead.

By the way, thanks to all the maintainers for your awesome work! I cannot image my life without ESS.


jabranham:
Thanks for the contribution.

@vspinu perhaps we should just remove the aliases since this issue went undetected for so long.

vspinu:
Many people don't report even if discover it. We should check for such errors everywhere. 

I agree with removal. Maybe it's better to obsolete in the master and remove in the next though.

jabranham:
OK, done

PR #732, opened 2018-10-26
Obsolete literate data analysis functionality

### LINKED ISSUE #98: ess-noweb-post-command-function may return errors, may be called inappropriately
I've been playing around with `knitr` and `ess-noweb-mode`. I've noticed some problems occasionally when leaving a `*.Rnw` file open. Occasionally the post-command function seems to get called even for modes that aren't running `ess-noweb-mode`. It's hard to notice most of the time, but when opening a file using `emacsclient` from a terminal, it will attempt to call the post-command hook and return an error:

```
Error in post-command-hook (ess-noweb-post-command-function): (wrong-type-argument arrayp nil)
```

This has two unfortunate consequences, the `emacsclient` gets killed, and `ess-noweb-post-command-function` gets removed from the `post-command-hook`, as do all functions that raise an error. The confusing part is that, as far as I understand it, `post-command-hook` should remain buffer-local.

It's hard to consistently reproduce this, but I'm working on a minimal recipe for doing so. In the meantime the workaround I'm using is the following:

```
(defun ess-noweb-post-command-function ()
  "The hook being run after each command in noweb mode."
  (condition-case err
      (ess-noweb-select-mode)
    (error 
     (message "Error: in ess-noweb-post-command-function %s" (error-message-string err)))))
```

### LINKED ISSUE #114: Starting R from a Sweave file (somehow) fails
I am curenlty using ess_14.09-1precise0_all.deb together with emacs23. Suddenly (perhaps after an update?), I have problems when I start an R session from a Sweave file by `M-n s`. An R buffer is opened but I get the following error:

> Debugger entered--Lisp error: (error "No ESS process is associated with this buffer now")

The opened buffer then doesn't auto-load 

```
options(STERM='iESS', str.dendrogram.last="'", editor='emacsclient', 
        show.error.locations=TRUE)
```

as it would do usually when I start R; Thus it doesn't know the emacs specific commands such as auto-complete, help, etc. 

On the other hand, everything works fine if I do not use `M-n s` to start R but if I execute a code line from the Sweave file to start the R process (by using `Shift-Enter` with these scripts: http://www.emacswiki.org/emacs/ESSShiftEnter).

Please let me know if you need more specific information.
Can anyone help me?

### LINKED ISSUE #136: ess-swv-PDF Bug on Windows
In GNU Emacs 24.4.1, in AUCTeX 18.88, in file ess-swv.el, the function ess-swv-PDF format string used to create a value bound to cmdstr-win prevents the use of any but the system default program to open the PDF file.  In particular, it prevents the use of a user-supplied pdfviewer.  To fix this bug, in the body of the ess-swv-PDF function, the line:

(cmdstr-win (format "start \"%s\" \"%s.pdf\"" pdfviewer namestem))

should be changed to:

(cmdstr-win (format "start \"\" \"%s\" \"%s.pdf\"" pdfviewer namestem))

The value of "pdfviewer" should be inserted with surrounding quotes in cmdstr-win in case that value contains spaces.  But if a string surrounded by quotes is the first argument to the Windows shell command "start", it is interpreted by that command as the optional "title" input instead of the optional "command/program" input.  Then "start" uses the system default program to open the PDF file.

An simple workaround is for the user to redefine ess-swv-PDF in their .emacs file after loading the original function by evaluating "(require 'ess-site)".

In GNU Emacs 24.4.1 (i686-pc-mingw32)
 of 2014-10-24 on LEG570
Windowing system distributor `Microsoft Corp.', version 6.1.7601
Configured using:
`configure --prefix=/c/usr'

Important settings:
  value of $LANG: ENU
  locale-coding-system: cp1252

Major mode: Fundamental

Minor modes in effect:
  ess-noweb-font-lock-mode: t
  ess-noweb-mode: t
  TeX-PDF-mode: t
  diff-auto-refine-mode: t
  shell-dirtrack-mode: t
  display-time-mode: t
  show-paren-mode: t
  tooltip-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  blink-cursor-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  line-number-mode: t
  transient-mark-mode: t

### LINKED ISSUE #159: AUCTeX, ESS, Noweb, and "Error running timer `font-latex-jit-lock-force-redisplay': (wrong-number-of-arguments (2 . 2) 3)"
The effect is to generate the error message:
  Error running timer `font-latex-jit-lock-force-redisplay': (wrong-number-of-arguments (2 . 2) 3)
or drop into the debugger with debug-on-error set to t.

Contents of file foo.Rnw:
\documentclass[12pt]{article}
\begin{document}
\end{document}
%%% Local Variables:
%%% eval: (ess-noweb-mode t)
%%% End:

Actions:
Position the cursor anywhere in the file.
Run the command "LaTeX-section".
Type a level or no level at all, and press enter.
Type a title or no title at all, and press enter.

Note: The "Rnw" file type makes the major mode be LaTeX and adds the polymode for Rnw minor mode.  The motivation for turning on the ess-noweb-mode minor mode is to obtain the default bindings for knitr-ing and then pdflatex-ing the document.

### LINKED ISSUE #215: R noweb and helm: ess-noweb-make-variable-permanent-local hook issues for other modes
The combination of R, noweb files, and helm cause problems due to local values of post-command-hook.  A description of the problem can be found in a helm bug I filed here:
https://github.com/emacs-helm/helm/issues/1126.

Emacs  : GNU Emacs 24.5.1 (amd64-portbld-freebsd10.1, X toolkit)
 of 2015-04-20 on 10amd64-default-job-01
Package: ess-mode 15.09-devel [elpa: 20150810.826]

P.S. I originally submitted this using `M-x ess-submit-bug-report` as requested on the ESS-bugs mailing list web page.

### LINKED ISSUE #260: [feature request] inline figure
imaxima can support inline figure:

So I think it's possible for ESS too.

### LINKED ISSUE #338: Make weavers respect TeX-master
The AUCTeX helper functions provided via `ess-swv-add-TeX-commands` and friends respect AUCTeX's master file specifications, but the generic ones implemented via `ess-swv-run-in-R` don't. This is a bit annoying, because the AUCTeX commands start a new process every time, so depending on the warm-up time spent on library loading etc. it is often significantly faster to use the existing R process with most objects already in memory etc. Would it be possible to check if TeX-master is set and respect that if possible? Something along the lines of the redefinition of `rnw-file` in [this commit](https://github.com/anghyflawn/dot-emacs/commit/428ca0fa4f03d06c8c9d74d93624474dd4be2963), I suppose (but presumably less fragile). Thanks!

### LINKED ISSUE #354: Probable infinite loop when uncommenting an environment with Rnw-mode and AUCTeX
Open the following `test.rnw` file with `Rnw-mode`, using AUCTeX's `LaTeX-mode`:

``` latex
\documentclass{article}
\begin{document}

% \begin{equation}
%   1+2=3
% \end{equation}

\end{document}
```

If I try to uncomment the `equation` environment Emacs freezes, apparently because a function enters an infinite loop that you can break with <kbd>C-g</kbd>.  Verified in Emacs 24.5.1.  Doesn't happen if AUCTeX is not used, and doesn't occur in a plain LaTeX file opened with just `LaTeX-mode`, so this may be due to some strange interaction between AUCTeX and ESS.

This has been first reported in AUCTeX's bug tracker: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24079  I wasn't able to find the culprit.  The function that halts is `uncomment-region-default`, but when I `edebug` it, the problem [disappears](https://en.wikipedia.org/wiki/Heisenbug).

### LINKED ISSUE #405: Question: callbacks for ess-swv-run-in-R
I would like to run an Emacs command after `ess-swv-knit` generates a TeX file. How do I use a callback which will be called after `ess-swv-knit` generates a TeX file? See https://github.com/sagemath/sage-shell-mode/issues/31 for the motivation.

### LINKED ISSUE #438: Note:  Variable binding depth exceeds max-specpdl-size
Hi all,
in ESS 16.10 "ess-eval-chunk" failed when a "space" followed the "=" at the beginning of a code chunk "<<>>= ":

generate-new-buffer: Variable binding depth exceeds max-specpdl-size

or

preview-clearout: Variable binding depth exceeds max-specpdl-size

Interestingly, marking the code and calling "ess-eval-region" works as expected. Deleting the space "<<>>=" and calling again "ess-eval-chunk" works.

Maybe this is of interest, Sven

### LINKED ISSUE #487: Rnw-mode "loses" the effect of its file-local variables when exiting R-mode
Note : this is a **duplicate** of an issue posted a few days ago on the `ess-bugs list`, not yet acknowledged nor included in the mailing list archive.

I noticed the same issue when trying to use Rnw as the host mode pf a polymode, and Vitalie Spinu, who couldn't reproduce my problem, suggested to post my findings abot `Rnw-mode` here.

How to reproduce : use the `initMinimal.el` fle as the init filen to edit `Essai0.el` : `emacs -q -l initMinimal.el Essai0` : 

[MinimalExample.zip](https://github.com/emacs-ess/ESS/files/1717133/MinimalExample.zip)

* After a fresh start, the file variable `TeX-master` is set to `"Essai0"`(as seen by `M-: TeX-master`), thus allowing AUCTeX's `C-c C-c` to compile `Essai0.Rnw`. At this point : 
    - the mode indicator of emacs reads `(LaTeX/PS Ref Wrap Noweb NWFL)`, 
    - `M-: file-local-variables-alist` returns `((TeX-master . "Essai0") (TeX-engine . default) (mode . Rnw-mode))`.
* Move after `2+2` in `Firstrchunk`. At this point : 
    - the Emacs mode indicator reads `(ESS[S] [none] ElDoc Noweb NWFL)`, 
    - `M-: TeX-master` returns `t`,
    - `M-: file-local-variables-alist` still returns `((TeX-master . "Essai0") (TeX-engine . default) (mode . Rnw-mode))`.
* Move _outside* the R chunk (e. g. on the empty line below it). At this point : 
    - The Emacs mode line reads `LaTeX/PS Ref Wrap Noweb NWFL)`
    - `M-: TeX-master` still returns `t`,
    - `M-: file-local-variables-alist` still returns `((TeX-master . "Essai0") (TeX-engine . default) (mode . Rnw-mode))`.

In other words, `TeX-master` is always in the file's local variables, but is _ignored_ or _unset_ after passing in an Rnw-mode chunk. Ditto for TeX-engine, BTW.

The consequences are what one can expect : `M-r` correctly compiles `Essai0.Rnw` to `Essai0.tex`, but `C-c C-c` tries to LaTeX `Essai0.Rnw`, which typesets the R chunk literally, and not its outputs. Further trials show that the value of `TeX-engine` is also ignored, and that after moving out of an R chunk, `C-c C-c` will invoke `pdflatex` (the  default) and _not_ `xelatex` nor `lualatex` if so requited in the file variables.

All is fixed by `C-u C-c C-n` (reset AUXTeX). But it is annoying...

I'll try to debug that, but I'm a bit out of my depth here...

jabranham:
Closes #98, closes #114, closes #136, closes #159, closes #215,
closes #260, closes #338, closes #354, closes #338, closes #405,
closes #438, and closes #487.


doc/newfeat.texi
```diff
@@ -59,6 +59,13 @@ The following have been made obsolete, see their documentation for more detail:
 
 @itemize @bullet
 
+@item Libraries for literate data analysis.
+This includes @code{ess-noweb}, @code{ess-swv}, and related
+functionality like @code{Rnw-mode}. Users are encouraged to switch to
+one of several other packages that deal with these modes. For example,
+polymode @url{https://github.com/polymode/polymode} or markdown-mode
```
vspinu: Users should be directed to [poly-R](https://github.com/polymode/poly-R/) and polymode [doc page](https://polymode.github.io/). Just installing poly-R from MELPA should be enough for most users. 

Also maybe mention that `ess-noweb-mode` can be still used by explicitly requiring them.

doc/newfeat.texi
```diff
@@ -59,6 +59,13 @@ The following have been made obsolete, see their documentation for more detail:
 
 @itemize @bullet
 
+@item Libraries for literate data analysis.
```
vspinu: Did you forget to say that they are "obsoleted and not loaded by default"?

lisp/obsolete/ess-noweb-mode.el
```diff
@@ -1789,6 +1789,32 @@ thread."
       (ess-noweb-tangle-thread chunk-name buffer))))
                                         ;menu functions
 
+;;;###autoload
```
vspinu: Shall we remove this autoload and file associations? If we don't people might not noticie the difference.

lisp/obsolete/ess-noweb-mode.el
```diff
@@ -1789,6 +1789,32 @@ thread."
...
+(defun Rnw-mode ()
+  "Major mode for editing Sweave(R) source.
+See `ess-noweb-mode' and `R-mode' for more help."
+  (interactive)
+  (require 'ess-noweb);; << probably someplace else
+  (setq ess--make-local-vars-permanent t)
```
vspinu: How about adding a obsoletion warning here? Or how should we go about it?

jabranham: Anytime you load a lisp file from an obsolete directory Emacs automatically issues a warning message, so this is taken care of for us.

lisp/obsolete/ess-noweb-mode.el
jabranham: I think the autoloads don't get generated from the obsolete/ directory but will double-check.

doc/newfeat.texi
jabranham: Indeed, I'll add that.

lisp/obsolete/ess-noweb-mode.el
mmaechler: Well, I'd be happy if I could switch away from ess-noweb to something that's uniformly better.
.. and I have used versions of polymode for a long time until a few months ago.
But it's just much less usable for me than ess-noweb last I tried.  Also polymode was split in submodules, but MELPA or ELPA or github all looked confusingly different and I gave up for the time being.   

Second issue: Suddenly to use ESS with R you need more than ESS.   This is a big change from a user / installation point of view.  Notably think of people using this via an ubuntu package, or via Vincent's Windows package.  Maybe there's should then be a super package that contains both ESS and the *.Rnw *.Rmd ... enabling packages ?

At the moment, in any case, I'd be horrified if I would be coerced into not having ess-noweb available in ESS anymore.

vspinu: >  I have used versions of polymode for a long time until a few months ago.

polymode has been rewritten at the end of August. Should provide a much better user experience. 

> I gave up for the time being.

You just have to do `M-x package-install poly-R`.

> I'd be horrified if I would be coerced into not having ess-noweb available in ESS anymore.

At the moment other packages are horrified because ess-noweb is a very bad citizen. It took many hours to some package maintainers to figure out that ess-noweb is their source of problems. 

In any case, ess-noweb is not going away for the foreseeable future. It's just obsoleted and not loaded by default.

lisp/obsolete/ess-noweb-mode.el
jabranham: They don't get generated by our makefile.

doc/newfeat.texi
jabranham: I added those links.

lisp/obsolete/ess-noweb-mode.el
jabranham: poly-R is essentially the super package that enables R with literate markdown or latex. If you install it then ESS and markdown get pulled in as dependencies.

vspinu: ESS is actually not an explicit dependency because of the plethora of ways it could be installed.

jabranham: Leaving it out of the dependency list doesn't mean that poly-R doesn't depend on it. The second line is (require 'ess), which seems pretty explicit to me. Any decent package management system will pull in ESS if you install poly-R.

doc/newfeat.texi
```diff
@@ -59,6 +59,14 @@ The following have been made obsolete, see their documentation for more detail:
 
 @itemize @bullet
 
+@item Libraries for literate data analysis are obsolete and not loaded by default.
+This includes @code{ess-noweb}, @code{ess-swv}, and related
+functionality like @code{Rnw-mode}. Users are encouraged to switch to
+one of several other packages that deal with these modes. For example,
+polymode @url{https://github.com/polymode/poly-R/},
+@url{https://polymode.github.io/}, or markdown-mode with edit-indirect
+@url{https://jblevins.org/projects/markdown-mode}.
```
mmaechler: Markdown mode (without polymode etc) really is not a substitute for ess-noweb:
1. It does not deal with *.Rnw at all.
2. The most important functionality for *.Rmd and *.Rnw is a good user interface to send R code to the R process, current chunk, all chunks,  all chunks above 'point', etc,  but also to Sweave, knit, tangle, or purl the whole file.. (and of course, moving from chunk to chunk etc but that's almost too fundamental to be noteworthy) *and* `ess-noweb` has most of that in its mode specific menu  `Noweb` including submenus.

So I don't see "several" packages.  It's the part of the polymode bundle you get via poly-R  only, AFAICS.

lisp/obsolete/ess-noweb-mode.el
vspinu: I didn't say it wasn't requiring ESS, of course it does because it needs R-mode. Maybe it's a good time to think about extracting R-mode from ESS actually. 

> Any decent package management system will pull in ESS if you install poly-R.

I don't know what "decent" package management system you have in mind but MELPA does not fetch packages just based on some requires in the source code. 

I am not entirely sure why you are bringing all this up. Are you unhappy with how poly-R is packaged? I don't really see any other way atm.

doc/newfeat.texi
jabranham: Markdown-mode lets you edit code blocks similar to how org-mode does: you do `C-c '` which pops up a new buffer in the proper major mode. Since many Emacs users also use org-mode, this is pretty intuitive.

lisp/obsolete/ess-noweb-mode.el
jabranham: It seems odd not to declare ESS as a dependency. Installing poly-R from MELPA currently fails because it can't compile without ESS. But it's not really any of my business since I don't work on polymode. :-)

jabranham:
Any objections to merging this?

doc/newfeat.texi
mmaechler: Yes (about editing), and indeed thank you for teaching me about `C-c '`.  But that's only one important feature when working with *.Rmd.  The other is interaction with the *R* process buffer. Sending chunks (one by one, conveniently /  all chunks above point / all chunks / knit the whole document / render the whole document / .. etc.    Things that were even missing with polymode-R (forgot the correct name) when I last looked (which was several months ago). *And* for people not using this daily, having nice Emacs menus, which show you possibilities and simultaneously mention the key strokes.

mmaechler:
in the above sense I'm unhappy about obsoleting something with important functionality that is completely unbundled from ESS.   People who have been using Debian, Ubuntu, SuSE, even Windows have known for 20 years they need Emacs + ESS.    That would no longer be true;  they now also need poly-R  .. something not related to ESS in any formal or automatic way.

doc/newfeat.texi
vspinu: All the things that you mention (menus, evaluation, navigation) are now in polymode since end of September after the rewrite. Might not yet be perfect, but things could always be smoothed by next release of ESS.

vspinu:
> That would no longer be true; they'd now also need poly-R .. something not related to ESS in any formal or automatic way. 

There are pros and cons to this but I think the general trend these days is towards disintegration of big monolithic packages into smaller re-usable parts. 

> I'd really like we think of a setup where the user of next ESS who does not have poly-R will be told about it, 

Cannot get much simpler than `M-x package-install poly-R`. Not sure, but in principle there can be a menu item "Install poly-R from MELPA", or do you have something else in mind?

vspinu:
> poly-R because it's not on ELPA 

One possibility is to check if MELPA is in the package-list and ask the user if he or she wants to add MELPA to it. This might be a bit too user-friendly and might lead to confusing behavior. I think every user has to get accustomed to package archive in emacs and MELPA.

> But actionally more: something that's triggered when opening an *.Rmd

A fake mode of some sort. Could be, but then we have to play hacks with auto-mode-alist because the order of poly-R/ESS initialization would play a role. Will give it a thought.

rsparapa:
I have to agree with Martin.  Let's just include poly-R and markdown mode with ESS and make everyone's life easier (except for maybe Vitalie's ;o)

lionel-:
Maybe we should have another repo for "ESS the distribution of Emacs packages"? This (emacs-ess/ESS) is the repo for "ESS the Emacs package".

rsparapa:
Hi Lionel:  Not sure what you mean by that.  We are already including the ESSR package.  Let's just include poly-R and be done with it.  This has been my top request for several years.  Thanks, Rodney

eddelbuettel:
I can't help but think that this is related to the parallel discussion about `make dist` and/or installations from tarball.  And I can't shake the feeling that parts of this team is so focused on some variant of (perceived) purity in release or distribution that the long-standing _other_ ways get the shorter stick.

So in other words, if you contemplate changes is exactly _what_ is packaged as 'ESS' and _how_ it is shipped, it might be a good idea to test installations on more than one way.  Maybe check with Detlef. Maybe check with Vincent.  Maybe check if it works with all of Martin's workflows.  Maybe check with Rodney and his requirements.   It is complicated as there are so many ways to get and install ESS, but that is the price you pay for creating such a wonderful and widely used tool.

vspinu:
It's indeed about package and not about including polymode in ESS code base per se. I think the uniform-distro initiative on the other thread is great, but yet again the multitude of ways ESS is distributed and decades old habits seem to be putting breaks over there as well. Maybe we should have two tarballs (and paralleling distro targets), one for "pure" ESS and one for batteries included ESS+julia+poly-R+poly-markdown+poly-noweb.

And BTW, judging from the latest announcement, Vincent removed polymode from his package.

Back to the case. This PR should be merge or ess-noweb should be fixed. The latter is extremely unlikely to happen, so I don't think we have a real choice here.

rsparapa:
The reason that Vincent dropped polymode is probably because it is now unmaintained and you changed the name to poly-R.  This came as a surprise to many of us.  If these packages were wrapped within ESS, then much of the confusion would disappear.  I'm sure Vincent would welcome that and include it.

mmaechler:
@vspinu yes, you are right.  This is primarily about packaging and distribution  and much less about how exactly the different parts should be source-code maintained.  
As Dirk @eddelbuettel mentioned,  and Tony @blindglobe has reminded us a few times, ESS has also been the longest maintained piece of FOSS statistical software in existence *ever* - notably if you count its first years as 'S-mode'.  And in all that time, it has been distributed as a tarball (and then additionally zip) produced via 'make'... and the users got everything they needed from that tarball and could use 'make install' themselves to get going.  Rodney @rsparapa , Dirk @eddelbuettel , me and other long timers, including most of the downstream packagers have been going happy quite successfully with that setup since the early 1990's (S-mode has existed since 1989)...
Even the name ESS (created only in the later 90's) was used after basically all reasonable stats software had gotten interfaces for Emacs bundled together .. to make it seemless and coherent for statisticians using emacs and get a "uniform" experience using different stats software packages, see our (quite late) publication about it (section 4, history of ESS).
https://www.researchgate.net/publication/261578330_Emacs_Speaks_Statistics_A_Multiplatform_Multipackage_Development_Environment_for_Statistical_Analysis  (or from the publisher)
https://www.tandfonline.com/doi/abs/10.1198/1061860042985 
(which is pointed to also from our web page: https://ess.r-project.org/index.php?Section=documentation&subSection=articles
---------------------------------------------------------------------------
To continue that tradition it would really be nice if'd continue packaging the ESS polymode sources for distribution.. and notably if we'd also provide tests (maybe we do) for the correct "polymode bundle + ESS" functioning together correctly.

lionel-:
> Hi Lionel: Not sure what you mean by that. We are already including the ESSR package. Let's just include poly-R and be done with it. 

I mean that polymode, ESS, and even ESSR should be developed as separate Emacs packages, just like base, utils, stats, or graphics are developed as separate R packages even though they are distributed with R.

New standards of Emacs development have emerged and we should aim towards those. That's why I also think the most important vector of package distribution (especially for general purpose packages like polymode) is package.el. Other means of distribution are possible and desirable, but should be kept separate. Ideally the developers of ESS and ESSR wouldn't have to worry about these other means of distributions at all.

I don't think our goals are contradictory. We just need to find the right workflow.

blindglobe:
Separate packages were always an idea - however ensuring a contract between
packages to ensure static API behavior was/is hard.  (That is - you want
the other packages to evolve, but not the functionality that is leveraged
and necessary for the current "local" /immediate package that you are
working on).  The easiest and most bloated solution was what I chose -
integrate everything into a huge single package.   This was reasonable in
1994, but maybe we can do better now....

jabranham:
Sorry, didn't mean to close this.

dsteuer:
Sorry for coming late to this discussion. I'm just downstream, therefore never had a desire or need to comment on ESS development. Everything went on in nice incremental steps. At least as far as I was concerned. The announcement told me to get a new tarball and build a new rpm. Fine!

May be *ELPA is the most important way to go for an emacs package. Nevertheless you'll lose
users, if it's the only. I have a surprisingly large number of requests for corporate settings, where only
rpms are allowed, centrally installed by some IT guys. If you drop the requirement that ESS is self-contained, it will be much harder to package outside of *ELPA. 

In the same mind setting Dirk's @eddelbuettel work on deb packages for CRAN, or similar approaches for SuSE or Fedora, would be useless, if the *.packages() functions in R would be the only interface relevant for R users. Download numbers contradict that impression! Therefore I think being able to easily generate distribution packages should have some priority, if you are interested in visibility of ESS.  

Dropping functionality we old timers are used to, wouldn't help, too. So I wholeheartedly want to put my vote for @mmaechler 's comment above.

It is very easy to lose trust in a software project and very hard to get it back. If "revolutionary" changes are planned or needed I, as downstream and user, would like something like ESS-ng to try it and change over if ESS-ng is a full replacement or improvement. And meanwhile have my workhorse ESS working as expected.

My 2¢.

lionel-:
> May be *ELPA is the most important way to go for an emacs package. Nevertheless you'll lose
users, if it's the only

Just to be clear, no one is arguing for that.

jabranham:
This discussion seems to have drifted away a bit, let me see if I can put it back on track:

I'm proposing to obsolete the ess-noweb (and related) stuff. It's broken in many ways (see all the bugs I reference at the top) and, worse, breaks third party packages because it changes some fundamental Emacs behavior. It's not under active development and hasn't been for some time. A replacement is available that learns from many of the mistakes in ess-noweb. It seems desirable to obsolete ess-noweb if a better replacement is available to get users to switch. It is normal software development to obsolete old tools that aren't maintained for newer tools.

As far as distribution goes, we could distribute two tarballs: A "just ESS" one and an "ESS+" tarball that includes polymode, julia-mode, and any other third-party packages (potentially edit-indirect for example, see #756). I think that would make everyone happy, right?

eddelbuettel:
Thumbs up for ESS+.  That is effectively what many downstream aggregators (not blessed with package management system) do. Vincent is one example,  and Kieran Healey has a social-sciences toolbox containing this.  I may even use it because I am fairly positive Debian does not have polymode packaged, and am unsure if Julia.el is/how current it is.  Having a single reliable bundle helps.

rsparapa:
I concur with Tony and Dirk.  Good strategy with respest to polymode, etc.  As far as noweb goes, I also concur.  noweb has frustrated me for decades!  If there is a better alternative, please replace it: we will not agonize its loss ;o)  Go for it Alex!

vspinu:
It would make a lot of sense if the new tarballs would follow package.el requirement. Then they could be immediately installed with `package-install-file` by the end users and used as is by downstream distributors. #752 for the discussion.

@rsparapa 
> The reason that Vincent dropped polymode is probably because it is now unmaintained and you changed the name to poly-R. This came as a surprise to many of us. 

Sorry for the confusion. It's not what happened. `polymode` was just dis-aggregated into multiple libraries. Core `polymade` and separate packages for each mode, `poly-R` being one of them. I cannot possibly maintain every single poly-mode that people might need inside the core repo.

rsparapa:
What is the issue with maintaining polymode?  The various versions/platforms of emacs?

vspinu:
Polymode and all other poly-modes on the [org](https://github.com/polymode/) are maintained.

dickmao:
> It would make a lot of sense if the new tarballs would follow package.el requirement. Then they could be immediately installed with `package-install-file` by the end users and used as is by downstream distributors.

This was my inclination as well but senior ESS members have [disabused](https://github.com/emacs-ess/ESS/issues/754#issuecomment-440602575)  me of it.  Debian introduced a dual elpa-debian packaging scheme that @eddelbuettel struggled mightily with, and probably would not wish upon @dsteuer *et al*.

I think the path forward is to repurpose #747 with a `make package-install` that does not touch the time-honored semantics of `make install`.

eddelbuettel:
Allow us to disagree and please stop wasting our time.  You insisted on knowing better how to package for Debian, I demonstrated to you that that appraoch was neither fully tested nor actually installable.

jabranham:
@eddelbuettel please do not try to dissuade anyone from contributing to ESS.

mmaechler:
Apropos poly-R:  It is not even in `melpa-stable`.  Do you expect decent companies to allow their non-IT staff to use something as unstable as `melpa` (without `-stable`) ?

blindglobe:
Does Novartis count as decent? :).

I know, we are sometimes all too questionable...

(Thanks for the laugh this morning, I really needed it :) )

Best,
Tony

vspinu:
> Apropos poly-R: It is not even in melpa-stable.

[it is](https://stable.melpa.org/#/poly-R)

jabranham:
Merged this and opened #780 for what to include in ESS+.

PR #740, opened 2018-11-09
Make ess-roxy work with arbitrary prog-modes

### LINKED ISSUE #439: C++ support for roxygen editing?
Rcpp makes it somewhat easy to write code on the C++ side.  The tooling then carries code over into a generated R file where `roxygenize()` will process any included documentation.

So many of us have gotten to writing documentation _meant for roxygen_ in the C++ file.   

But the editing support is just not there, whereas RStudio does it a little better.   I really know next to nothing about multi-mode but could the existing (and excellent !!) roxygen support for R files be carried over to C++ files?

[ I have the feeling I asked that once before years ago but my either my search-fu lets me down, or my memory is faulty.  ]

jabranham:
Closes #439 

One of the tests is complaining about not finding the right face but I think it may be an issue with the tests, it works locally for me. Don't have time right now so I'll dig in later.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -47,6 +47,10 @@ your Emacs file:
 See the documentation of @code{display-buffer-alist},
 @code{display-buffer}, and @xref{Window Choice,,, emacs} for details.
 
+@item @code{ess-roxy-mode} can now be enabled in non-R buffers.
+This is primarily intended to support roxygen documentation for cpp
+buffers. Preview functionality is not supported outside R buffers.
+
 @item Options for 'ess-gen-proc-buffer-name-function' have been renamed.
 ess-gen-proc-buffer-name:projectile-or-simple was renamed to
 ess-gen-proc-buffer-name:project-or-simple and
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -1494,11 +1494,6 @@ the line at point"
   :group 'ess-roxy
   :type 'string)
 
-(defcustom ess-roxy-re "^#+'"
-  "Regular expression to recognize roxygen blocks."
-  :group 'ess-roxy
-  :type 'string)
-
 (defcustom ess-swv-pdflatex-commands '("texi2pdf" "pdflatex" "make")
   "Commands to run a version of pdflatex in  \\[ess-swv-PDF];
 the first entry is the default command."
@@ -1513,7 +1508,7 @@ to sweave the current noweb file and latex the result."
                  (const :tag "On" t)))
 
 (defvar ess-roxy-insert-prefix-on-newline t
-  "When non-nil, `ess-newline-and-indent' will make sure the new
+  "When non-nil, `ess-roxy-newline-and-indent' will make sure the new
 line starts with the roxy prefix.")
 
  ; System variables
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -631,9 +631,7 @@ Executed in process buffer."
                 (ess-beginning-of-function 'no-error)))
   (setq-local end-of-defun-function
               'ess-end-of-function)
-
-  (ess-roxy-mode t)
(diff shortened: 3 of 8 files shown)


eddelbuettel:
Getting C++ support would be wicked. I just bugged Kevin about it for the 'other IDE' I sometimes use :)
May as well lead the way in ESS.

jabranham:
@lionel- I now recall why I hooked it into the r package stuff: to turn it on in cpp buffers that are in R packages. I see a few options:

- leave it off in cpp buffers
- add it to package hook and R mode
- add it to the package hook and not R mode

I'm not sure one of those is better than the others. Number two is probably most convenient but also the least nice code-wise and might leave some users stumped as to how to turn off ess-roxy in R packages (if anyone even does that?)

WDYT?

mmaechler:
> > start working on documentation in scripts, before moving functions to a proper package.
> 
> I don't do that so I didn't consider that workflow. Makes sense though. I'll modify this commit so that roxygen is on by default in R mode.

Good.  I *only* use roxygen for documenting functions that are not yet exported in a package.
I only do roxygenize *once* in the life time of a function  -- often once I decide to export it -- and then only work with the *.Rd  file which has much richer format than is typically used by people who keep their docs in roxy-comments.   I've not seen much nice help pages derived from roxygen.  Rather those look typically minimal rather than much enlightening... .. (and yes, of course, I'm biased as we all are).

lionel-:
I wonder if we need some kind of rcpp-minor-mode. It'd be useful to bind Rcpp-only keys. It could be enabled based on some heuristics (presence of include Rcpp.h, Rcpp comments, or Rcpp types). Best not to tie it to package mode since you can have Rcpp scratch files.

vspinu:
Once we have specific functionality for rcpp files (front end for sourceCpp for example) that should be the way to go. But it's not substitute of r-package-mode in `./src` because most of the files in a package might not have `include Rcpp` in them. Roxy is a minor-mode by itself and doesn't require an intermediate minor-mode to be activaed. 

To make it worth the trouble, for such a minor mode we would need something more substantive than sourceCpp.

lionel-:
> most of the files in a package might not have include Rcpp in them

So we need other heuristics. I just think we should not enable it unconditionally in all C++ files because not all packages use Rcpp. In packages I guess it is sufficient to check the LinkingTo field of DESCRIPTION. This should cover most cases: check for an include to cover scratch files, or for the linkingto field in packages.

vspinu:
So, you say we still activate it per-package but only if linking-to is right. This could work. 

We could set a buffer local from our r-package initializer where we sniff for the package name in DESCRIPTION file. Then other minor-modes like roxy could pick it up. This is not a bad idea AFAIC.

jabranham:
I think adding this diff is enough for the current PR, right? I don't think we need to check LinkingTo just for roxy stuff. It's on by default in all R buffers whether or not we're in a package or if that package actually uses roxygen and (to my knowledge) no one has complained about that.

```diff
diff --git a/lisp/ess-roxy.el b/lisp/ess-roxy.el
index fcbd278a..c7f36255 100644
--- a/lisp/ess-roxy.el
+++ b/lisp/ess-roxy.el
@@ -1035,6 +1035,14 @@ Only do this if in a roxygen block and
 
 (advice-add 'c-fill-paragraph :before-while 'ess-roxy-cpp-fill-paragraph)
 
+(defun ess-roxy-enable-in-cpp ()
+  "Enable `ess-roxy-mode' in C++ buffers in R packages."
+  (when (ess-r-package-project)
+    (ess-roxy-mode)))
+
+(with-eval-after-load "cc-mode"
+  (add-hook 'c++-mode-hook #'ess-roxy-enable-in-cpp))
+
 (provide 'ess-roxy)
 
 ;;; ess-roxy.el ends here
```

jabranham:
I applied that diff and merged into `next`.

PR #743, opened 2018-11-12
lisp/Makefile: Copy el files into LISPDIR

### LINKED ISSUE #741: ess-autoloads.el not copied by make install
With ESS 18.10.2, `ess-autoloads.el` is created by `make install`, but not copied over to the lisp installation directory. I'm not sure if it is intended or not.

I tried manually copying the file to the ess lisp installation directory and it seems to work correctly. However (and this may well be a separate issue), loading ESS with `(load "ess-autoloads")` causes a problem with my redefinition of the variable `ess-swv-pdflatex-commands` in my .emacs init file. I suspect this could cause problem for others too.

Should I just stick to the "classic" initilization mechanism `(require "ess-site")`?

> (This is for my Emacs Modified for [Windows|macOS] distributions.)

### LINKED ISSUE #742: Cannot open ess-rd.el
Loading ESS with `(require 'ess-site)` in a `site-lisp.el` file for a by compiled distribution of ESS results in the following error at startup (from the `*Messages*` buffer):

```
byte-code: Cannot open load file: No such file or directory, ess-rd.el
```

The file `ess-rd.elc` is present in the appropriate `lisp` directory.

Copying the original `ess-rd.el` in the aforementioned `lisp` directory solves the problem and Emacs and ESS start normally.

- GNU Emacs 26.1 (build 1, x86_64-apple-darwin14.5.0, NS appkit-1348.17 Version 10.10.5 (Build 14F2511)) of 2018-05-30
- ess-version: 18.10.2 [Released git: d4cd65da6dbf]

jabranham:
Closes #741 closes #742

### DIFF
--- a/lisp/Makefile
+++ b/lisp/Makefile
@@ -39,6 +39,7 @@ dist: all
 
 install: dist
 	-$(INSTALLDIR) $(LISPDIR)
+	$(INSTALL) $(EL) $(LISPDIR)
 	$(INSTALL) $(ELC) $(LISPDIR)
 	if [ -f /etc/debian_version -a -n "$(SITELISP)" -a ! -f "$(SITELISP)/ess-site.el" ] ; \
 	then \


lisp/Makefile
```diff
@@ -39,6 +39,7 @@ dist: all
 
 install: dist
 	-$(INSTALLDIR) $(LISPDIR)
+	$(INSTALL) $(EL) $(LISPDIR)
```
dickmao: $(ELS), perhaps, or am I missing something.

vspinu: That autoloads filter in ELS is not triggered on the clean repo. When ess-autoloads.el is in the repo the elc is not generated. I think `batch-byte-compile` just skips it or something.

vspinu:
Why is this going into master? Aren't we done with ESS18?

mmaechler:
Hmm, I'm happy if obvious small bugfixes still go to master.  At this point, we don't really know if it wouldn't make sense to release a stable  18.10.<final>   with all "obvious small buglets" fixed.

I'd personally be happy to have that as a 'stable benchmark' for a while.  Also, if possible (and nothing of that kind is impossible with git AFAIK) I'd be happy if you'd __eventually__ (not just these days, yet!)  do the following simultaneously once we swtich : 
turn  current 'master' into a branch '18.10.x' (or '18.10')
turn  current 'next' into 'master'

Issue #745, opened 2018-11-13
Julia complete code in ESS REPL error


stardiviner:
![image](https://user-images.githubusercontent.com/378699/48384614-22000400-e726-11e8-9093-2cf8d3c5588d.png)

This is the screenshot, I start a fresh new REPL with `julia` command commands from ESS. But it does not show methods like `Pkg.add` `Pkg.update` etc.

jabranham:
This was closed by e92d3304cc974b7e06c1df0b47f988904d53f155

stardiviner:
Thanks for quick fix.

PR #746, opened 2018-11-13
list all functions in module

nverno:
update julia completion to list all the imported functions in module, like 
the REPL does. Filters out what I assume are compiler generated symbols
starting with "#", seems to work for now

#745

### DIFF
--- a/etc/ess-julia.jl
+++ b/etc/ess-julia.jl
@@ -76,9 +76,9 @@ end
 # add                           Function
 # free                          Function
 function components(m::Module)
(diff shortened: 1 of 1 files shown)


jabranham:
Thanks!

A minor nit - in the future, can you make sure your branch is up to date? The parent of the commit you sent was about a month old. It's not a huge deal; I rebased and merged it into master.

nverno:
Oh sorry about that, I thought I had rebased onto master. Will make sure in future, thanks

PR #747, opened 2018-11-13
Home directory installs

dickmao:
Remove any suggestion that `Makeconf` should be edited.
Remove any suggestion that package can be installed to systemwide directories (a home directory install becomes the *only* option).
Add an `INSTALL` file with instructions for install.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ ess-[1-9]*.[0-9][0-9]-[0-9]
 ess-[1-9]*.[0-9][0-9]-[0-9][-.]???
 lisp/julia-mode.el
 lisp/ess-autoloads.el
+lisp/dist
 tmp*
 .DS_Store
 ._.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ doc/ess.*
 doc/info/ess.info
 doc/readme.*
 doc/refcard/refcard.*
+doc/*.log
 *history
 *NEWS
 *VERSION
@@ -31,3 +32,4 @@ lisp/ess-autoloads.el
 tmp*
 .DS_Store
 ._.DS_Store
+Makeconf.local
--- a/INSTALL
+++ b/INSTALL
@@ -1,0 +1,7 @@
+To install from source
+
+cp Makeconf.local.template Makeconf.local
+
+and set variables in Makeconf.local as necessary.  Then
+
+make install
--- a/Makeconf
+++ b/Makeconf
@@ -4,6 +4,15 @@
 ## 1. Edit Section 1 and 2 below
 ## 2. Execute: gmake install
 
+## To install ESS for yourself:
+## 1. cat > Makeconf.local -<<EOF
+##    PREFIX=/home/<user>/.local
+##    SITELISP=/home/<user>/.emacs.d/elpa
+##    INFODIR=$(LISPDIR)
+##    ETCDIR=$(LISPDIR)/etc
+##    EOF
+## 2. Execute: gmake install
+
 ## Section 1
 ## Installation variables for your emacs variant
 ##
@@ -17,14 +26,14 @@
 ## PREFIX(DESTDIR) Directory prepended to SITELISP, INFODIR, DOCDIR & ETCDIR
 ##                 Specify either PREFIX or DESTDIR to over-ride /usr
 DESTDIR=/usr
-PREFIX=$(DESTDIR)
+PREFIX ?= $(DESTDIR)
 
 ##__ GNU Emacs __
 EMACS ?= emacs
-SITELISP=$(PREFIX)/share/emacs/site-lisp
+SITELISP ?= $(PREFIX)/share/emacs/site-lisp
 LISPDIR=$(SITELISP)/ess
-INFODIR=$(PREFIX)/share/info
-ETCDIR =$(PREFIX)/share/emacs/etc/ess
+INFODIR ?= $(PREFIX)/share/info
+ETCDIR  ?= $(PREFIX)/share/emacs/etc/ess
 
 ##__ XEmacs __
 #EMACS=$(PREFIX)/bin/xemacs
@@ -50,7 +59,7 @@ ETCDIR =$(PREFIX)/share/emacs/etc/ess
 # ETCDIR=$(PREFIX)/ess-mode/etc
 
 ## the following command is the same for recent versions of both Emacs and XEmacs
-EMACSBATCH = $(EMACS) -batch -no-site-file -no-init-file
+EMACSBATCH = $(EMACS) -Q --batch
 
 ## Section 2
 ## Installation variables for your unix variant
--- a/Makeconf
+++ b/Makeconf
@@ -1,18 +1,5 @@
 ####======= Common Declarations for *all* ESS -*-Makefile-*-s ==========
 
-## To install ESS for all users on your unix system:
-## 1. Edit Section 1 and 2 below
-## 2. Execute: gmake install
-
-## To install ESS for yourself:
-## 1. cat > Makeconf.local -<<EOF
-##    PREFIX=/home/<user>/.local
-##    SITELISP=/home/<user>/.emacs.d/elpa
-##    INFODIR=$(LISPDIR)
-##    ETCDIR=$(LISPDIR)/etc
-##    EOF
-## 2. Execute: gmake install
-
 ## Section 1
 ## Installation variables for your emacs variant
 ##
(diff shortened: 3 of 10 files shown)


jabranham:
Thanks for the PR! A few comments/questions:

> As much as practicable, installation from melpa and from github should execute the same code.

I'm not sure I agree. MELPA (and this PR) use package-build.el but I don't think we should rely on that. I'd like ESS to be installable just with the built-in package.el. We're close to that already. Possibly all that's required is adding ess-pkg.el and renaming the tarball releases to `ess-18.10.2.tar.gz`. 

> Allow users to create their own Makeconf.local to perform home directory installs as opposed to /usr installs.

How is this better than editing makeconf? Can't people already set the appropriate variables there? We should probably add how to set them for a typical gnu/linux machine (e.g. LISPDIR should be `~/.emacs.d/lib/ess-18.10.2/` or something like that.)

dickmao:
`cask` which I view as the *de facto* elisp dependency manager (like gradle for android, or sbt for scala) also uses `package-build` to generate that pesky `ess-pkg.el`.  I rather like cask -- it reduces the dist making to a single line `cask package`.

It seems packaging is fraught with some controversy #707 #670 .  mmaechler's remarks on GH lock-in are well taken, but I personally am more in the camp of "separation of responsibility" and knowledge delegation to downstream distributors.

The `test-install` code stands on its own as (what I consider) an important hedge against new user dismissal of ESS.  One of its artifacts is the `ess-pkg.el` which I exploit for local installs.  I don't know of any other non-manual way to construct that file, but I haven't researched it.

As for `Makefile.local`, I generally don't customize files that are git-tracked such as `Makeconf`, lest it find its way back into `master`.

vspinu:
How about piking Makevars.conf from a variable? 

I agree with @jabranham that Makeconf was designed to be modified, but I don't think that Makeconf is necessary. We better be assigning all vars with  `=?`.  Users can then source their Makeconf to initialize those variables. 

Regarding cask and pkg-build.el. More dependencies we have harder it is to understand what is going on.  Cask gave big headaches in the past when it was breaking. But I like the idea of mimicking the melpa install with pkg-build.el as MELPA is our main distribution channel and wrestling with package.el is one big pain in the neck. Would be nice to extract that multi-line --eval into a separate elisp target though.

Makeconf
```diff
@@ -4,6 +4,15 @@
...
 ## 2. Execute: gmake install
 
+## To install ESS for yourself:
+## 1. cat > Makeconf.local -<<EOF
+##    PREFIX=/home/<user>/.local
+##    SITELISP=/home/<user>/.emacs.d/elpa
+##    INFODIR=$(LISPDIR)
+##    ETCDIR=$(LISPDIR)/etc
+##    EOF
+## 2. Execute: gmake install
```
vspinu: Why not `source Makeconf.local` before the `gmake install` without us bothering with it?

lisp/Makefile
```diff
@@ -37,9 +43,44 @@ all: julia-mode.el $(ELC) ess-custom.el ess-autoloads.el
...
+	$(INSTALLDIR) dist
+	if [ ! -s "dist/$(PKBUILD).tar.gz" ] ; then \
+	  cd dist ; curl -sLOk https://github.com/melpa/package-build/archive/$(PKBUILD).tar.gz ; fi
+	cd dist ; tar xfz $(PKBUILD).tar.gz
+	cd dist/package-build-$(PKBUILD) ; make -s loaddefs
+	$(INSTALLDIR) dist/recipes
+	if [ ! -s "dist/recipes/ess" ] ; then \
+	  cd dist/recipes ; curl -sLOk https://raw.githubusercontent.com/melpa/melpa/master/recipes/ess ; fi
+	$(EMACSBATCH) -L dist/package-build-$(PKBUILD) \
+	--eval "(require 'package-build)" \
```
vspinu: Let's extract this spaghetti into a separate .el file inside etc/targets/ subfolder.

lisp/Makefile
```diff
@@ -37,9 +43,44 @@ all: julia-mode.el $(ELC) ess-custom.el ess-autoloads.el
...
+	--eval "(setq debug-on-error t)" \
+	--eval "(setq package-user-dir (expand-file-name \"dist\"))" \
+	--eval "(package-initialize)" \
+	--eval "(add-to-list 'package-archives '(\"melpa\" . \"http://melpa.org/packages/\"))" \
+	--eval "(package-refresh-contents)" \
+	--eval "(setq rcp (package-recipe-lookup \"ess\"))" \
+	--eval "(unless (file-exists-p package-build-archive-dir) \
+	           (make-directory package-build-archive-dir))" \
+	--eval "(unless (zerop (length \"$(TRAVIS_PULL_REQUEST_SLUG)\")) \
+	          (oset rcp :repo \"$(TRAVIS_PULL_REQUEST_SLUG)\") \
```
vspinu: Can we avoid these variables? Would be nice to be able to run it locally.

dickmao: it can be run locally.  perhaps you should try (although you would need a Makeconf.local)

Makeconf
dickmao: Sure, that's possible but taking this to its logical conclusion I could also put them in my `.bashrc`.
As an old timer (and which emacs and R users aren't?), I too am accustomed to the traditional download-untar-{xmkmf,configure,autogen,etc} way of installation.  But we're now in an age of "configuration as source" and having `git diff` show my customizations to `Makeconf` is imo unsettling.  `Makeconf.local` seemed like the right balance of tastefulness and not going whole hog with a gnu autotools solution (Makefile.am, configure.in, etc.).  I'm perfectly okay with its rejection, however.

vspinu: > git diff show my customizations to Makeconf is imo unsettling. 

I completely agree. Makeconf setup as we have doesn't seem make much sense these days. It's much more fruitful to pick local configuration of some kind.

@mmaechler, @jabranham, @lionel-  would you be against  removing Makeconf in favor of dir local and user root local .ess-makeconfig (or .ess-makevars)?

lisp/Makefile
dickmao: done... for lack of better place moved to `lisp/tmp/ess-package.el`

mmaechler:
1. I'm unhappy that you remove commented donated code that help people installing into e.g. Aquamacs.
AFAICS, it *is* used and can bring more people to use Emacs.  These comments were there to help people.  Why drop them just because they are not useful to you ?

2. The Makefile (+ *local / config / ..) are typically used by downstream packagers.  Are you sure that it is good idea to  "_disallow installing into systemwide directories_" ?

.gitignore
```diff
@@ -28,6 +29,8 @@ ess-[1-9]*.[0-9][0-9]-[0-9]
 ess-[1-9]*.[0-9][0-9]-[0-9][-.]???
 lisp/julia-mode.el
 lisp/ess-autoloads.el
+lisp/dist
```
vspinu: Why within lisp? Doesn't look right to me. Can be `dist` in the root dir.

INSTALL
```diff
@@ -0,0 +1,7 @@
+To install from source
+
+cp Makeconf.local.template Makeconf.local
+
```
vspinu: I don't see Makeconf.local.template. I think you forgot to add it to the git. 

This is a good idea in my opinion. I would say we move current config from makeconf into makefile and rename Makeconf into Makeconf.template (with adjustments). Then people could do `cp Makeconf.template Makeconf`. How does this sound?

lisp/tmp/ess-package.el
```diff
@@ -0,0 +1,58 @@
+;;; ess-package.el --- user customization of ESS
```
vspinu: Can we move this into `etc`? Polluting lisp dir doesn't feel right to me. There will likely be  other helper scripts for our targets in the future.

INSTALL
mmaechler: No, sorry!  `Makeconf` is crucial in the current setup, and for normal users, including me, for *Release*  it is both necessary and sufficient.

Please do not (again!) break the  'make' seetup  by incorrect cleanup (!).
This was particularly bad, for 18.10 and notably broke things for the downstream packaging people.

eddelbuettel:
> Okay, let me take a step back and try to understand the existing rpm-debian support. 

Debian maintainer and lurker here.  I am mostly on board with what Martin says and clashed with nonsensical statements by one of the ESS devs (not naming names here) before.   I have looked after ESS for over a decade now, and based and that periods I can tell you that the install was (from my downstream view) never quite as broken as it was in 2018.  Now, part of this was also my/our doing as I moved in parallel to a (m)elpa setup with Debian (which is the right call).  That now almost works: I currently have to do one extra hoop (as I said before) but as always but of that may have to do with the fact that I don't speak or write elisp well -- but I seem to require `ess-pkg.el` which is AFAICT not provided by ESS itself.  So with that I am reopening this.

In case you are curious my `debian/` directory used to be part of this repo; it no longer is.  You can find it easily on salsa.debian.org under my edd handle.

INSTALL
vspinu: @mmaechler, Makeconf is tracked by git, and that's bad setup for a custom configuration file. What @dickmao proposes is superior to what we have right now. You will still have your Makeconf as Makeconf.template.

jabranham: Can't we just define all the variables in Makeconf with `=?` and then put `-include config.mk` before `include Makeconf` in all the various makefiles? That seems much cleaner to me.

dickmao: The order is currently `include Makeconf`, and then `include Makeconf.local`.  I would avoid `?=` if it's unnecessary (which is why I took most of them out in 17b99cf.  I also don't know why people cannot see `Makeconf.local.template` -- it seems to be git added.

jabranham: > avoid ?= if it's unnecessary 

Why? Using it how I suggested lets users override Makeconf variables if they want and by using `-include` we don't have to do all the checks you do for whether or not the file exists.

dickmao: We can either do as you propose which is replacing all assignments in `Makeconf` with `?=` and include `Makeconf.local` *before* `Makeconf`, ... or....

We can leave `Makeconf` alone, and include `Makeconf.local` *after* `Makeconf`.  Perhaps I am missing something.

jabranham: and include Makeconf.local before Makeconf

`-include`, not `include`. It simplifies this:

```
LOCAL_CONF := $(strip $(wildcard ./Makeconf.local))
ifeq ($(LOCAL_CONF),)
$(error "No ./Makeconf.local found.  Please read INSTALL")
else
include ./Makeconf.local
endif
```

to just 

```
-include Makeconf.local
include Makeconf
```

lisp/tmp/ess-package.el
jabranham: Shouldn't it go into `test/`?

Also, please use lexical binding in all new lisp files.

lisp/Makefile
```diff
@@ -79,13 +106,8 @@ julia-mode.el:
...
 
-ess-autoloads.el:
-	@printf	"\nGenerating $@\n"
-	$(EMACSBATCH) --eval "(progn\
-	(setq make-backup-files nil)\
-	(setq generated-autoload-file (expand-file-name \"$@\"))\
-	(setq find-file-visit-truename t)\
-	(update-directory-autoloads default-directory))"
+ess-autoloads.el: $(filter-out ess-autoloads.el, $(ELS))
+	$(EMACSBATCH) --eval "(package-initialize)" --eval "(package-generate-autoloads \"ess\" \".\")"
```
jabranham: Generating the autoloads should not rely on package.el.

INSTALL
dickmao: I would prefer the former (despite the fact that removing lines is (almost) always better than adding lines).  Lest we get ahead of ourselves, I am in favor of suspending discussion over this minutiae until as such time I can test the "rules of engagement" for debian and/or rpm.

doc/Makefile
```diff
@@ -1,8 +1,13 @@
 ### Makefile - for DOCUMENTATION (./doc) of ESS distribution.
 ###
 
-## Before making changes here, please take a look at Makeconf
 include ../Makeconf
+LOCAL_CONF := $(strip $(wildcard ../Makeconf.local))
```
vspinu: I guess, you could add this at the end of `Makeconf` and not alter all these makefiles.

lisp/tmp/ess-package.el
vspinu: Could be, but I think it could be used for other purposes than tests, like distro building. I guess that's the idea here.

INSTALL
vspinu: The former throws and that will confuse a lot of users who are used to `make install` just working out of the box.

How about splitting this in 2 parts, package-build test target and makeconf.local + install? We pull the non-controversial package-build  and have a separate discussion on the other part.

PR #748, opened 2018-11-13
Add links to help topics in help buffers

jabranham:
I like the links you get from the HTML help pages but I want to be able to view them in Emacs instead of a browser. This works fairly well by opening them up in eww in Emacs:

```R
if (Sys.getenv("INSIDE_EMACS") != ""){
  options(browser =
            function(url)
              system2("emacsclient",
                      paste0("--eval '(eww \"", url, "\")'")))
  options(help_type = "html")
}
```

but it's still slower than the plain text pages we serve by default.

This PR adds links to the plain text style pages. It works by searching for pairs of matching "fancy" (i.e. unicode) quotes ~and seeing if we know about a help topic for the word between them. If so,~ we add a button that links to the corresponding help page.

For all the help pages I tested, the slowdown as we regex-search through the buffer is not noticeable. I imagine if you find a super complicated or long buffer it might take a quarter of a second or something to add the links. ~So I added a defcustom to control this behavior.~

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -2243,6 +2243,8 @@ state.")
 
 (defvar ess-r-help-mode-map
   (let ((map (make-sparse-keymap)))
+    (set-keymap-parent map (make-composed-keymap button-buffer-map
+                                                 ess-help-mode-map))
     (define-key map "s<" 'beginning-of-buffer)
     (define-key map "s>" 'end-of-buffer)
     (define-key map "sa" 'ess-skip-to-help-section)
@@ -2267,7 +2269,36 @@ state.")
         ess-help-sec-keys-alist ess-help-r-sec-keys-alist ; TODO: Still necessary?
         inferior-ess-help-command inferior-ess-r-help-command
         ess-get-help-topics-function #'ess-s-get-help-topics-function
-        inferior-ess-help-filetype nil))
+        inferior-ess-help-filetype nil)
+  (ess-r-help-add-links))
+
+(define-button-type 'ess-r-help-link
+  'follow-link t
+  'action (lambda (_) (ess-r-help-button-action)))
+
+(defun ess-r-help-button-action ()
+  "Display help for button at point."
+  (let ((text (get-text-property (point) 'ess-r-help-link-text)))
+    (ess-display-help-on-object text)))
+
+(defun ess-r-help-add-links ()
+  "Add links to the help buffer."
+  (save-excursion
+    ;; Search for fancy quotes only. If users have
+    ;; options(useFancyQuotes) set to something other than TRUE this
+    ;; probably won't work. If it's FALSE, R outputs ascii ', but
+    ;; searching through the whole buffer takes too long.
+    (while (re-search-forward "‘[^[:space:]]+?’" nil t)
+      (let ((inhibit-read-only t)
+            (beg (match-beginning 0))
+            (end (match-end 0))
+            ;; Remove surrounding quotes so the text property is the
+            ;; name of the link:
+            (text (string-trim (match-string 0) "[\"‘]*" "[\"’]*")))
+        (add-text-properties beg end (list 'ess-r-help-link-text text))
+        (make-text-button beg end
+                          'type 'ess-r-help-link
+                          'help-echo (format "mouse-2, RET: Help on %s" text))))))
 
 
 
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -61,6 +61,12 @@ ess-gen-proc-buffer-name:project-or-directory. As the name suggests,
 these now rely on project.el (included with Emacs) rather than
 projectile.el, which is a third-party package.
 
+@item ESS[R] help pages now provide links to other help topics.
+This is similar with what you would see with, for example
+@code{options(help_type = ``html'')} but works with the plain-text
+version as well. This only works with @code{options(useFancyQuotes =
+TRUE)} (the default).
+
 @item ESS removed support for many unused languages.
 This includes old versions of S+, ARC, OMG, VST, and XLS.
 
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -2668,7 +2668,6 @@ the variable `ess-help-own-frame' is non-nil."
   :type 'alist
   :package-version '(ess . "18.10"))
 
-
  ; Faces
 ;;;=====================================================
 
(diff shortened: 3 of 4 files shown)


lisp/ess-r-mode.el
```diff
@@ -2260,7 +2260,40 @@ state.")
...
+
+(defun ess-r-help-add-links ()
+  "Add links to the help buffer."
+  (save-excursion
+    ;; Search for fancy quotes only. If users have
+    ;; options(useFancyQuotes) set to something other than TRUE this
+    ;; probably won't work. If it's FALSE, R outputs ascii ', but
+    ;; searching through the whole buffer takes too long.
+    (while (re-search-forward
+            (rx (minimal-match (and "‘" (1+ not-newline) "’"))) nil t)
```
vspinu: Can we not rely on rx? I think the rx package is ridiculous idea. Everyone knows regex syntax from whatever language they are coming from. rx just makes people learn another emacs-only-much-longer "syntax" to achieve the same thing. I find it pretty cryptic.

lisp/ess-r-mode.el
```diff
@@ -2260,7 +2260,40 @@ state.")
...
+      (let ((inhibit-read-only t)
+            (beg (match-beginning 0))
+            (end (match-end 0))
+            ;; Remove surrounding quotes so the text property is the
+            ;; name of the link:
+            (text (string-trim (match-string 0) (rx (0+ (or "‘" "\""))) (rx (0+ (or "’" "\""))))))
+        ;; Only add links for help topics we can find
+        (when (member text (ess-s-get-help-topics-function ess-local-process-name))
```
vspinu: Do we really need this check? I think occasional false positives are harmless but you can introduce false negative with this. 

The result of  `ess-s-get-helpt-topics-function` should be let-cached locally. It writes to dribble buffer each time it's called. 

I think the defcustom is not warranted either way.

jabranham:
> Do I understand correctly that the benefit of this is to be able to click with the mouse? 

That's /a/ benefit, but for me the main benefits are 1) being able to use TAB/backtab to navigate to the next/previous button and 2) the visual cue to users that they can click (or hit RET) to get help on the highlighted thing.


... I see now that I somehow didn't push the addition to the mode map for the TAB/backtab navigation. I'll push that now.

lisp/ess-r-mode.el
jabranham: I like the rx syntax. It keeps me from having to remember how many times I have to escape things since that varies quite a bit across languages. But I'll take it out.

lisp/ess-r-mode.el
jabranham: We get quite a few false positives without it since we catch anything that's in between fancy quotes. I guess we can try it out that way though during testing and see if it's too annoying.

I removed the defcustom.

lisp/ess-r-mode.el
```diff
@@ -2260,7 +2262,36 @@ state.")
...
+    (ess-display-help-on-object text)))
+
+(defun ess-r-help-add-links ()
+  "Add links to the help buffer."
+  (save-excursion
+    ;; Search for fancy quotes only. If users have
+    ;; options(useFancyQuotes) set to something other than TRUE this
+    ;; probably won't work. If it's FALSE, R outputs ascii ', but
+    ;; searching through the whole buffer takes too long.
+    (while (re-search-forward "‘.+?’" nil t)
```
vspinu: I think we can assume that help topics don't contain spaces. This might help with false positives. 

The non-greedy `+?` doesn't do much here I think.

jabranham: True, I'll add that in. 

We need the non greedy regex because otherwise if a line contains two matches we match from the beginning of the first to the end of the second. Given `foo 'bar' foo 'baz'` we'd buttonize `'bar' foo 'baz'` as one link.

jabranham:
OK to merge, right @vspinu ?

vspinu:
Yes, sure. Thanks!

lisp/ess-r-mode.el
```diff
@@ -2267,7 +2269,36 @@ state.")
...
+    ;; options(useFancyQuotes) set to something other than TRUE this
+    ;; probably won't work. If it's FALSE, R outputs ascii ', but
+    ;; searching through the whole buffer takes too long.
+    (while (re-search-forward "‘[^[:space:]]+?’" nil t)
+      (let ((inhibit-read-only t)
+            (beg (match-beginning 0))
+            (end (match-end 0))
+            ;; Remove surrounding quotes so the text property is the
+            ;; name of the link:
+            (text (string-trim (match-string 0) "[\"‘]*" "[\"’]*")))
```
dickmao: Line 2297 doesn't compile for me.   Perhaps
```
(text (replace-regexp-in-string "[\"']" "" (match-string 0))
```

jabranham: Thanks, fixed. string-trim didn't gain arguments until emacs 26.1.

vspinu:
BTW, I think underlying only the symbol between the quotes would look cleanner. Don't you think?

lionel-:
Ideally we'd remove the quotes from links. I noticed the elements wrapped with `\code{}` get the button treatment even though they are not links. I think that makes the docs hard to read. Could this be fixed?

vspinu:
Most `\code{` are existing symbols, isn't it?

jabranham:
The original PR checked that the link went somewhere. We could do that
again. It's slightly slower though and might not buttonize some things that
should be.

lionel-:
> Most \code{ are existing symbols, isn't it?

They also contain example code and refer to the parameters of the function being documented. I'd say most `\code{` are not links in the documentation I write.

vspinu:
If we tighten the regexp to match only alphanums and `.:_` we could improve on this situation a bit even without matching to an existing page. In either case there will be garbage like `NA`, `NULL` and `TRUE` which we probably don't want to highlight.

mmaechler:
> > Most \code{ are existing symbols, isn't it?
> 
> They also contain example code and refer to the parameters of the function being documented. I'd say most `\code{` are not links in the documentation I write.

The same for me.  We really should not hyperlink to `\code{..}` but to `\link{}`.    While some people use
`\code{\link{foo}}`  I (and @lionel- and quite a few others, incluing other R core members) I very often have more inside the `\code{..}` than just the `link{.}` ... and there are even example usages of `\link{}` *not* wrapped inside code, namely for cases where the help page is about a concept rather than an existing R object (function or data, ..).    Could we reconsider, and possibly become smarter here... consider maybe even providing a patch to R to optionally provide slightly more sophisticated text help? ... well I've started phantasizing... ;-)

Issue #731, opened 2018-10-26
Problem building rpm for OpenSUSE


dsteuer:
Hi! 

Packager for OpenSUSE here.
This release is somewhat more demanding than any release the last 10 years :-)

I have 18.10-1 and the Makefile for lisp/ provided by Martin.
(btw 18.10.1 would be much easier, at least for me!)

Nevertheless I still run into problems using my "good old spec file":

First: If I invoke the build service using my spec I get this:

```
[   13s] ./installation.texi:117: warning: node up `Check Installation' in menu `Installation' and in sectioning `General Information' differ
[   13s] readme.texi:8: node `General Information' lacks menu item for `Check Installation' despite being its Up target
[   13s] make[1]: [Makefile:117: html/readme.html] Error 1 (ignored)
[   13s] ##                   ^^^^^^^^^^^^^ (design bug: "FIXME" in ess.texi)
```

So it seems readme.html is not created and consequently

```
[   13s] cp -p html/ess.html html/readme.html html/news.html /home/abuild/rpmbuild/BUILDROOT/emacs-ess-18.10.1-1.x86_64/usr/share/doc/packages/emacs-ess
[   13s] cp: cannot stat 'html/readme.html': No such file or directory
```

the build stops.

Second: I wanted to debug locally and just invoke make by hand, but

```
steuer@gaia:~/OBS/devel:languages:R:released/emacs-ess/ess-18.10.1> make
cd lisp; make all
make[1]: Verzeichnis „/home/steuer/OBS/devel:languages:R:released/emacs-ess/ess-18.10.1/lisp“ wird betreten
emacs -batch -no-site-file -no-init-file --directory . -f batch-byte-compile ess.el
Eager macro-expansion failure: (error "‘tramp-file-name-structure’ didn’t match!")
Eager macro-expansion failure: (error "Variable binding depth exceeds max-specpdl-size")
>>Error occurred processing ess.el: error (("‘tramp-file-name-structure’ didn’t match!"))  
make[1]: *** [Makefile:99: ess.elc] Fehler 1
make[1]: Verzeichnis „/home/steuer/OBS/devel:languages:R:released/emacs-ess/ess-18.10.1/lisp“ wird verlassen
make: *** [Makefile:9: all] Fehler 2
```

And here i have no idea what happens. The same make works inside the build service, but make
always worked for older releases.

my emacs:

```
steuer@gaia:~> emacs --version
GNU Emacs 26.1
```

Any ideas?

Detlef

[...]

PR #751, opened 2018-11-16
`make dist` as singular dependency for downstream distributors

dickmao:
rpm and debian maintainers should only ever run `make dist` from their
clone of the upstream repository.  ESS maintainers should only
need to ensure `make dist` works (and produces a complete set of artifacts).

@dsteuer We need to know what `make` targets you use from our toplevel Makefile, and your `rpmbuild` incantation, generally.  You speak of a "good old spec file" in #731 .

@eddelbuettel We have good visibility into the debian build.  An unnecessary but "would be nice" modification in https://salsa.debian.org/edd/ess/merge_requests/1/diffs

Subsumes #707


eddelbuettel:
Actually, I tend to work off the tar.gz and I would be happy as a clam if that tar.gz were self-sufficient.  As I said before, I have reason to believe it is currently not self-sufficient what we can take those Debian aspect up over there.

dickmao:
Ah, using the tarball is something of a showstopper for this PR.
You don't use `gbp buildpackage`?  That seemed to work for me, but I don't have your experience.

eddelbuettel:
The work flow typically consists of importing a new upstream release, commonly a compressed tar file, into its branch first (most often via `gbp import-orig --uscan` which utilizes the `debian/watch` files and the `uscan` tool to locate and download the tarball) and to then build off the updated repo.  I do call `gbp buildpackage` but it also calls into `pbuilder` to have a "clean" and "pristine" empty machine alike (`pbuilder` is like an adapted `chroot` system from the pre-Docker days). It's been a while but I think that is a one-time setup to declare the (already existing on my box) `pbuilder` inside the `~/.gbp.conf`.   Maybe we should have that conversation over at salsa, it may not have too much relevance for anybody else here.  Appreciate you digging into it though.

Makefile
```diff
@@ -44,75 +44,11 @@ julia:
 autoloads:
 	cd lisp; $(MAKE) ess-autoloads.el
 
-## the rest of the targets are for ESS developer's use only :
-
-## --- PRE-release ---
-
-# Create .tgz and .zip files only
-# GNUTAR=gtar make tarballs
-tarballs: $(ESSDIR)
```
jabranham: Why remove the tarballs target?

Makefile
```diff
@@ -30,7 +30,7 @@ etc: version
 	cd etc; $(MAKE)
 
 .PHONY: test
-test: version
+test: version julia
```
jabranham: Why does julia get added here? AFAIR we don't test julia at all

dickmao:
> The work flow typically consists of importing a new upstream release, commonly a compressed tar file, into its branch first (most often via `gbp import-orig --uscan` which utilizes the `debian/watch` files

Okay, I'm unaware of a lot.  The `debian/watch` points to http://ess.r-project.org/downloads/ess which contains tarballs in all their resplendence (I excised the tarball making thinking they were largely historical).  I need to close this PR.

mmaechler:
> Okay, I'm unaware of a lot. ............ I need to close this PR.

Indeed, thank you.

For future reference:  The *main* release targets are the tarballs (now with their signature files) and the `make rel` target includes "uploading" them to the download area of  https://ess.r-project.org/
and to automagically update those web pages.

Even if that target is only used by me currently, we have put quite a bit of effort into it, if only to ensure nothing is forgotten.  (@jabranham and we still must fix something there such that git `tag` is not just pushed but specifically "pushed into master" (something I really don't quite get - one of the wonders of git ..).

PR #753, opened 2018-11-17
Makefile: pull versions from centralized location and automatic prerequisites generation

vspinu:
A set of 3, hopefully non-contentious, changes to Makefiles:

   - Pick ESS (and ESSR) version from ess.el from where package.el, melpa, pkg-build, debian-elpa etc. pick it
   - Simplify lisp/Makefile by automatically generating prerequisite files
   - Add ESSR target to build and commit new versions of ESSR


lisp/ess-roxy.el
```diff
@@ -68,7 +68,7 @@
 (require 'outline)
 (eval-when-compile
   (require 'cl-lib))
-(require 'essddr "ess-rd.el")
```
jabranham: This change has already been made in `next`

lisp/Makefile
```diff
@@ -56,28 +52,31 @@ uninstall:
...
 
 distclean: clean
-# and potentially more
 
-### Targets below here are only for developers - and these must have perl
+

+### File Targets
+
+.dependencies: $(ELS)
```
jabranham: Why is this necessary?

VERSION
```diff
@@ -1 +1,2 @@
-18.10.2
+!!! Auto generated by make. Update version in ess.el instead !!!
```
jabranham: We should just remove this version file. I think it's only needed by a couple doc files and we can update them with sed.

vspinu: It's a prerequisite for a few targets and it's not easy to get rid of but could be removed from git repo and ignored in .gitignore.

lisp/Makefile
```diff
@@ -56,28 +52,31 @@ uninstall:
...
+.el.elc:
+	$(COMPILE) $<
 
 ## Should happen before building ESS; definitely *NOT* after unpacking tarball :
 julia-mode.el:
 	test -f ../etc/.IS.RELEASE || $(DOWNLOAD) \
 	https://raw.githubusercontent.com/JuliaEditorSupport/julia-emacs/master/julia-mode.el > $@
 
-ess-julia.elc: julia-mode.elc
+ess-julia.elc: julia-mode.el julia-mode.elc
```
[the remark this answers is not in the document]
vspinu: Right. The old one wasn't right, the new one is not right either :)

lisp/ess.el
```diff
@@ -16,8 +16,9 @@
 ;; Maintainer: ESS Core Team <ESS-core@r-project.org>
 ;; Copyright (C) 1997-2018 ESS Core Team <ESS-core@r-project.org>
 ;; Created: 7 Jan 1994
-;; Version: 18.10.2
+;; Version: 18.10.3-snapshot
```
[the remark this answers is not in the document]
vspinu: Snapshot means " the build is a snapshot of your code at a given time". It's how things are done in java and clojure which I am familiar with but if -git is more suggestive let's do -git.

vspinu:
And as I said before, it's good to keep all metadata in one well formatted file like most of packages do. All build tools are able to pick it from ess.el because they auto-generate ess-pkg.el and that's good enough. For tar.gz for package-install-file we can generate ess-pkg.el from ess.el, that's an one-liner; there is no contradiction there, or do you see one? There is much more going on in the issue which you cited, and much more that you don't see in there. Would appreciate if could not bring it up again.

lisp/Makefile
vspinu: Auto-generates the pre-requisites for all our elc files. We don't need to hard-code them in our Makefile any longer.

lisp/ess-roxy.el
vspinu: Maybe this one should go on next actually, or after you merge next.

lisp/Makefile
vspinu: There is actually a more elegant way of doing that with .SECONDEXPANSION but it doesn't work on ancient make 3.8 on Travis' ubuntu trusty. So I gave up after a night of struggle.

vspinu:
> this is not where package.el would get it.

Let me attempt to put a period in this pkg.el dispute. Package.el does not mandate to keep NAME-pkg.el inside a repo. Packaging is unrelated to the organization of our code base. Its requirement is exclusive to the package archive which is defined as 

```
a multi-file package is stored in a package
archive as a tar file.  The tar file must be named ‘NAME-VERSION.tar’,
where NAME is the package name and VERSION is the version number.  
...
‘NAME-pkg.el’  ... must contain a single Lisp form, consisting of a call
to the function ‘define-package’,

```
The `define-package` form has only 3 arguments name, version, docstring and requirements. This means that we cannot add our other metadata there. So we will still have to keep other metadata in ess.el. Plus all lisp version vars (`ess-eversion`, `essr-version`, `ess-version` function etc) and probably much more metadata related tooling will be in the ess.el in the future. So it just doesn't make sense in my mind to keep this artificial split as part of the code base. All package build tools out there (package.el is not a build tool) auto-generate pkg.el and I think this is the right approach. So can we, if we really want to bypass build tools.  It's really common sense, not my subjective quirk.

lisp/ess.el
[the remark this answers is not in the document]
jabranham: Snapshot to me implies an intended release, sort of like a release candidate. I don't have a strong feeling about it though. `version-regexp-alist` treats `-snapshot` and `-git` the same I think.

VERSION
vspinu: Changelog, dist, essr and one RPM related target depend on VERSION file changes. We need a marker of some kind to keep track of the old version. I don't think there is a more elegant solution to this, but I am not a make expert.

lisp/Makefile
jabranham: Sorry, I wasn't clear. I was thinking more along the lines of "why is it necessary to specify .elc dependencies in the first place?" If .elc doesn't exist yet, the byte compiler will just use the corresponding .el file. Can't we just say "all .elc files depend on the corresponding .el file" and not care about the order?

vspinu: But if one .el changes you need to recompile all els which depend on it. Or you wonder something else?

jabranham: Ah, I knew I was missing something obvious. I wasn't thinking about non-clean checkouts.

lisp/ess-roxy.el
```diff
@@ -938,7 +938,7 @@ passes ARGS to it."
            (ess-roxy-goto-end-of-roxy-comment)
            (looking-at "#")))
     (ess-roxy-with-filling-context t
-        (apply orig-fun args)))
+                                   (apply orig-fun args)))
```
jabranham: This whitespace doesn't look right.

vspinu:
I am pulling this in. I changed the ground truth of version back to VERSION file. There are too many uses of it in texi files and sed is not the right way to approach it there.

PR #761, opened 2018-11-20
Move a few files to lexical binding

jabranham:
The changes aren't as big as github makes them look. I removed a `let` binding in inferior-ess which changes the indentation for a large part of that big function.

I think the biggest change is `inferior-ess-make-comint`. I think the way the code was written we'd never reach one part of the if statement that calls that function so I collapsed all that (hence the removed let binding). Seems to work on my end.

Other than that, it's mostly cleaning up unused variables and arguments.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1,4 +1,4 @@
-;;; ess-inf.el --- Support for running S as an inferior Emacs process
+;;; ess-inf.el --- Support for running S as an inferior Emacs process  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 1989-1994 Bates, Kademan, Ritter and Smith
 ;; Copyright (C) 1997-1999 A.J. Rossini <rossini@u.washington.edu>,
@@ -58,12 +58,6 @@
 
 (declare-function ess-mode "ess")
 (declare-function ess-complete-object-name "ess-r-completion")
-
-;; TODO: refactor and remove file-local variable
-;; byte-compile-warnings. See ess-r-mode.el also
-(defvar proc)
-(defvar start)
-(defvar end)
 
 (defvar inferior-ess-mode-syntax-table
   (let ((tab (copy-syntax-table comint-mode-syntax-table)))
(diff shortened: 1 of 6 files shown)


lisp/ess-inf.el
```diff
@@ -512,11 +485,9 @@ This marks the process with a message, at a particular time point."
      (format "\nProcess %s %s at %s\n"
              (process-name proc) message (current-time-string)))))
 
-(defun inferior-ess-make-comint (bufname procname infargs &rest switches)
+(defun inferior-ess-make-comint (bufname procname switches)
   "Make a comint process in buffer BUFNAME with process PROCNAME.
```
vspinu: Which if did you remove? `inferior-ess-make-comint` doesn't have any if removed.

lisp/ess-inf.el
```diff
@@ -1321,16 +1284,7 @@ Example (ess-boolean-command \"2>1\n\")"
     (ess-kill-last-line)
     (buffer-substring (point-min) (point-max))))
 
-;; (ess-async-command "{cat(1:5);Sys.sleep(5);cat(2:6)}\n" nil (get-process "R")
```
vspinu: could you please move these commetns to inf test file?

lisp/ess-inf.el
jabranham: This is much easier to see on a local git checkout. Not sure why github is rendering it so poorly.

In `inferior-ess`, there's a line `(setq buf (if switches ...` and I removed the "then" part of the if statement, keeping the "else" part.

lisp/ess-inf.el
jabranham: ess-async-command seems to return a buffer. Is that right?

lisp/ess-inf.el
vspinu: Ah, I see. Indeed you refactored that infarg argument. Sure, looks right.

lisp/ess-inf.el
vspinu: Yes, that's correct because `ess-command` returns the buffer. Sometimes consumers of `ess-command` need to investigate the output buffer. Less useful for async though.

jabranham: Doesn't seem to work for me:

```
(ess-async-command "{cat(1:3);Sys.sleep(1);cat(4:5)}\n" (get-buffer-create "test") nil
                         (lambda (proc string) (message "done")))
```

results in the "test" buffer getting "1 2 3" but not the "4 5".

vspinu: :) The test buffer returns immediately and there is only a bit of output there, for async commands that buffer should be used for errors or something, not for output. The output comes in the callback "string" argument. 

Try

```el
(ess-async-command "{cat(1:5);Sys.sleep(5);cat(2:6)}\n" nil (get-process "R")
                   (lambda (proc string) (message "done %s" string)))
```

You should get `done 2 3 4 5 6> ` message in 5 seconds. 

This async commands are not really async in R sense (obviously). It's async in the sense that it's non-blocking on elisp level, unlike ess-command. A poor man pathetic attempt ... but what can one do :/. I am using it in polymode though.

jabranham: Thanks. I was misunderstanding how the function worked. I just added a test. Will merge after making sure it doesn't break on travis.

PR #763, opened 2018-11-22
Compilation tests

dickmao:
I am unaware of a `make compile` target.  As a C dev, I continually `make compile` after a string of changes to ensure sanity.  Likewise, the unconventional incantation in `lisp/Makefile` has served me in good stead in the past for developing elisp.

@jabranham I agree the incantation isn't the "best practices" way to `make compile`.  Can you help?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ etc: version
 
 .PHONY: test
 test: version
+	cd lisp; $(MAKE) test
 	cd test; $(EMACS) --script run-tests
 
 generate-indent-cases:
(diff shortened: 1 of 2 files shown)


vspinu:
At some point we might want to use [elisp-lint.el](https://github.com/gonewest818/elisp-lint) which does compilation, checkdoc and linting. 

We have compilation check for emacs 26+ and e25 is probably not worth it. It's surprising that compilation error on e25 went unnoticed though.

jabranham:
`make lisp` top-level or `make all` in lisp/ compiles all the lisp files. 

The error was a sneaky one. Because string-trim is an inline function, the elc file was still produced so make didn't end with a nonzero exit code.

I'm working on fixing the remaining compiler warnings in Emacs 25 and will turn on byte-compile-error-on-warn on travis after that's fixed.

PR #764, opened 2018-11-22
Cask (in conflict with #762)

dickmao:
To effect a self-contained archive amenable to `package.el`, we should use the standard tools.  While some deem `cask` to be nonstandard, it is in my opinion our best option.  We use it to great effect at [this FOSS library](https://github.com/millejoh/emacs-ipython-notebook).

See discussion #762


lionel-:
Thanks, but I think you should have suggested this alternative in #762 instead, before sending a new PR. Would you mind closing this, suggest this course of action in the ongoing PR, and wait for the opinion of Alex and Vitalie?

dickmao:
I am fine with closing it.  But as a software developer you should well know that arguing the case with prose, e.g., "how about this?" is far less effective than arguing with code in the form of a PR.

lionel-:
And as an open source developer you should know that it's a bit overwhelming getting all these pull requests in the midst of several ongoing discussions. Also Alex is already working on this, so it's a bit aggressive to open a pull request that closes his. If you'd like to show what it'd look to depend on Cask, you could open a pull request in your own repo and link to it (but now we have this one, so no need for this). Or open a pull request to Alex's branch.

vspinu:
Indeed, we appreciate all the effort and time spent. You are really seriously into helping. But we are used to more incremental and isolated changes. We are all here very strong on simplicity, so every massive change like this should be conferred with us first because we might already have an opinion on that.

The case of cask is far from clear cut. It's an extra dependency and you have added 4 new files with non-trivial bash scripts. The managerial taks is a bit too high as all of the devs will have to get familiar with cask and understand those scripts. Granted that most of those are buirocrasy in `targets`, but still ...

Cask  is clearly not warranted for the heavy-duty task of managing our only dependency julia-mode.el.  
But the `package` function is a different story. It's indeed nice that we can declare package-files and final structure in a tiny MELPA like config and build the package with a single command. For this reason I completely differ to @jabranham to decide on this. If codding our own build system turns to be more than 50 lines of elisp, using cask might be a better idea indeed. The PR is very clean AFAIC.

dickmao:
Well, I am gratified that at least you see the economy of code that accomplishes the packaging.
You can also probably see the bash scripts are not required for the task at hand.  I included them for completeness.

dickmao:
I've gone ahead and removed the scripts and the travis changes from the PR to make it blazingly clear how `cask` can facilitate the task at hand.

vspinu:
Thanks, it's more vendible like this indeed. Qualifying, that if we end up using this will need the install script as we don't expect our users to have cask installed. I guess this is the biggest tradeoff here.

jabranham:
I don't think cask is the way to go. We're very light on dependencies right now (just Emacs and julia-mode) so cask doesn't help much there. Producing the .tar file from the package function is nice, but I think we're going to need more flexibility than cask offers since we're planning on distributing ESS and ESS+ (#732). Finally, I am uncomfortable with adding python and then adding a binary to $PATH as a prerequisite for packaging ESS. We should try to keep ESS as lean as possible, within reason of course.

I'm willing to be convinced otherwise, though.

Issue #553, opened 2018-05-25
srcref jumping with dangling srcfiles


lionel-:
When a package is installed from a transient location (as would be the case when installing from CRAN or github), the source files recorded in the srcrefs no longer exist and this prevents xref jumping.

As it turns out, the srcrefs contain a full copy of their original file:

```r
srcref <- attr(fn, "srcref")
srcfile <- attr(srcref, "srcfile")
srcfile$lines
```

We could first check if `srcfile$filename` points to an actual file. If not the case, we'd walk over all functions in the namespace to collect all `srcfile` environments and then regenerate a copy of the source in a temporary location. Further jumps would redirect there.

What do you think @vspinu @atheriel ?

lionel-:
Also Tomas is working on fixing the overhead caused by installing packages with srcrefs. His goal is to make `options(keep.source = TRUE)` the default which would make the mechanism above immediately useful to all users.

vspinu:
Regenerate full namespace is an overkill (plus, sources of some functions can change and keeping a cache asks for trouble). 

Second step should be (and is already there) to look into `ess-r-package-library-paths` for the package with the same name, if not found a temporary copy of the source should be created through dump-to-file functionality (not yet implemented; was my task but I am slacking :/).

lionel-:
I don't think it's an overkill because it's helpful to navigate files when you're exploring a code base. I think we'll want this functionality to work for all functions and not just those installed locally, especially if srcrefs get installed by default.

Cache invalidation is easy: srcfile envs are time-stamped.

lionel-:
Only the `src/` folder would be missing, all the rest would be there (the whole `R/` folder, `NAMESPACE`, `DESCRIPTION`). I'd say it's more than a half-measure and would bring the source jumping experience close to emacs lisp.

lionel-:
I really want this to just work. Maybe we should pull the whole CRAN package as you suggest but that would result in a noticeable delay.

vspinu:
> the whole R/ folder, 

Are you saying that the whole R file could be retrived 100% identical with the original source? I didn't know that. In that case, it might be worth a try indeed.

lionel-:
Yes the srcrefs of a function have lines number + a reference to a sourcefile environment which contains the whole file (in the `lines` field). I didn't know about this either, though in retrospect that should have been obvious since roxygen is based on srcrefs.

vspinu:
[...]

PR #765, opened 2018-11-22
Reappeal ess-r-source-mode to make xref useful (was #475 xref backend for R)

dickmao:
From #475 

> > If you just remove the ess-r-source-mode I think this is good to merge!
> 
> I have removed this code, as you both requested. However, I'll make one final plea for its inclusion:

Not only do I see a real difference between `ess-r-source-mode` and `ess-dump-object-info-edit-buffer` in both function and intent, I also don't see how they could share code.  My experience has been that @atheriel's original PR works exactly how I'd want it -- the source viewing buffers are read-only and buried via `q`.  I can `M-.` and `M-,` to my heart's content.  

Other changes include a preliminary `test_that` for ESSR and setting `ess-r-package-library-paths` to a `libPaths()` call.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -1501,6 +1501,9 @@ line starts with the roxy prefix.")
 
 ;; SJE -- this should not be defcustom - user does not set it.
 (defvaralias 'ess-current-process-name 'ess-local-process-name)
+
+;; should this have been make-buffer-local-variable?
+;; we always setq-local `ess-local-process-name' per buffer
 (defvar ess-local-process-name nil
   "The name of the ESS process associated with the current buffer.")
 (put 'ess-local-process-name 'risky-local-variable t)
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -163,9 +163,13 @@ into a list and call REPORT-FN on it."
        for msg = (match-string 4)
        for (beg . end) = (let ((line (string-to-number (match-string 1)))
                                (col (string-to-number (match-string 2))))
-                           (flymake-diag-region src-buffer line col))
+                           (if (fboundp 'flymake-diag-region)
+                               (flymake-diag-region src-buffer line col)
+                             (error "flymake-diag-region unbound (emacs-26 required)")))
        for type = (ess-r--flymake-msg-type (match-string 3))
-       collect (flymake-make-diagnostic src-buffer beg end type msg)
+       collect (if (fboundp 'flymake-make-diagnostic)
+                   (flymake-make-diagnostic src-buffer beg end type msg)
+                 (error "flymake-make-diagnostic unbound (emacs-26 required)"))
        into diags
        finally (funcall report-fn diags)))))
 
@@ -182,7 +186,9 @@ REPORT-FN is flymake's callback function."
            (not (ess-process-live-p)))
       (progn
         (funcall report-fn nil)
-        (mapc #'delete-overlay (flymake--overlays)))
+        (if (fboundp 'flymake--overlays)
+            (mapc #'delete-overlay (flymake--overlays))
+          (error "flymake--overlays unbound (emacs-26 required)")))
     (let ((src-buffer (current-buffer)))
       (setq ess-r--flymake-proc
             (make-process
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -235,7 +235,6 @@
     table)
   "Syntax table for `ess-r-mode'.")
 
-
 (defvar ess-r-completion-syntax-table
   (let ((table (copy-syntax-table ess-r-mode-syntax-table)))
     (modify-syntax-entry ?. "_" table)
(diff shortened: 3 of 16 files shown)


lisp/ess-r-flymake.el
```diff
@@ -163,9 +163,13 @@ into a list and call REPORT-FN on it."
        for msg = (match-string 4)
        for (beg . end) = (let ((line (string-to-number (match-string 1)))
                                (col (string-to-number (match-string 2))))
-                           (flymake-diag-region src-buffer line col))
+                           (if (fboundp 'flymake-diag-region)
```
lionel-: Could you make unrelated changes in other PRs please?

lisp/ess-r-mode.el
```diff
@@ -2312,6 +2311,44 @@ state.")
...
+  (let ((inhibit-read-only t))
+    (setq header-line-format
+          (concat "Environment: " 
+                  (ess-r-source--parse-environment (current-buffer))
+                  (and (ess-r-source--parse-byte-compiled-p (current-buffer))
+                       " (byte-compiled)")))))
+
+(defun ess-r-source--parse-environment (buff)
+  "Determine the environment of the R function printed in BUFF.
+
```
lionel-: We don't use newlines after the first line. Can you please remove all of them?

lisp/ess-custom.el
```diff
@@ -1501,6 +1501,9 @@ line starts with the roxy prefix.")
 
 ;; SJE -- this should not be defcustom - user does not set it.
 (defvaralias 'ess-current-process-name 'ess-local-process-name)
+
+;; should this have been make-buffer-local-variable?
+;; we always setq-local `ess-local-process-name' per buffer
```
lionel-: Since you are not a core developer, can you ask questions in an issue or github comment rather than in the source code?

dickmao:
> Thanks. This is well written code...

Much of it is @atheriel 

> With this change, xref basically becomes useless for me.

I don't doubt your claim, but I don't know why.  `master` is only practical for your workflow of xref'ing your own package (it is not practical for the general case of inspecting outside packages).  I tried to preserve that.

> R is likely going to enable srcrefs by default and the long term plan is to reconstruct package directories from those, when local sources are available.

I don't doubt that claim, but I'd rather not wait for it.  Before I got into/created the packaging mess, I spent a couple days poring over the (very verbose)  discussion in #475 and #553 .  I tried to make the srcref chicanery work (and yes, chicanery is the right word as you know from #553).  It'd be hard work to tastefully get the user to configure all the right variables to download the srcrefs for *all* the tokens he might `M-.`  It would result in duplicated sources (one for the original libraries in /usr and one for the downloaded keep-source libraries in his home directory).  But worst of all, it results in the same outcome as dumping the function definitions.  In other languages, you see the context of the meta-dotted function.  In R, you just see the function (*with comments*) because R convention is to write one function per file (like matlab).

> You are introducing a new major mode instead of using the dump-object mechanism. As we have discussed it last time, we think the existing mechanism should be used, and perhaps improved.

Yes, I went over this argument with @atheriel in #475.  I can only regurgitate what he said -- that to a pair of fresh eyes the "edit function" and "read function" are sufficiently different in function and intent, and that the `ess-r-source-mode` which he derived from the emacs `special-mode` -- a mode specifically designed for quick one-off reads -- is highly apt.

Now I am guilty of the same verbosity I dreaded in #475.  The xref functionality is much desired by any ESS user who knows about `M-.`  This is probably a low percentage but still numbers in the hundreds.  As it stands, the configuration demands on the user are way too high for any kind of adoption.

lionel-:
> is only practical for your workflow of xref'ing your own package 

It is practical for any package you have installed locally. Long term plan is to make it practical for all packages.

>  get the user to configure all the right variables to download the srcrefs for all the tokens

I don't understand what you mean.

> R convention is to write one function per file

I am not aware of such a convention.

> the "edit function" and "read function" are sufficiently different in function and intent

I disagree. There's no advantage in making the dump buffer artificially read-only. Besides what's the advantage over just printing the function in the console? More importantly, xref is for jumping to source location, not for dumping into a standalone buffer.

dickmao:
Okay, I see your position hasn't change since May.  I am closing this PR.

mmaechler:
> > R convention is to write one function per file
> 
> I am not aware of such a convention.

It is 100%  non-sense (put *politely* !) !     ... and don't go for "the average package" -- rather e.g. base R, recommended packages, other pieces which have stood the time  --  but even with the average package I guess,  your claim would be  wrong by orders.

@dickmao : We can't even know who you are (as you've decided to remain anonymous on GH).
What are your credentials  in Statistics and in R ?
Reading  statements  as the one above (plus an attitude of being radical and pretty carefree instead of considerate and  cautious) make me wonder...

I'm sorry I've felt to be blunt... but I'd rather enjoy my work, including on ESS., so I do hope you can adapt to how we collaborate here.

atheriel:
@dickmao FWIW, although I'm flattered that you agree with me, ultimately I think that @lionel- and the other maintainers deserve our deference in what is or is not going to become part of ESS. Making decisions is a key part of their role, after all.

A more constructive route here would have been to open an issue asking if the views expressed in earlier discussion had changed before embarking on a PR. In this case, maybe this feature could be released as a separate package first to gauge interest.

> if you hypothetically suspended your biases

Some unsolicited advice: I have **never** seen anything good come of attacking the character of other developers. It's hard not to feel hurt when your contributions (or arguments) are rejected, but that's not a good reason to lash out.

dickmao:
@atheriel Thank you for trying to make `xref` happen.

In the week that I've been involved in the project I don't think I've said anything unprofessional, certainly not commensurate to the comments levelled against me.  The line regarding biases I concede is borderline.

vspinu:
As an observer from the inner side, I back it up, you haven't said anything unprofessional or attacked anyone personally. You are just not a good salesman. You want too many things and all at once. :wink:. Anyways, back to the PR ...

As always there are many unrelated to the main issue fixes which are good. I would like to cherry pick and pull those like last time. Would you mind? 

Regarding the core issue @lionel- pointed pretty much everything. I would just add here that that hole was left intentionally as a commitment device to all of us. First, by not having 100% functional xref users are forced to discover `ess-r-package-library-paths` which otherwise might be simply unnoticed. Second, frustrated devs like you (and me and @lionel- and ...) will bring PRs which will eventually land in master and do something much more useful than a simple `dump`. 

Till @lionel-  brings along his reconstruction idea to life (I still think it's pretty crazy, but well, it's his job to prove me wrong :stuck_out_tongue_winking_eye: ) we can do fetching CRAN and git archives into a dedicated directory and unpack those there (also @lionel-'s original idea if I am not mistaken). If you want to help with xref this is where you can really help.

Even more, I have recently pushed through AVFS support to CIDER (see end tip [here](https://cider.readthedocs.io/en/latest/interactive_programming/)) and I couldn't be happier with it. In a nutshell, on a xref you land inside a jar archive and all emacs works there as in ordinary file system (including non-trivial stuff like projectile, grep etc). Something along those lines would be nice here as well, but not a big deal though.

dickmao:
> I would like to cherry pick and pull those like last time. Would you mind?

That'd be swell, and appreciate your vouching for my online behavior.

PR #773, opened 2018-11-29
Remove inferior-*-start-file

jabranham:
I don't think it's respected anyway since the customize-alists override it with `nil` in almost all cases.


jabranham:
> Not in case of STATA files.

Right, that's why I said "almost" :-)

> Why removing and not fixing?

Users can achieve the same thing by using `ess-load-file` and the various `-post-run-hook`s. I could add a comment about this to the manual, if needed?

vspinu:
True, though it requires some, admittedly trivial, lisp. I am indifferent about this feature and would be ok with removing it if others don't mind.

lionel-:
I think simplifying the internals and the interface of ESS and teaching users about ESS hooks and functions in the manual is a win-win.

jabranham:
I've added some documentation to the manual about how to do that. I'll merge this soonish unless someone objects.

PR #774, opened 2018-11-29
Improve testing

jabranham:
A few improvements: 

- Provide a way to mimic travis erroring on compile warnings locally (@vspinu asked for this)
- Eliminate byte compile warnings on Emacs 25
- Simplify travis script

I've moved the git-snapshot travis build to an allowed failure for the moment. The EVM build is 9 months old and if-let has been un-obsoleted since then. I've raised an issue over there and will move it back to required successes once the git snapshot gets updated.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -59,6 +59,13 @@
 (declare-function ess-mode "ess")
 (declare-function ess-complete-object-name "ess-r-completion")
 
+;; The following declares can be removed once we drop Emacs 25
+(declare-function tramp-file-name-method "tramp")
+(declare-function tramp-file-name-user "tramp")
+(declare-function tramp-file-name-host "tramp")
+(declare-function tramp-file-name-localname "tramp")
+(declare-function tramp-file-name-hop "tramp")
+
 (defvar inferior-ess-mode-syntax-table
   (let ((tab (copy-syntax-table comint-mode-syntax-table)))
     tab)
--- a/test/Makefile
+++ b/test/Makefile
@@ -3,7 +3,7 @@ EMACS?=emacs
 .PHONY: literate
 
 
-all:
+all: compile
 	${EMACS} -Q --script run-tests
 
 ess:
@@ -29,3 +29,8 @@ indent-cases:
 
 literate-cases:
 	${EMACS} -Q --script generate-literate-cases
+
+.PHONY: compile
+compile:
+	@cd ../lisp && make -k all \
+		"EMACSBATCH = emacs --batch --eval \"(setq byte-compile-error-on-warn t)\""
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ env:
   - EVM_EMACS=emacs-26.1-travis
   - EVM_EMACS=emacs-git-snapshot-travis
 
+matrix:
+  allow_failures:
+    - env: EVM_EMACS=emacs-git-snapshot-travis
+
 before_install:
   - echo "deb https://cloud.r-project.org/bin/linux/ubuntu trusty/" | sudo tee -a /etc/apt/sources.list
   - sudo add-apt-repository ppa:marutter/c2d4u -y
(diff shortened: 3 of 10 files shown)


jabranham:
Though perhaps `make test` should first run `make clean` in lisp/ to make sure all files are re-byte compiled? WDYT?

vspinu:
> in lisp/ to make sure all files are re-byte compiled? WDYT?

I don't think this is necessary. make should figure out obsoleted dependencies and re-build only those modified.

test/Makefile
```diff
@@ -29,3 +29,8 @@ indent-cases:
 
 literate-cases:
 	${EMACS} -Q --script generate-literate-cases
+
+.PHONY: compile
+compile:
+	@cd ../lisp && make -k all \
```
vspinu: Let's add `compile` in lisp sub-folder for consistency.

lisp/ess-inf.el
```diff
@@ -59,6 +59,13 @@
 (declare-function ess-mode "ess")
 (declare-function ess-complete-object-name "ess-r-completion")
 
+;; The following declares can be removed once we drop Emacs 25
+(declare-function tramp-file-name-method "tramp")
+(declare-function tramp-file-name-user "tramp")
```
vspinu: Cannot we just require tramp?

test/Makefile
jabranham: Done

lisp/ess-inf.el
jabranham: No, it takes too long to load if we don't need it (see the commit message in 30a3acf7e18f87ee7541275d2e7198769f1c633e)

vspinu:
Now tests fail for me because of the julia's warnings

```
julia-mode.el:681:10:Error: reference to free variable ‘julia-syntax-propertize-function’
Makefile:69: recipe for target 'julia-mode.elc' failed
make[2]: *** [julia-mode.elc] Error 1
```

Also, shall we not use `-k` on the local tests? It's inconvenient to scroll through all the output. Or even better suppress those loading messages somehow:

```
emacs --batch --eval "(setq byte-compile-error-on-warn t)" --directory . -f batch-byte-compile ess-gretl.el
Loading 00debian-vars...
Loading /etc/emacs/site-start.d/50autoconf.el (source)...
Loading /etc/emacs/site-start.d/50cmake-data.el (source)...
Loading /etc/emacs/site-start.d/50dictionaries-common.el (source)...
Loading debian-ispell...
Loading /var/cache/dictionaries-common/emacsen-ispell-default.el (source)...
Loading /var/cache/dictionaries-common/emacsen-ispell-dicts.el (source)...
```

test/Makefile
```diff
@@ -29,3 +29,8 @@ indent-cases:
...
+		"EMACSBATCH = emacs --batch --eval \"(setq byte-compile-error-on-warn t)\""
```
vspinu: I don't think we want `emacs` here.

vspinu:
> Or even better suppress those loading messages somehow:

Not a big deal though. Those messages show on system emacs only. On emacs from source it shows clean trace.

test/Makefile
vspinu: :) `$(EMACS)`. It's done in master.

vspinu:
> Now tests fail for me because of the julia's warnings

One e27 I am also seeing 

```
In toplevel form:
obsolete/ess-r-args.el:224:23:Error: ‘ess-r-args-current-function’ is an obsolete function (as of ESS 18.10); See ‘eldoc-mode’
Makefile:72: recipe for target 'obsolete/ess-r-args.elc' failed
make[2]: *** [obsolete/ess-r-args.elc] Error 1
```

I have added `test-lisp` target to overcome such temporary glitches but I wonder if we should keep `make test` for lisp tests only (as before) and `make test-all` do the compile + lisp tests. I am finding it slightly annoying and unusual to type `make test-lisp` all the time during the development, plus the compile doesn't leave the repo in clean state. I think one should run `make test-all` once before push to origin and use `make test` through the development. Objections?

jabranham:
I'd prefer that `make test` runs all the tests we know about, including the compiler warnings. But I'll live if people want to change it :-)

I've noticed those obsolete compiler warnings before and have wondered why we don't see them on Emacs < 27. We can either A) fix them or B) remove ess-r-args. It was already obsolete one release and eldoc is better, so I'm fine with removal.

For the julia warnings, we should have make test run make julia-mode.elc in lisp before compiling our files.

PR #775, opened 2018-12-01
Support ess-load-library outside of R

### LINKED ISSUE #299: make `ess-load-library` generic
If possible, it would be nice to make `ess-load-library` generic, so that I could define it for Julia. Currently it checks `ess-dialect` and aborts if it is not R.

I understand that something more complicated than a simple string expansion (like `ess-setwd-command` for `ess-set-directory`) would be required. Possibly `ess-load-library` could call a `ess-load-library-function` when not `nil`. If would be happy to submit a PR, but not sure how to organize the code, since prompting with completion assumes a list of packages (eg if it is nil, there could be no completion for some dialects). A separate function could govern completion, but maybe that is getting to be too complicated.

In Julia, I would use `Base.reload`, with the assumption that one uses this feature during interactive development.

jabranham:
New generics ess-installed-packages and ess--load-library provide
support for ess-load-library, which is moved from ess-r-mode.el to
ess-inf.el

Closes #299

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1340,6 +1340,27 @@ of `ess-async-command' with an explicit interrupt-callback."
                        (ess-async-command ,com ,buf ,proc ,callback ',int-cb)))))
     (run-with-idle-timer delay nil com-fun)))
 
+(defun ess-load-library ()
+  "Prompt and load dialect specific library/package/module.
+Note that add-ons in R are called 'packages' and the name of this
+function has nothing to do with R package mechanism, but it
+rather serves a generic, dialect independent purpose. It is also
+similar to `load-library' Emacs function."
+  (interactive)
+  (let ((ess-eval-visibly-p t)
+        (packs (ess-installed-packages))
+        pack)
+    (setq pack (ess-completing-read "Load" packs))
+    (ess-load-library--override pack)
+    (ess--mark-search-list-as-changed)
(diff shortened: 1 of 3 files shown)


lisp/ess-inf.el
```diff
@@ -1340,6 +1340,28 @@ of `ess-async-command' with an explicit interrupt-callback."
...
+        pack)
+    (setq pack (ess-completing-read "Load" packs))
+    (ess--load-library pack)
+    (ess--mark-search-list-as-changed)
+    (display-buffer (buffer-name (process-buffer (get-process ess-current-process-name))))))
+
+(cl-defgeneric ess-installed-packages ()
+  "Return a list of installed packages.")
+
+(cl-defgeneric ess--load-library (pack)
```
vspinu: Shouldn't we use the same convention `--override` here as well?

lisp/ess-inf.el
```diff
@@ -1340,6 +1340,28 @@ of `ess-async-command' with an explicit interrupt-callback."
                        (ess-async-command ,com ,buf ,proc ,callback ',int-cb)))))
     (run-with-idle-timer delay nil com-fun)))
 
+(defun ess-load-library ()
+  "Prompt and load dialect specific library/package/module.
+
```
vspinu: No extra line please. Needless fragmentation of the beautiful lisp code :wink:

PR #776, opened 2018-12-01
Simplify package info retrieval

vspinu:
Simplifications of pkg-info retrieval. Basically make `ess-r-package-info` the only work-horse which retrieves and sets `ess-r-package--project-cache`. It was quite confusing that we were manipulation the cache in multiple places.  

The adaptation to `package.el` confused things a bit. Now we have two functions with separate concerns `ess-r-package-project` and `ess-r-package-info`. 

(overwrites @lionel-'s recent fixes)

### DIFF
--- a/lisp/ess-r-package.el
+++ b/lisp/ess-r-package.el
@@ -56,10 +56,9 @@ See also `ess-r-set-evaluation-env' and `ess-r-evaluation-env'."
   :group 'ess-r-package
   :type 'boolean)
 
-(defvar-local ess-r-package--project-cache nil
+(defvar-local ess-r-package--info-cache nil
   "Current package info cache.
-Cons cell of two strings. CAR is the package name active in the
-current buffer. CDR is the path to its source directory.")
+See `ess-r-package-info' for its structure.")
 
 (define-obsolete-variable-alias 'ess-r-package-library-path 'ess-r-package-library-paths "v18.04")
 (defcustom ess-r-package-library-paths nil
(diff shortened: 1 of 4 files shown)


lionel-:
Looks good, thanks. I would just make `ess-r-package-info` an internal variable because we might want to add more fields to it, like dependencies and such.

lisp/ess-r-package.el
```diff
@@ -102,33 +101,32 @@ all source dirs recursively within the current package.")
...
+  "Get the description of the R project in directory DIR.
+Return a cons of the form (pkg-name . pkg-path). This value is
+cached for efficiency reasons."
+  (if (and (null dir) (car ess-r-package--project-cache))
+      ess-r-package--project-cache
+    (let* ((pkg-path (ess-r-package--find-package-path (or dir default-directory)))
+           (pkg-name (when pkg-path
+                       (ess-r-package--find-package-name pkg-path)))
+           (pkg-info (cons pkg-name pkg-path)))
+      ;; don't cache if DIR was supplied
```
lionel-: (Did we settle on sentence case for comments?)

Because `dir` might not be the buffer-local directory right? Perhaps slightly better would be to cache based on expanded directory, though that doesn't seem very important. Perhaps just make the comment clearer about the choices involved?

vspinu: Hash seems more trouble here because it would need to be cleared with some logic and deal somehow with stale paths. With buffer locals the user can revert the buffer and reset the cache.

vspinu:
> I would just make ess-r-package-info an internal variable. 

How about making it return a list instead of cons? Then we can add new entries at the end without breaking compatibility. This is such a fundamental and useful functionality that making it internal doesn't look right.

lionel-:
Probably better as an alist indexable by symbol then?

vspinu:
Good idea, but it's unfortunate that we cannot use alist-get. Shall we create a compat `alist-get` if it's not defined (emacs25)?

lionel-:
I thought we were dropping support for emacs24? Otherwise yes, a compat sounds good.

PR #779, opened 2018-12-04
julia beginning/end defun functions, allowing movement/eval-function

nverno:
Adds versions of beginning-of-defun/end-of-defun for julia so functions like
ess-eval-function should work properly. Doesn't recognize anonymous functions, eg.

```
x -> ...
(x, y) -> ..
```
since I couldn't think of a reason to. `julia-mode` doesn't have a notion of `forward-sexp` so 
it is a bit hacky, but seems to work. Ignores nested functions, evaluation outermost function bodies.


nverno:
hey, yea I got it, I'll try to do that soonish. Making generic seems like a good idea. The functions I wrote should work for cases like

```
func(x, y) = x +
     y
f1(x, y) =
    x - y

func2(x, y, ...) = x * y + x

function factorial1(x)
    
    function fact()
        if (n == 0)
            return 1
        else
            n = n - 1
            return (n+1) * fact()
        end
    end

    n = x
    return fact()
end

```

jabranham:
> I don't care about being on the copyright or whatever so just consider my contribution as do whatever the hell you want with it.

Unfortunately, that's not enough for the FSF/Emacs. They want formal copyright assignment. It's usually a fairly painless process and can usually be done entirely over email.

PR #782, opened 2018-12-04
Support function navigation in julia buffers

### LINKED ISSUE #777: Julia functions are not recognized by ess-function-pattern
Julia functions are not recognized when running `ess-eval-function` with the cursor inside the function

```julia
function test(a::Int64)
    @show a
end
```
results in 
```
Point is not in a function according to `ess-function-pattern`

### PR DESCRIPTION
**Pull request by jabranham**
ess-beginning-of-function and ess-end-of-function now allow generic
overrides

Closes #777, closes #779

Co-authored-by: nverno <noah.v.peart@gmail.com>

@nverno - I combined our ideas for julia function navigation. WDYT?

### DIFF
--- a/lisp/ess-julia.el
+++ b/lisp/ess-julia.el
@@ -364,6 +364,52 @@ It makes underscores and dots word constituent chars.")
 
 (defvar ess-julia-mode-syntax-table (copy-syntax-table julia-mode-syntax-table))
 
+(defvar ess-julia-function-pattern
+  ;; (rx (or "function"
+  ;;         "macro"
+  ;;         (seq (0+ nonl) "(" (1+ nonl) ")" (0+ whitespace) "=")))
+  "function\\|macro\\|.*(.+)[[:space:]]*="
+  "Regular expression to match the beginning of a function in Julia buffers.")
+
+(cl-defmethod ess-beginning-of-function--override
+  (no-error &context ((string= ess-dialect "julia") (eql t)))
+  "Find the beginning of the function in Julia."
+  (let ((p (point))
+        (bound (or (julia-last-open-block (point-min))
+                   (point-min))))
+    (goto-char p)
+    ;; Calling `beginning-of-line' here excludes nested fu
+    (beginning-of-line)
+    (while (and (not (looking-at-p ess-function-pattern))
+                (< bound (point)))
+      (forward-line -1)
+      (beginning-of-line))
+    (if (looking-at-p ess-function-pattern)
+        (point)
+      (unless no-error (user-error "Point not inside function")))))
+
+(cl-defmethod ess-end-of-function--override
+  (&context ((string= ess-dialect "julia") (eql t)))
+  "Find the end of the function in Julia."
+  (if (looking-at-p "\\(function\\|macro\\)")
+      ;; TODO: Can probably do better than assuming whitespace is
+      ;; proper.
+      (let ((col (current-column))
+            stop)
+        (while (not stop)
+          (search-forward "end")
+          (goto-char (match-beginning 0))
+          (setq stop (eql col (current-column)))
+          (goto-char (match-end 0)))
+        (point))
+    ;; Else, we're in a function like f(x) = ...
+    (while (and (forward-line)
+                (not (eobp))
+                (julia-indent-hanging)))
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lisp/ess-julia.el [hunk 1]
```diff
@@ -364,6 +364,52 @@ It makes underscores and dots word constituent chars.")
 
 (defvar ess-julia-mode-syntax-table (copy-syntax-table julia-mode-syntax-table))
 
+(defvar ess-julia-function-pattern
+  ;; (rx (or "function"
+  ;;         "macro"
+  ;;         (seq (0+ nonl) "(" (1+ nonl) ")" (0+ whitespace) "=")))
+  "function\\|macro\\|.*(.+)[[:space:]]*="
```
nverno: The regex for parameters should allow for empty parens, eg. `(.*)` to support `f() = 1`. Also, something like
```
f(x
  ) =
      x +
      1
```
is legal syntax, so something like `([^)]*)` might be better.

lisp/ess-julia.el
```diff
@@ -389,7 +436,8 @@ It makes underscores and dots word constituent chars.")
   (add-hook 'completion-at-point-functions 'ess-julia-object-completion nil 'local)
   (add-hook 'completion-at-point-functions 'ess-filename-completion nil 'local)
   (if (fboundp 'ess-add-toolbar) (ess-add-toolbar))
-  (set (make-local-variable 'end-of-defun-function) 'ess-end-of-function)
+  (setq-local beginning-of-defun-function #'ess-beginning-of-function)
```
nverno: I though about adding this too, but `ess-beginning-of-function` doesn't have the same semantics as `beginning-of-defun`, although maybe that isn't an issue.

lisp/ess-julia.el
```diff
@@ -364,6 +364,52 @@ It makes underscores and dots word constituent chars.")
...
+    (goto-char p)
+    ;; Calling `beginning-of-line' here excludes nested fu
+    (beginning-of-line)
+    (while (and (not (looking-at-p ess-function-pattern))
+                (< bound (point)))
+      (forward-line -1)
+      (beginning-of-line))
+    (if (looking-at-p ess-function-pattern)
+        (point)
+      (unless no-error (user-error "Point not inside function")))))
```
nverno: if the point is outside a function, it will still jump back to the previous function here. Personally, I would throw a real error so it could be caught in a condition-case block

lisp/ess-julia.el
```diff
@@ -364,6 +364,52 @@ It makes underscores and dots word constituent chars.")
...
+          (goto-char (match-end 0)))
+        (point))
+    ;; Else, we're in a function like f(x) = ...
+    (while (and (forward-line)
+                (not (eobp))
+                (julia-indent-hanging)))
+    (forward-line -1)
+    (end-of-line)
+    (point)))
+
```
nverno: This function also moves the point when called outside of a function.

lionel-:
Looks good, but since we're making it a generic, I have some questions about the point of this API in ESS. As it stands, `ess-beginning-of-function` looks for the outermost function def in the current expression. This functionality is totally overlapping with the Emacs-wide `beginning-of-defun` customisation point.

The fact that we have this in ESS is a bit confusing. Why implement this in `ess-julia` rather in `julia-mode`? This is obviously a mode functionality. I think the implementation you provide here should be proposed for julia-mode, rather than merged in ESS.

I suggest we use the formal Emacs customisation point and deprecate `ess-beginning-of-function`. Or at least make it a simple wrapper around `beginning-of-defun` if we really need it. There are other types of syntax marking/motion we'd like to implement in ESS (there are open issues about it), and when we do implement them we should think about whether they'd make sense in upstream Emacs as well, at some point, and in that case we probably shouldn't use ESS specific mechanisms like cl-generics to implement them.

vspinu:
I agree with @lionel- on all points. I think we should keep ess-r-mode as clean and close to emacs standards as we can. We will likely be moving it out of ESS sooner or later.

jabranham:
This PR (and #785) don't change the current behavior in R buffers. This PR introduced some ess-specific generics, but #785 is better since it just relies on built-in variables that allow us to change how Emacs finds the beg/end of functions.

PR #783, opened 2018-12-05
Remove ess-step-line and fix evaluation of paragraphs on empty lines

vspinu:
Making a PR as per discussion in #721. I am removing `ess-step-line` because it does two things , has a non-suggestive name and, in fact, is not needed. Most of it was swallowed by ess-next-code-line and the places of its previous usage are now more explicit.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1546,6 +1546,10 @@ current function, otherwise (in case of an error) return nil."
 Prefix arg VIS toggles visibility of ess-code as for `ess-eval-region'."
   (interactive "P")
   (save-excursion
+    (ignore-errors
+      ;; Evaluation is forward oriented
+      (forward-line -1)
+      (ess-next-code-line 1))
     (let ((beg (progn (backward-paragraph) (point)))
           (end (progn (forward-paragraph) (point))))
       (ess-eval-region beg end vis))))
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1568,7 +1568,8 @@ Otherwise send the current paragraph to the inferior ESS process.
 Prefix arg VIS toggles visibility of ess-code as for
 `ess-eval-region'."
   (interactive "P")
-  (ess-step-line (ess-eval-function-or-paragraph vis)))
+  (ess-skip-thing (ess-eval-function-or-paragraph vis))
+  (ess-next-code-line))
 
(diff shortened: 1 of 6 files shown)


jabranham:
I'm still not sure why you're so opposed to this function. It's called by all the `ess-eval-*-and-step` functions, which ensures they act the same. It also explains the name. By removing it you're replacing it two two function calls for all those functions, which, while not the end of the world, is a violation of DRY.

vspinu:
One extra line is not a violation of DRYl. Sometimes it's useful to be more explicit about what the procedure is. Logic is much clearer this way and we don't have an extra function. We cannot add a function each time one wants to combine the logic of two existing functions. Plus most part of that function probably belongs to `ess-next-code-line` anyhow.

> I'm still not sure why you're so opposed to this function. It's called by all the ess-eval-*-and-step functions,

You cannot name a function by a name which derives its meaning from the names which call that function. You have to name a function based on what it does, in our case `skip-a-thing-and-step-to-next-code-line`.

lisp/ess-inf.el
```diff
@@ -1546,6 +1546,10 @@ current function, otherwise (in case of an error) return nil."
 Prefix arg VIS toggles visibility of ess-code as for `ess-eval-region'."
   (interactive "P")
   (save-excursion
+    (ignore-errors
```
jabranham: What's the purpose of this block? We're already calling backward-paragraph.

jabranham: Oh, nevermind. I see.

vspinu:
Sorry,  but the DRY argument does not outweighs the non-aptly named function and the unnecessary burden of an extra concept which I insisted on in #721 and did not repeat here.

Good names are hard and if we don't have one we should either make it internal or do what I did. Making it internal in this case is an inferior solution because we would still have to deal with awkward semantics of the name. Plus, an extra `ess-next-code-line` makes it crystal clear what that `-step` means in those eval functions. It's not a bad solution. 

If we have better ideas later we could always change this in the future.

jabranham:
This has broken `C-c C-c`:

```R
library(tidyverse)
library(rstanarm)
library(coda)
options(mc.cores = parallel::detectCores())

x <- 1
```

with point at the beginning of the bufffer C-c C-c should leave point on `x <- 1` but it now leaves point on `library(rstanarm)`

vspinu:
Indeed. No idea how I missed that. Fixed. Thanks!

PR #785, opened 2018-12-05
Obsolete ess-{beginning|end}-of-function

### LINKED ISSUE #777: Julia functions are not recognized by ess-function-pattern

jabranham:
Use beginning-of-defun and end-of-defun instead since they're more
generic. Major modes can set the appropriate
beginning/end-of-defun-function variables so that e.g.
ess-eval-function works out of the box.

Related to #779, #782, and closes #777


lisp/ess-inf.el
```diff
@@ -1507,39 +1507,17 @@ VIS has same meaning as for `ess-eval-region'."
...
-Prefix argument VIS toggles the meaning of `ess-eval-visibly'. If
-NO-ERROR is non-nil and the function was successfully evaluated,
-return '(beg end) representing the beginning and end of the
-current function, otherwise (in case of an error) return nil."
+Prefix arg VIS toggles visibility of ess-code as for
+`ess-eval-region'. Returns nil if not inside a function."
   (interactive "P")
-  (ess-force-buffer-current "Process to use: ")
   (save-excursion
-    (ignore-errors
```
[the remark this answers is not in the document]
jabranham: Should be fixed

lisp/ess-r-mode.el
```diff
@@ -604,17 +714,12 @@ Executed in process buffer."
...
     (imenu-add-to-menubar "Imenu-R"))
-
-  ;; useful for swankr/slime:
-  (setq-local beginning-of-defun-function
-              (lambda (&optional arg)
-                (skip-chars-backward " \t\n")
-                (ess-beginning-of-function 'no-error)))
-  (setq-local end-of-defun-function
-              'ess-end-of-function)
+  (setq-local beginning-of-defun-function #'ess-r-beginning-of-function)
```
vspinu: Let's call them `ess-r-beginning-of-defun` and `*-end-of-defun` as per emacs standards.

mmaechler:
For me and I think the other old ESS programmers, 
`defun` has been about emacs lisp functions, and `function` was about R/S/xlispstat/stata/julia/ ...... functions.    
But of course I do trust your judgement about emacs standards.

vspinu:
The `beginning-of-defun-function` has a generic purpose of identifying the start of all sorts of syntactic constructs (in elisp all top-level constructs and `defun`, `defvar`, `defconst` etc.). It's a misnomer but it's standard and would make it clear the purpose of our functions. Plus, our functions should probably be enhanced to work for other standard constructs (like `setClass`, `R6Class` etc) and maybe go to top-level form if everything else failed.

mmaechler:
> our functions should probably be enhanced to work for other standard constructs (like setClass, R6Class etc) and maybe go to top-level form if everything else failed.

They (i.e. the old `ess-beginning-of-function` with the very sophisticated beginning-of-function regular expression that I had written and improved upon in *many* man hours) *did* work almost always, notably including for `setGeneric(..)` and `setMethod(..)`, till some ESS core youngster decided they should be simplified...   I'm now fearing that your cleanup efforts will clean up even more functionality in favor of simplicity...
`setMethod()` etc has been standard R for more than ten years, and been part of S even before, and that's why it had nice support in older ESSen...

vspinu:
> with the very sophisticated beginning-of-function regular expression that I had written and improved upon in many man hours)

Your sophisticated regexp is still [there](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-utils.el#L897) and still works. I said `setClass`, not `setMethod`.

lisp/ess-r-mode.el
jabranham: If we do that, should we also remove the interactive spec and rely on them getting called directly or through beg/end-of-defun?

vspinu: I think so, though it might be better to keep them interactive as long as obsoleted defaliases are in place.

nverno:
the regexp may work, but it could use some significant cleanup IMO. when I was looking at it to construct one to work with julia it was basically impossible to make sense of. A simple regex instead of a bunch of let bound variables would be much easier to grok

lisp/ess-r-mode.el
jabranham: Alrighty, I renamed them and kept the interactive spec.

jabranham:
I've made the suggested changes and will merge this soon unless someone has more comments. 

@nverno Yes, the function could be better written and have better performance (see the other comments in this thread). Code contributions are welcome, assuming you can sign the FSF copyright assignment paperwork :-)

nverno:
Should the `ess` versions of beginning/end of defun functions follow the same semantics
as normal, eg. jump forward backward a given number of functions with a prefix argument?
`python` mode seems to handle this quite well

lionel-:
> till some ESS core youngster decided they should be simplified... I'm now fearing that your cleanup efforts will clean up even more functionality in favor of simplicity...

I don't think this is true, there was no simplification and the regexps are still there. On a side note, the ESS core "youngsters" spend a lot of time writing unit tests for existing code to prevent breakage of functionality before refactoring.

> Should the ess versions of beginning/end of defun functions follow the same semantics
as normal, eg. jump forward backward a given number of functions with a prefix argument?

They should.

> The beginning-of-defun-function has a generic purpose of identifying the start of all sorts of syntactic constructs (in elisp all top-level constructs and defun, defvar, defconst etc.). It's a misnomer but it's standard and would make it clear the purpose of our functions. Plus, our functions should probably be enhanced to work for other standard constructs (like setClass, R6Class etc) and maybe go to top-level form if everything else failed.

Maybe we should just evaluate the top-level R form. The current mechanisms to do this do not work for 100% of cases, but they should be reliable enough. If everyone agrees to this, I'll send a PR.

mmaechler:
> > till some ESS core youngster decided they should be simplified... I'm now fearing that your cleanup efforts will clean up even more functionality in favor of simplicity...

> I don't think this is true, there was no simplification and the regexps are still there. On a side note, the ESS core "youngsters" spend a lot of time writing unit tests for existing code to prevent breakage of functionality before refactoring.

I apologize for telling untrue things,  I've been pretty sure there was some partly harmful simplification at some point in time.   But I really wrongly concluded from your remarks the regexps were gone.
Sorry.

Lastly, I do very very much appreciate your introduction of unit tests;  in my view one of the really great improvements in the ESS code base.  Thank you very much!

Problem with the current setup is that it almost always fails inside long functions which have definitions of  local functions inside the long function.... and that is often a good style of R programming in quite a few places of R code I edit.   I had the impression that this had deteriorated quite a bit, but rather 1-3 years ago, than really recently.

jabranham:
>>  Should the ess versions of beginning/end of defun functions follow the same semantics
>>  as normal, eg. jump forward backward a given number of functions with a prefix argument?
> 
> They should.

[...]

> Maybe we should just evaluate the top-level R form. The current mechanisms to do this do not work for 100% of cases, but they should be reliable enough. If everyone agrees to this, I'll send a PR.

I agree with both of these things.

I'll go ahead and merge this so people can work on those parts.

vspinu:
> Maybe we should just evaluate the top-level R form. 

Ideally `C-u` would evaluate the inner form, but unfortunately we are stuck with a much less useful visibility toggling. Can put on `M-0` prefix maybe?

lionel-:
> we are stuck with a much less useful visibility toggling

I don't think it's that important, we can probably break this. I have never toggled off visibility, personally, and never experienced a slow down while evaluating a function.

My guess is that it was important pre-tracebug. When I disable tracebug and evaluate `[.data.table` (the longest function I know about), I get a substantial slowdown. With tracebug enabled, it's ok.

nverno:
There could be functions with normal semantics, eg `beginning-of-defun` would go to beginning of nested function, but default evaluation would be top-level. This is what `python-mode` does, for example, and seems the most useful since navigating between nested functions is often desired, but generally you would want to evaluate the entire block.

nverno:
I also was wondering if there should be a generic way to report function evaluation in `ess-eval-function`? I added some beginning/end functions to `julia-mode`, but the evaluation reporting is wrong due to differences in R vs. Julia function syntax.

jabranham:
@nverno what do you mean "report function evaluation"? Are you talking about the message that gets displayed? We're using `add-log-current-defun` right now. You likely need to modify either `add-log-current-defun-header-regexp` or `add-log-current-defun-function`

nverno:
I was referring to [this](https://github.com/emacs-ess/ESS/blob/efafb6060e7f058c8e0ead956d7aad3f9a50dad6/lisp/ess-inf.el#L1523) message. It's not a major issue, just a minor annoyance for something of the form `function f(x)` reporting "Eval function function" instead of "Eval function f" as it would for R. I wasn't aware of the `add-log-*` functions, I'll look at that.

jabranham:
Right, we get the name straight from `add-log-current-defun`. It should be easy enough to implement for julia-mode too.

Ideally we'd use `which-function`, which has a way more descriptive name and is a bit more flexible, but it is bad at the moment because it uses an imenu index that can easily get [outdated.](https://lists.gnu.org/archive/html/bug-gnu-emacs/2018-12/msg00364.html)

nverno:
Yea I saw you posted a bug on that. Ok, I was looking at ESS prior to the `add-log` function. That looks like it provides an easy solution.

mmaechler:
> > we are stuck with a much less useful visibility toggling
> 
> I don't think it's that important, we can probably break this. I have never toggled off visibility, personally, and never experienced a slow down while evaluating a function.

I've definitely needed the default of no-visibility... and at some point in time most experienced ESS users had done this.  So a way to temporarily *add* visibility  has been the prefix,  used importantly in teaching or showing someone when usually visibility is turned off.

vspinu:
Indeed, I use the visibility toggling for the same reason, mostly to posting input-output to forums. Till `ess-eval-visibly` exists we won't be able to change the prefix I am afraid.

PR #788, opened 2018-12-11
list julia packages

nverno:
just adds "using Pkg" to list installed Julia packages


jabranham:
I think we should use `import`, not `using`, right?

The reason I didn't do this in #775 is because I wasn't sure if we wanted to modify the environment without user's explicit consent. Perhaps we should first check if `Pkg` is available, query users if it's not, then only load a library if they want it?

PR #790, opened 2018-12-11
Fix finding Rterm on Windows

### LINKED ISSUE #787: M-x R in 18.10.2 does not work on Windows

jabranham:
'inferior-ess-r-program' might not point to an R executable,
especially on MS Windows, so we check it in 'run-ess-r'. If it's not
executable, set it to (ess-find-newest-R) which should be able to find
an R executable on any OS.

Closes #787

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -252,6 +252,11 @@ When t, loading a file into a namespaced will output information
 about which objects are exported and which stay hidden in the
 namespace.")
 
+(defvar ess-newest-R nil
+  "Stores the newest version of R that has been found.
+Used as a cache, within `ess-find-newest-R'. Do not use this value
+directly, but instead call the function \\[ess-find-newest-R].")
+
 (defun ess-r-font-lock-syntactic-face-function (state)
   (let ((string-end (save-excursion
                       (and (nth 3 state)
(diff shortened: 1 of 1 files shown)


lisp/ess-r-mode.el
```diff
@@ -917,6 +921,7 @@ Examples: (ess-current-R-at-least '2.7.0)
 (defun ess-newest-r (rvers)
```
vspinu: The  semantics of names `ess-find-newest-R` and `ess-newest-r` is unfortunate. I thought that the latter is actually running the newest R.  Shall we make the latter internal and name it more specifically, like `ess-r--newest-r-path`? Or just inline it whenever needed; it's a 2 liner. Capital R in one and not in the other is confusing.

lisp/ess-r-mode.el
```diff
@@ -476,6 +481,12 @@ will be prompted to enter arguments interactively."
    (format
     "\n(R): ess-dialect=%s, buf=%s, start-arg=%s\n current-prefix-arg=%s\n"
     ess-dialect (current-buffer) start-args current-prefix-arg))
+  (unless (executable-find inferior-ess-r-program)
```
vspinu: This setq side effect here is not good. Such an important variable should be set in one place and not interfered with.

lisp/ess-r-mode.el
```diff
@@ -252,6 +252,11 @@ When t, loading a file into a namespaced will output information
 about which objects are exported and which stay hidden in the
 namespace.")
 
+(defvar ess-newest-R nil
```
vspinu: Shall we make this one internal and make it more explicit like `ess-r--newest-executable` or something?

lisp/ess-r-mode.el
vspinu: I think this logic can be used whenever needed without setting `inefrior-ess-r-program`.

lisp/ess-r-mode.el
vspinu: Could be, but it then becomes semantically similar to `inferior-ess-r-program` which is not desirable.

Both `ess-newest-r` and the inner `ess-find-newest-date` are used once and have bad names. I would suggest we make a local `cl-labels` inside `ess-find-newest-R-version` and remove both of the above. Or at least inline `ess-find-newest-date` into `ess-newest-r` and rename the latter into an internal with a more suggestive name.

lisp/ess-r-mode.el
jabranham: Agreed in principal, but I'm not sure the best way to handle it. One option is to put something like this in the defcustom, which will increase the time it takes to load ESS.

> this logic can be used whenever needed without setting inefrior-ess-r-program

probably, but it sounds like more work than I'm willing to put in for an evil, proprietary operating system. :-)

PR #772, opened 2018-11-29
Only insert help links that link somewhere

jabranham:
Without this we buttonize too much, see discussion in #748

PR #798, opened 2018-12-17
R help links: Do not link to currently-displayed page

jabranham:
* lisp/ess-help.el (ess-help-object): Make permanent-local.

* lisp/ess-help.el (ess-display-help-on-object): Set ess-help-object
  before setting the major mode.

* lisp/ess-r-mode.el (ess-r-help-add-links): Do not link to current page

As discussed in #772. I'd go ahead and push this to master except for making ess-help-object permanent-local. I glanced through the code and didn't see anywhere we reuse existing help buffers. Is this safe as far as you know, @vspinu ?

cc @mmaechler


vspinu:
I think it's ok to make it peramnent local but I think this part should be redesigned into inherited help modes if possible.

PR #800, opened 2018-12-19
Tracebug cleanup

jabranham:
This PR does two things:

1. Enable lexical binding in two files (ess-trns and ess-tracebug), along with the usual cleanup to keep the byte compiler happy (usually removing unused lexically-bound variables)

2. Removes some 5+ year old obsolete stuff in ess-tracebug.el

The commits keep all that separate, if you'd rather view the changes one-at-a-time than all at once.


vspinu:
Looks good. Thanks!

PR #801, opened 2018-12-21
minimize ess-async-command-test

dickmao:
The vagaries of timing between R's process stdout and the idiosyncratic
`suppress-output` flag cause this test to occasionally fail, e.g.,
```
    (ert-test-failed
     ((should
       (string= "2 3 4 5 6> " output))
      :form
      (string= "2 3 4 5 6> " "6> ")
      :value nil))
```
Rework the test to merely ensure the callback eventually gets called, without
regard to what its output looks like.

### DIFF
--- a/test/ess-inf-tests.el
+++ b/test/ess-inf-tests.el
@@ -49,12 +49,14 @@
   ;; (ess-async-command "{cat(1:5);Sys.sleep(5);cat(2:6)}\n" nil (get-process "R")
   ;;                    (lambda (proc) (message "done"))
   (with-r-running nil
-    (let ((output-buffer (get-buffer-create " *ess-async-text-command-output*"))
-          output)
-      (ess-async-command "{cat(1:5);Sys.sleep(0.5);cat(2:6)}\n" output-buffer (get-process "R")
-                         (lambda (proc string) (setq output string)))
-      (sleep-for 0.6)
-      (should (string= "2 3 4 5 6> " output)))))
+    (ess-async-command "{cat(1:5);Sys.sleep(0.5);cat(2:6)}\n"
+                       (get-buffer-create " *ess-async-text-command-output*")
+                       (get-process "R")
+                       (lambda (proc string)))
+    (loop repeat 3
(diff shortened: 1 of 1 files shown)


test/ess-inf-tests.el
```diff
@@ -49,12 +49,14 @@
...
-      (sleep-for 0.6)
-      (should (string= "2 3 4 5 6> " output)))))
+    (ess-async-command "{cat(1:5);Sys.sleep(0.5);cat(2:6)}\n"
+                       (get-buffer-create " *ess-async-text-command-output*")
+                       (get-process "R")
+                       (lambda (proc string)))
+    (loop repeat 3
+          until (null (process-get (get-process "R") 'callbacks))
+          do (sleep-for 0 600)
+          finally (should-not (process-get (get-process "R") 'callbacks)))))
```
lionel-: Does this need a `(should (process-get (get-process "R") 'callbacks))` before the loop? This way we test that a callback has indeed be registered, since you no longer check for side effects in the inferior.

You could also assign a flag variable like `called_back <- TRUE` in the callback and check for it after the loop, if you think that's more robust than checking for output. Perhaps doing both of these is best.

dickmao: > Does this need a `(should (process-get (get-process "R") 'callbacks))` before the loop?

Yes, it does.

lionel-: Thanks, these new checks are useful. As mentioned above, I would also like to check for the side effect in the inferior, please. It could be with a regexp search to make it more robust.

dickmao: Done.

lionel-:
Thanks!

jabranham:
@dickmao Thanks for the contribution! Two minor points

1. Please avoid using the `cl` library in the future. It's deprecated in favor of `cl-lib`.

2. We're in the process of including ESS in Emacs, which involves assigning copyright to the FSF. If you want to contribute more to ESS, you'll need to fill out the paperwork. You can get the paperwork either by emailing me or assign at gnu dot org and requesting the paperwork to assign copyright for Emacs to the FSF.

PR #808, opened 2019-01-10
Remove lisp/ess-dde.el and lisp/ess-r-gui.el

### LINKED ISSUE #736: Remove ess-dde?
I'm not quite sure what ess-dde is. Some sort of windows REPL for R?  Judging by the git log, there hasn't really been any substantive changes to that file in ...8? years? Does anyone know if it still works? Can we remove it?

jabranham:
Closes #736

* lisp/ess-mouse.el:
* lisp/ess-sp6-d.el:
* lisp/ess-custom.el: Modify to account for dde removal


mmaechler:
Are you positive that these are not needed for Windows, the first one possibly for S-PLUS which we know has ESS users still ? @rmheiberger should now better.

We had always been able to keep this running for those users who had been very grateful for it.
Yes, there's big inertia in some companies, but the scientists working there have really been happy and grateful for ESS.

jabranham:
Over in #736 @vspinu said that Kim (the S-PLUS user) was able to get S-PLUS working with ess-remote. That sidesteps the need for ess-dde and ess-r-gui, which she couldn't get working anyway. At least, that's my understanding.

vspinu:
I don't think that we can make dde work any longer nor S+. It's a dead weight. Over the past 10 years I haven't seen a single developer using dde. 

Until an experienced elisp developer emerges which would be willing to maintain this functionality I think we should remove it from the code base.

rmheiberger:
ESS for SPlus GUI was broken when the relevant code was removed several
years ago.
ESS-5.14 is the last ESS that works with the standalone S-Plus GUI.
I don't remember what is in ess-dde.  The code I am talking about is an
if-then-else branch in one of the other files.

if ess-dde is not used except for that one situation which no longer works,
it can be archived.
I don't remember what ess-r-gui does, even though I probably participated
in writing it.

SPlus on Windows 10 works with the *S* buffer inside an emacs window using
ESS-5.14
 M-x Sqpe

I just tried on Windows 10 with ESS-16.10 and that freezes, requiring emacs
to be killed.
I don't know why.  I think it is unrelated to the dde issue.  I will look
at it later in January.


The major reason at one point for needing S-Plus GUI access from emacs was
interactive graphics.
The S-Plus interactive graphics window was accessible only from the GUI,
not from the *S* buffer.
At some point S-Plus
added a different interactive graphics device (java.graph) that works with
the command line S-Plus,
hence with the *S* buffer.
At that point the imperativeness of the S-Plus GUI access went away.  I
still think it was a mistake to remove
that functionality from ess, but I am also not motivated to restore it as
the restoration is not trivial.  Probably not
difficult, but would still require close attention.

mmaechler:
Yes, ok from my side.

PR #812, opened 2019-01-14
Refactor ess-rdired.el

jabranham:
This PR re-implements ess-rdired so that it derives from `tabulated-list-mode`. It's overall much cleaner (I think, anyway. It's at least shorter by about 140 lines!) and a little snappier.


jabranham:
I think the only difference is that instead of "marking" objects for deletion, then deleting them, it now operates on only the current object. Fairly minor, I think. I'll go ahead and merge then.

I'm planning on adding auto-update functionality soonish so you don't have to manually refresh the buffer with `g`.

PR #815, opened 2019-01-16
Effectful style

### LINKED ISSUE #661: Effectful style configuration
I have code like this:
```
tags$div(
      bar
      baz
    )

foo <- if(xxx) {
         xxx
       } else {
         yyy
       }
```

is it possible to make it indent like this?

```
tags$div(
  bar
  baz
)

foo <- if(xxx) {
  xxx
} else {
  yyy
}
```

the same happen to:
```
tibble::tibble(
          foo()
        )
```

jabranham:
This PR is three commits, see the commit messages for details:

1. Make it possible for file/dir-local variables to change the style setup
2. Obsolete ess-default-style since there is no longer a need for it
3. Use ess-style in ESS's directory local folder to set the style of R files to `RRR`. 

Closes #661.


lisp/ess-mode.el
```diff
@@ -317,9 +315,16 @@ indentation style. See `ess-style-alist' for predefined styles."
   (delq t comint-dynamic-complete-functions)
   (set (make-local-variable 'comint-completion-addsuffix)
        (cons "/" ""))
-
+  (add-hook 'hack-local-variables-hook #'ess--set-style-in-buffer nil t)
   (add-hook 'ess-idle-timer-functions 'ess-synchronize-dirs nil 'local))
 
+(defun ess--set-style-in-buffer ()
+  "Set `ess-style' taking into account file and directory local variables.
+This is added to `hack-local-variables-hook'."
```
vspinu: Cannot this one be done in ess-mode directly? Why to bother with `hack-local-variables-hook`?

lisp/ess-custom.el
```diff
@@ -1282,11 +1287,12 @@ global) values from ESS indentation variables."
...
                  (const RStudio-)
                  (const DEFAULT))
-  :group 'ess-edit)
+  :group 'ess-edit
+  :safe #'ess-style-safe)
 
-;; the real setting of this happens via <foo>-editing-alist:
-(defvar ess-style ess-default-style
-  "Current ESS indentation style, see `ess-style-alist' for more.")
+(defun ess-style-safe (v)
```
vspinu: Internal function?

lisp/ess-custom.el
```diff
@@ -1282,11 +1287,12 @@ global) values from ESS indentation variables."
...
+  "Return non-nil if V is a style in `ess-style-alist'."
+  (member v (mapcar 'car ess-style-alist)))
```
vspinu: Are you sure this one doesn't break when a user adds a personal style to ess-style-alist?

lisp/ess-mode.el
jabranham: Local variables are set after the major modes, so we'll miss them if it's inside `ess-mode`.

lisp/ess-custom.el
jabranham: I think it makes sense for this not to be internal. If people implement their own styles, they'll want to make sure they're treated as safe. WDYT?

lisp/ess-custom.el
jabranham: It appears to work fine. I'll add some ERT tests and push them in a minute.

jabranham: OK, done

lisp/ess-custom.el
vspinu: I think this function is useless to users. Safe is `ess-style` variable not individual styles. In order to be safe, aka assignable to `ess-style` a style need to be part of `ess-style-alist`.

lisp/ess-custom.el
jabranham: > How about just safing it to a symbolp?

That works, I guess. No need to be clever here. Done.

vspinu: I think we need to autoload it still, otherwise people with melpa might see the warning depending on how .dir-locals is configured.

jabranham: OK, done. I think I addressed all your feedback, so I'll merge soonish unless I hear otherwise.

PR #817, opened 2019-01-22
Query to save modified files in R packages during devtools commands

jabranham:
This PR saves, or offers to save (depending on `ess-silently-save`) modified buffers in the current R project before running devtools commands.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -674,10 +674,18 @@ blink the filling region."
   :group 'ess-edit
   :type 'boolean)
 
-(defcustom ess-mode-silently-save t
-  "Non-nil means automatically save ESS source buffers before loading."
+(define-obsolete-variable-alias 'ess-mode-silently-save 'ess-save-silently "ESS 19.04")
+(defcustom ess-save-silently 'auto
+  "If non-nil, possibly save buffers without asking.
+If t, save without asking. If 'auto, save without asking if
+either `compilation-ask-about-save' or `auto-save-visited-mode'
+is non-nil. Affects `ess-save-file'."
   :group 'ess-edit
-  :type 'boolean)
+  :type '(choice (const :tag "Do not save without asking." :value nil)
+                 (const :tag "Use compilation-ask-about-save and auto-save-visited-mode."
+                        :value 'auto)
+                 (const :tag "Save without asking." :value t))
+  :package-version '(ess . "19.04"))
 
 ;;*;; Variables controlling editing
 
--- a/lisp/ess-r-package.el
+++ b/lisp/ess-r-package.el
@@ -250,6 +250,7 @@ arguments, or expressions which return R arguments."
         (args (ess-r-command--build-args p actions)))
     (unless (car pkg-info)
       (user-error "Not in a package"))
+    (ess-project-save-buffers)
     (message msg (alist-get :name pkg-info))
     (display-buffer (ess-get-process-buffer))
     (let ((pkg-path (concat "'" (abbreviate-file-name (alist-get :root pkg-info)) "'")))
@@ -362,6 +363,16 @@ With prefix argument ARG, run tests on current file only."
   devtools::revdep_check_save_logs(res, logs_path)
 })
 ")
+
+(defun ess-project-save-buffers ()
+  "Offer to save modified files in the current project.
+Respects `ess-save-silently', which see."
+  (let ((cur-proj ess-r-package--info-cache))
+    (dolist (buf (buffer-list))
+      (when-let ((file (buffer-file-name buf))
+                 (buf-proj (buffer-local-value 'ess-r-package--info-cache buf)))
+        (when (equal cur-proj buf-proj)
+          (ess-save-file file))))))
 
 (defun ess-r-devtools-document-package (&optional arg)
   "Interface for `devtools::document()'.
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -51,6 +51,9 @@ ess-gen-proc-buffer-name:project-or-directory. As the name suggests,
 these now rely on project.el (included with Emacs) rather than
 projectile.el, which is a third-party package.
 
+@item ESS[R]: devtools commands ask about saving modified buffers before running.
+Users can disable the questioning with @code{ess-save-silently}.
+
 @item ESS[R] help pages now provide links to other help topics.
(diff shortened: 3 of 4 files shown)


jabranham:
This won't work as-is since `project-files` was only recently added to Emacs in git. Let me know if it's something yall would want though and I'll go about backporting it.

lisp/ess-r-package.el
```diff
@@ -363,6 +364,12 @@ With prefix argument ARG, run tests on current file only."
 })
 ")
 
+(defun ess-r-package-save-buffers ()
```
lionel-: Should this be `ess-project-save-buffers`? This seems generic enough. We should eventually have an ESS specific notion of projects, with `ess-project-current` picking up `.ess` files, RStudio project file, R packages, etc. This function would save buffers for projects thusly defined.

lisp/ess-custom.el
```diff
@@ -674,10 +674,12 @@ blink the filling region."
   :group 'ess-edit
   :type 'boolean)
 
-(defcustom ess-mode-silently-save t
-  "Non-nil means automatically save ESS source buffers before loading."
+(define-obsolete-variable-alias 'ess-mode-silently-save 'ess-silently-save "ESS 19.04")
+(defcustom ess-silently-save nil
```
lionel-: `ess-save-silently` would sound better and would be better namespaced.

lisp/ess-custom.el
```diff
@@ -674,10 +674,12 @@ blink the filling region."
...
+  "Non-nil means ask about saving modified files before loading."
```
lionel-: Is it the opposite?

lisp/ess-custom.el
vspinu: Why the default has been changed to nil? I would prefer it to be t by default. It's very unlikely that peopled wouldn't want it.

lisp/ess-r-package.el
```diff
@@ -363,6 +364,12 @@ With prefix argument ARG, run tests on current file only."
...
+  "Offer to save modified files in the current project.
+Respects `ess-silently-save', which see."
+  (dolist (file (project-files (project-current)))
```
vspinu: What does project-files do internally? If it finds all project files then this logic is sub-optimal. What we really need is to check only open files. Those are likely a handful and already have our R project root cached. So mapping through buffer list is likely a much, much faster alternative than this, also backward compatible from the start.

lisp/ess-r-package.el
jabranham: Sounds good to me. Renamed.

lisp/ess-custom.el
jabranham: @lionel- Yes, thanks, done.

@vspinu I changed it to nil since I don't think we should be saving files to disk without asking. It's also in line with what Emacs does (e.g. `compilation-ask-about-save`).

lisp/ess-custom.el
jabranham: It is indeed. Fixed, thanks

lisp/ess-r-package.el
jabranham: > then this logic is sub-optimal. 

Well, perhaps. If you've got a ton of buffers open but only a few files in your project, then this way is faster. But since `project-files` doesn't exist yet, I went with your idea anyway :-)

lisp/ess-r-package.el
```diff
@@ -363,6 +364,20 @@ With prefix argument ARG, run tests on current file only."
...
+Respects `ess-save-silently', which see."
+  (let ((cur-proj (project-roots (project-current))))
+    (dolist (buf (buffer-list))
+      (when-let ((file (buffer-file-name buf))
+                 (dir (buffer-local-value 'default-directory buf)))
+        (when (equal cur-proj (ignore-errors
+                                ;; If `project-current' returns nil,
+                                ;; `project-roots' errors.
+                                (project-roots
+                                 (project-current nil dir))))
```
vspinu: This is not quite what I meant. This logic is likely to be extremely slow as project.el doesn't cache anything currently. On remotes it will effectively stall. Worse it might blow on buffers with expired tramp sessions forcing tramp to reconnect.

 We already do cache R projects on file opening and I believe we should reuse it here. That part is not generic, but we could make it with the rename of the `ess-r-package-info-cache` I believe.

lisp/ess-custom.el
vspinu: I understand the logic and I am myself not entirely comfortable with it but, as always, it's a trade-off. If most or all our users will chose to auto-save then we shouldn't bother them with yet another awkward default. 

We can settle in the middle and check if auto-save mode is on and don't save if so.

lisp/ess-r-package.el
jabranham: OK, pushed a small commit that should do the trick I think.

lisp/ess-custom.el
jabranham: How about we just use `compilation-ask-about-save` instead of having our own? The purpose is so similar I can't imagine people wanting different behaviors.

lionel-: I like that idea.

vspinu: It's a non-generic setting and for a user wondering how to disable our "ask-for-save" there would be no way to figure out what to change except of going and reading the code. Our setting could default to compilation-ask-about-save in principle but I think auto-save-mode is more relevant still.  

We could have it set to 'auto and check both compilation-ask-about-save and auto-save-mode.

jabranham: Can you explain a bit why you think auto-save-mode is relevant here? If I understand correctly, auto-save-mode saves to separate files, it doesn't overwrite files like we're doing here. 

Are you thinking of `auto-save-visited-mode`?

vspinu: > Are you thinking of `auto-save-visited-mode`?

Yes, precisely.

jabranham: Gotcha. I didn't know about that newish mode until just now :-)

So now `ess-save-silently` defaults to `'auto`, which just means respect both `compilation-ask-about-save` and `auto-save-visited-mode`. I think this makes everyone happy, so I'll push to master later today or tomorrow unless I hear back.

mmaechler: Thank you both for the above discussion .. and a very good resolution (from my reading)!

PR #832, opened 2019-02-04
Move ess test files under pseudo-namespace ess-test

jabranham:
Make it easier to add the test directory to load-path, require the tests, and use ERT interactively.


test/ess-test-org.el
```diff
@@ -1,8 +1,26 @@
 ;; ess-org-test.el --- Test for org-babel integration -*- lexical-binding: t -*-
```
vspinu: Forgot about this one?

jabranham:
> do you also want to change the names of all tests according to this scheme?

I'm fine either way with this one. I thought about it but didn't want to put the time in to do so if we didn't want to merge this for some reason. If you want it though I can go through and do so.

test/ess-test-org.el
jabranham: Fixed, thanks

vspinu:
I think appending "-test" at the end is a more commonly used convention, but I personally don't care as long as there is a "test" in the name.  It's just that with this new file scheme feels at at odds with the current "-test". But yeh, not a big deal, either is fine.

test/ess-test-r.el
```diff
@@ -1,12 +1,31 @@
...
+;; A copy of the GNU General Public License is available at
+;; https://www.r-project.org/Licenses/
+;;
+;;; Commentary:
+;;
 
 (require 'ert)
 (require 'ess-r-mode)
-(require 'ess-r-tests-utils)
+(require 'ess-test-r-utils)
```
lionel-: General comment: I think this kind of stuff should be namespaced under `ess-r-test-`. I think it's a good principle to develop R stuff as if it were a separate package. Not a big deal though, we can rename all of this later again if/when we extract the R mode as its own package.

lionel-: Counter to what I said above, most of these tests are integration tests for ESS rather than unit tests for ESSR. I.e. we test the inferior code mostly through R.

jabranham: I imagine if/when we split out languages it'll involve a decent amount of renaming anyway.

PR #837, opened 2019-02-09
Add test for R help buffers

jabranham:
I'm trying to add this test but can't get it to succeed under emacs batch. It complains about the buffer getting killed. Any ideas? Right now I've just got it disabled with `(skip-unless (not noninteractive))` but I'd like it to run on CI too...


vspinu:
Can #839 be related? 

I am on a long trip to China and might not be able to look at this by the end March.

Issue #780, opened 2018-12-04
Create ESS+ package


jabranham:
As discussed in #732 many people want ESS plus some other things so we've agreed to provided ESS and ESS+. 

I think we should remove julia-mode.el from ESS and include in ESS+

- julia-mode
- poly-R

and perhaps others? What do others think?

jabranham:
Need to add documentation for this and #762 once it's done

jabranham:
Would people prefer ESS+ to live here or at a new repo under the emacs-ess github organization?

lionel-:
I think a new repo makes sense to clearly separate the different projects.

vspinu:
I actually think it might be better to have it here. People could download ESS and do `make ess+` which would download necessary packages. This is not enough for a separate package IMO.

PR #838, opened 2019-02-12
Move lisp files top-level

jabranham:
If we move all the lisp files top-level like package.el prefers, we can reference subdirectories with load-file-name. That'll let us add to `Info-directory-list` and make our lives easier in the long run, I think.

This has the side effect of removing julia-mode from our makefile steps. I think this a good step since we don't want it in there except when we provide ESS+ (#780)


lionel-:
I like this because it is consistent with the structure expected by package.el. I agree it will make our life easier in the long run.

Other things to consider:

* Expect all the auxiliary files to be installed within `load-file-name` (or subdirectories). This will make packaging of ESS by downstream distributions easier and more robust to internal changes.

* Make sure all files can be compiled out of order.

jabranham:
> Expect all the auxiliary files to be installed within `load-file-name` (or subdirectories). This will make packaging of ESS by downstream distributions easier and more robust to internal changes.

Yes, that will be nice once we incorporate this change.

> Make sure all files can be compiled out of order.

Isn't that already the case? The only reason we track dependencies in our makefile is for the case that when `foo.el` changes, `bar.el` which requires foo gets recompiled. Or am I misunderstanding something?

lionel-:
> Isn't that already the case?

IIUC it must be the case already because we're on MELPA which doesn't track makefile deps, right? I thought we may be missing compilation warnings because we always compile with the same dependency order while developing. Not a big deal I guess, as long as all macros are properly pulled in.

vspinu:
I agree that in principle having same repo structure as package.el is nice, but I think this change brings more troubles than it's worth.  First, all those users adding `lisp` folder to the load path will have to figure out what happened because their .emacs will no longer work. And that's a LOT of users. Second, we will have to adjust MELPA recipe and that takes time. We just did that last year. 

package.el is happy with everything being at a top level, and that's not necessary a good organization of the source code. I am personally pretty happy with lisp files living in lisp folder which is how code is normally organized in other languages.

I don't see the life of downstream packagers becomes any easier by this. We already have a makefile target that packages it into a package.el compatible tar.gz if they really need that.

Historically the installation varies. Mostly the content of the lisp will go somewhere and the content of doc and etc will go somewhere else. So this change will likely also bring havoc on all downstream distributors which will have now to manually isolate lisp files. So I would say we don't make this change unless we have good reasons.

> This has the side effect of removing julia-mode from our makefile steps.

But that means not compiling `ess-julia.el` ?

> If we move all the lisp files top-level like package.el prefers, we can reference subdirectories with load-file-name. That'll let us add to `Info-directory-list` and make our lives easier in the long run, I think.

 Could you please elaborate? Why cannot we just do it right now?

lionel-:
>  First, all those users adding `lisp` folder to the load path will have to figure out what happened because their .emacs will no longer work. And that's a LOT of users.

We could symlink it as a transitory measure. IIUC you're thinking about tarball users? This set is probably shrinking by the year, and will keep shrinking at a faster rate once we're in ELPA.

> Historically the installation varies. Mostly the content of the lisp will go somewhere and the content of doc and etc will go somewhere else. So this change will likely also bring havoc on all downstream distributors which will have now to manually isolate lisp files.

I don't understand. Why would they need to isolate files? As for the doc, if we use the Info structure of Emacs packages, they will be installed within `load-file-name` which Emacs will automatically add to `Info-directory-list`. No action needed.

For me the biggest advantage of following the package.el structure is that releasing a new version of ESS becomes as simple as pushing a tag on Github. The generated tarball for that tag is then a full blown release. Anyone in the team will be able to push a quick bugfix release.

lionel-:
> package.el is happy with everything being at a top level, and that's not necessary a good organization of the source code. I am personally pretty happy with lisp files living in lisp folder which is how code is normally organized in other languages.

I agree with this by the way. However I think the path of least resistance is to follow the organisation expected by the package manager. Also it doesn't seem totally awful to have lisp files at top-level, if everything else (except the makefile) is neatly organised in subfolders.

vspinu:
> We could symlink it as a transitory measure. IIUC you're thinking about tarball users?

They would still need to add the root dir to the path instead of the lisp dir. I am referring to all those who use ESS from source (tarball or git).

> I don't understand. Why would they need to isolate files?

We have a lot of stuff at the root which doesn't belong to the package build. So distributors will have to cherry pick the lisp files. When installed system wise info goes into a separate dir than the lisp files. 

> As for the doc, if we use the Info structure of Emacs packages, they will be installed within `load-file-name` which Emacs will automatically add to `Info-directory-list`. No action needed.

What install? What doc? What package? I am still confused what you guys are talking about. What installation procedure do you have in mind and how our `make package` cannot fix that?

> The generated tarball for that tag is then a full blown release. 

Ok. This is convenient indeed,  in case we really want the entire repo in the distribution, which wasn't the case thus far.  I still think this reason alone doesn't warrant the infliction of the above-mentioned pain on the users and distributors. 

> Anyone in the team will be able to push a quick bugfix release.

I think only the project lead should do that.

lionel-:
> They would still need to add the root dir to the path instead of the lisp dir

Even if the lisp dir was symlinked to the root dir? Or would the potential recursion cause problems? I agree this would cause some pain to the tarball users. The next release already switches to proper R modes which is bound to cause some incompatibility. I think we should take this opportunity to bundle all disturbing changes in the same release and document the transition process for users.

> We have a lot of stuff at the root which doesn't belong to the package build.

Ah got it. I guess this makes it more important to clean up the repo, but it's probably ok to have extra stuff in there as long as it doesn't take too much space. For instance the auctex tarballs seem to contain extra files that are not needed by the package manager nor by AucTeX itself http://git.savannah.gnu.org/cgit/emacs/elpa.git/tree/?h=externals/auctex.

> What installation procedure do you have in mind

ELPA or `package-install-file`. We already have a Makefile target for this, but using the package.el structure is about removing complexity from our build process.

> I think only the project lead should do that.

To be clear, I wasn't talking about uncoordinated releases. In any case, making the process simpler is a win especially when we have small fix releases to push as was the case last time.

jabranham:
Ideally I'd like to support distribution by distros and by package.el. I'd like to make the end-product as similar as possible and for this support to be as painless as possible for everyone involved (us, the distro packagers, users, etc). 

Supporting both seems to have competing goals:

distros want:
- to split info from lisp
- (potentially) to split etc from lisp (though I'm not sure this is really necessary)

package.el wants:
- lisp files to be top-level
- `dir` and `ess.info` to also be top-level (though I think we can work around this with a little autoload magic and referencing the `doc` dir relative to `load-file-name`. 

I think we should just disallow splitting etc, that way it's always a subdirectory of /path/to/ess. The systemwide `etc` dir is usually for configuration, and ours really isn't configuration files at all.

We can keep `dir` and `ess.info` in `doc/`, that way distros can still move them to where the other info files are. Then, in `ess.el` (or `ess-mode.el`, whatever) we can do something like:

```elisp
;;;###autoload
(add-to-list 'Info-directory-list (expand-file-name "doc" (or load-file-name buffer-file-name)))
```

but I haven't tested if that actually works yet.

lionel-:
Interesting idea. But I assumed distribution packagers would have means of splitting the info with the standard package.el structure?

vspinu:
> this support to be as painless as possible for everyone involved (us, the distro packagers, users, etc).

I am just afraid that this, not strictly necessary change, will trigger an avalanche of inconvenience and complains. But I agree with @lionel- that if we do this change, we better do it for next release, or never.  

> I think we should just disallow splitting etc, that way it's always a subdirectory of /path/to/ess.

This makes sense to me, but I am not a distro manager. I wouldn't be surprised if there will be people who don't like this. 

> We can keep `dir` and `ess.info` in `doc/`,

This would be nice indeed. 

BTW, do you know who else keeps lisp in a dedicated folder? Maybe we can get some ideas from them.

lionel-:
> do you know who else keeps lisp in a dedicated folder? Maybe we can get some ideas from them.

There is https://github.com/magit/magit/

lionel-:
> This makes sense to me, but I am not a distro manager. I wouldn't be surprised if there will be people who don't like this.

They also don't like when we change the organisation of the package and then break things on their side. We should insulate internal changes from downstream packaging.

jabranham:
Discussion here seems to have stalled. @vspinu @lionel- have you given this any more thought? WDYT?

vspinu:
I am undecided, but more inclined towards leaving it as it is. 

I think we should compile two clear lists of pros and cons and then weigh again.

lionel-:
I think with ESS+ coming up, we have an opportunity to split ESS into different more manageable packages. If we do this, maybe we should try the package.el organisation in spinoff packages before ESS itself. Then we can develop experience and a feel for this approach. I'll try to take some time to open an issue about the organisation of ESS the package so we can discuss it.

vspinu:
I like the proposal. Splitting is a fruitful direction indeed.

mmaechler:
I hope that  `polymode` could also become part of  `ESS+` ... so we'd finally get back to situation which is sane for the non-emacs-hacker R user trying to use ESS as an alternative ..

jabranham:
So we'd maintain (as separate packages): 

- ess
- ess-r-mode
- ess-julia-mode
- ess-sas-mode
- ess-bugs-mode
- ess-jags-mode

where the language packages depend on ESS, right?

I worry a little about splitting these files. It seems like it might make development more difficult. For example, if we right now want to make a defun in ess.el a cl-defgeneric so we can have different behavior by language, it's simple to do. If we've got different packages though then we have to worry about potentially conflicting versions of ess, ess-r-mode, ess-julia-mode, etc.

vspinu:
> I worry a little about splitting these files. It seems like it might make development more difficult. 

It surely will. This is the biggest drawback of the proposal. After splitting polymode the infrastructure management on my side became considerably more involved. Keeping everything under the same repo, or even under the same `/lisp` dir as we have right now would simplify the administration. But then we would need specialized make targets and MELPA filters to cherry pick the relevant parts for each sub-project.

lionel-:
>  If we've got different packages though then we have to worry about potentially conflicting versions of ess, ess-r-mode, ess-julia-mode, etc.

I think the split packages should be maintained and versioned together. Similarly to how base R packages are maintained and versioned with R itself.

jabranham:
@lionel- So when ESS version X is released, we simultaneously release ess-r version (and ess-julia, ess-stata, etc) Y that requires /exactly/ ESS version X? I'm not sure that's possible in the package.el framework. Don't package requirements just mean "at least" version X?

lionel-:
oh good point! People might update ESS but not ESSR. I've had a similar issue with rlang, I released a version that was compatible with the current CRAN dplyr but not earlier versions. People who updated rlang but not dplyr ended up in a bad state. There's no way to express minimal versions for downstream dependencies, so the only thing to do is wait a sufficient amount of time before breaking older versions.

Perhaps ESS could add version checks in the ESSR etc hooks, and prompt users to update their ESS packages?

But point taken, it might not be practical to split ESS this way, at least when things are still in flux...

jabranham:
I don't think we came to a consensus here one way or the other, so I'm closing the PR.

PR #839, opened 2019-02-14
Avoid killing help buffer invoked from package index

dickmao:
```
M-x R
C-c C-d i base
Click on object
```
will result in the warning `Selecting deleted buffer` and no help
buffer.  Such noninteractive calls to `ess-display-help-on-object` do
not first settle the help_type variable, resulting in the help logic
killing the buffer.

### DIFF
--- a/lisp/ess-help.el
+++ b/lisp/ess-help.el
@@ -46,7 +46,6 @@
 (require 'ess-utils)
 
 (declare-function ess-r-help-mode 'ess-r-mode)
-(declare-function ess-help-r--check-last-help-type 'ess-r-mode)
 (declare-function ess-stata-help-mode "ess-stata-lang")
 
 (defvar ess--help-frame nil
@@ -135,8 +134,7 @@ suplied, it is used instead of `inferior-ess-help-command'."
      (when current-prefix-arg ;update cache if prefix
        (ess-process-put 'sp-for-help-changed? t))
      (list (ess-find-help-file "Help on"))))
-  (ess-help-r--check-last-help-type) ; stabilize ess-help-r--last-help-type to avoid
-                                     ; killing tbuffer if R's help_type changed
+  (ess-help-get-topics ess-current-process-name) 
   (let* ((hb-name (concat "*help[" ess-current-process-name "]("
                           (replace-regexp-in-string "^\\?\\|`" "" object) ")*"))
          (old-hb-p (get-buffer hb-name))
--- a/lisp/ess-help.el
+++ b/lisp/ess-help.el
@@ -46,6 +46,7 @@
 (require 'ess-utils)
 
 (declare-function ess-r-help-mode 'ess-r-mode)
+(declare-function ess-help-r--check-last-help-type 'ess-r-mode)
 (declare-function ess-stata-help-mode "ess-stata-lang")
 
 (defvar ess--help-frame nil
@@ -141,6 +142,8 @@ suplied, it is used instead of `inferior-ess-help-command'."
     (when (or (not old-hb-p)
               current-prefix-arg
               (ess--help-get-bogus-buffer-substring old-hb-p))
+      (ess-help-r--check-last-help-type) ; stabilize ess-help-r--last-help-type to avoid
+                                         ; killing tbuffer if R's help_type changed
       (ess-with-current-buffer tbuffer
         (ess--flush-help-into-current-buffer object command)
         (setq ess-help-object object)
--- a/lisp/ess-help.el
+++ b/lisp/ess-help.el
@@ -135,6 +135,8 @@ suplied, it is used instead of `inferior-ess-help-command'."
      (when current-prefix-arg ;update cache if prefix
        (ess-process-put 'sp-for-help-changed? t))
      (list (ess-find-help-file "Help on"))))
+  (ess-help-r--check-last-help-type) ; stabilize ess-help-r--last-help-type to avoid
+                                     ; killing tbuffer if R's help_type changed
   (let* ((hb-name (concat "*help[" ess-current-process-name "]("
                           (replace-regexp-in-string "^\\?\\|`" "" object) ")*"))
          (old-hb-p (get-buffer hb-name))
@@ -142,8 +144,6 @@ suplied, it is used instead of `inferior-ess-help-command'."
     (when (or (not old-hb-p)
               current-prefix-arg
               (ess--help-get-bogus-buffer-substring old-hb-p))
-      (ess-help-r--check-last-help-type) ; stabilize ess-help-r--last-help-type to avoid
-                                         ; killing tbuffer if R's help_type changed
       (ess-with-current-buffer tbuffer
         (ess--flush-help-into-current-buffer object command)
         (setq ess-help-object object)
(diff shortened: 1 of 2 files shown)


jabranham:
Stuff in `ess-help` generally shouldn't be language specific so that it's easy to extend to other languages (e.g. julia). I think this patch does the trick:

```
diff --git a/lisp/ess-r-mode.el b/lisp/ess-r-mode.el
index 3d6cf2f8..70476736 100644
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -1241,7 +1241,8 @@ If it changes, we flush the cache.")
 
 (defun ess-help-r--check-last-help-type ()
   (let ((help-type (ess-string-command "getOption('help_type')\n")))
-    (when (not (string= help-type ess-help-r--last-help-type))
+    (when (and ess-help-r--last-help-type
+               (not (string= help-type ess-help-r--last-help-type)))
       (let ((help-buffers (ess-help-get-local-help-buffers)))
         (mapc #'kill-buffer help-buffers))
       (setq ess-help-r--last-help-type help-type))))

```

@vspinu If that looks good to you, I'll go ahead and commit. It's probably possible to add a test for this too.

@dickmao If you are interested in continuing to contribute to ESS you'll need to sign the FSF copyright transfer paperwork. We're working to get ESS in GNU Emacs and as a part of that the FSF wants contributor's copyright. You can find the instructions for how to do this here: https://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future

dickmao:
I prefer your patch.  I also prefer getting rid of all the buffer-killing help-type logic as your change makes the whole function a no-op (the `when` clause will always be false).

jabranham:
> as your change makes the whole function a no-op (the `when` clause will always be false).

Ah, right. We can't just git rid of this, otherwise we won't switch to html help when the user changes help_type. Will have another look later.

vspinu:
I don't like that kill-all help buffers logic on help type change.  We shouldn't do that. I would say we just remove `ess-help-r--check-last-help-type`. 

Just to make sure that we are on the same page, I see the following error only once per emacs session. All subsequent clicks work as expected. Is this the same that you see? 

```
Debugger entered--Lisp error: (error "Selecting deleted buffer")
  set-buffer(#<killed buffer>)
  (save-current-buffer (set-buffer buffer) (let ((PM (point-min)) (case-fold-search t) searching res) (setq res (or (< (- (point-max) PM) 80) (not (setq searching t)) (progn (goto-char PM) (re-search-forward "Error in help" nr-first t)) (progn (goto-char PM) (re-search-forward "^cat: .*--" nr-first t)) (progn (goto-char PM) (re-search-forward "no documentation .+" nr-first t)) (progn (goto-char PM) (re-search-forward "^help .*not found" nr-first t)))) (ess-write-to-dribble-buffer (format " |--> %s [searching %s]\n" res searching)) (if res (progn (if searching (buffer-substring (match-beginning 0) (match-end 0)) (buffer-string))))))
  ess--help-get-bogus-buffer-substring(#<killed buffer>)
  (let ((bog-mes (ess--help-get-bogus-buffer-substring buffer))) (if bog-mes (progn (ess-write-to-dribble-buffer (format "(ess-help: kill bogus buffer %s ..\n" (buffer-name buffer))) (message "%s" (replace-regexp-in-string "\n" "" bog-mes)) (kill-buffer buffer))))
  (progn (let ((bog-mes (ess--help-get-bogus-buffer-substring buffer))) (if bog-mes (progn (ess-write-to-dribble-buffer (format "(ess-help: kill bogus buffer %s ..\n" (buffer-name buffer))) (message "%s" (replace-regexp-in-string "\n" "" bog-mes)) (kill-buffer buffer)))))
  (if ess-help-kill-bogus-buffers (progn (let ((bog-mes (ess--help-get-bogus-buffer-substring buffer))) (if bog-mes (progn (ess-write-to-dribble-buffer (format "(ess-help: kill bogus buffer %s ..\n" (buffer-name buffer))) (message "%s" (replace-regexp-in-string "\n" "" bog-mes)) (kill-buffer buffer))))))
  ess--help-kill-bogus-buffer-maybe(#<killed buffer>)
  (if (ess--help-kill-bogus-buffer-maybe tbuffer) nil (ess-display-help tbuffer))
  (let* ((hb-name (concat "*help[" ess-current-process-name "](" (replace-regexp-in-string "^\\?\\|`" "" object) ")*")) (old-hb-p (get-buffer hb-name)) (tbuffer (get-buffer-create hb-name))) (if (or (not old-hb-p) current-prefix-arg (ess--help-get-bogus-buffer-substring old-hb-p)) (progn (let ((buf (current-buffer))) (save-current-buffer (set-buffer tbuffer) (ess-setq-vars-local (eval (buffer-local-value ... buf) t)) (setq ess-dialect (buffer-local-value 'ess-dialect buf)) (setq ess-local-process-name (buffer-local-value 'ess-local-process-name buf)) (ess--flush-help-into-current-buffer object command) (setq ess-help-object object) (ess--help-major-mode) (setq truncate-lines nil ess-help-type 'help))))) (if (ess--help-kill-bogus-buffer-maybe tbuffer) nil (ess-display-help tbuffer)))
  ess-display-help-on-object(".Platform" ".ess.help('.Platform', package = 'base')\n")
  (let* ((string (button-label button)) (command (if (fboundp ess-build-help-command-function) (progn (funcall ess-build-help-command-function string))))) (ess-display-help-on-object string command))
  ess--button-action(#<marker (moves after insertion) at 745 in *help[R](index:base)*>)
  button-activate(#<marker (moves after insertion) at 745 in *help[R](index:base)*> t)
  push-button(745 t)
  push-button((mouse-2 (#<window 10 on *help[R](index:base)*> 745 (25 . 420) 82810873 nil 745 (3 . 24) nil (1 . 12) (8 . 17))))
  funcall-interactively(push-button (mouse-2 (#<window 10 on *help[R](index:base)*> 745 (25 . 420) 82810873 nil 745 (3 . 24) nil (1 . 12) (8 . 17))))
  #<subr call-interactively>(push-button nil nil)
  apply(#<subr call-interactively> push-button (nil nil))
  call-interactively@ido-cr+-record-current-command(#<subr call-interactively> push-button nil nil)
  apply(call-interactively@ido-cr+-record-current-command #<subr call-interactively> (push-button nil nil))
  call-interactively(push-button nil nil)
  command-execute(push-button)
````

dickmao:
>  All subsequent clicks work as expected. Is this the same that you see?

Yes, because the second time through notes no change in R's help_type.

I committed a language agnostic fix although it does so in a very opaque way.

I am trying to imagine the usefulness of the killing.  So, I open some help buffers, then change R's help_type to html, and now I so loathe the existing help buffers I want to kill them?

jabranham:
> I don't like that kill-all help buffers logic on help type change. We shouldn't do that. I would say we just remove `ess-help-r--check-last-help-type`.

@vspinu if we remove it, users who first display help on an object (e.g. `?lm`), then switch `help_type` from text to html, then re-display help on `lm` will see the text, not the html.

That's a bug I'm OK living with, just want to make sure we're clear before removal. Tagging @lionel- since he originally wrote this feature.

vspinu:
That's because we re-use the already existing help buffer.  This should be fixed in some other way instead of killing all help buffers.

 I am also more than fine with this behavior as his option is somewhat expected to take full effect in new sessions only.

jabranham:
OK, I removed that function. @dickmao can you try it out and let us know if you still experience this bug? Thanks!

dickmao:
LGTM.  Thank you.

PR #848, opened 2019-02-23
Do not create help links for objects documented in current buffer

### LINKED ISSUE #847: link fontification in R-help buffers: can we use \alias{} or at least \usage{} to be smarter?
Consider `?dwilcox`.
In there, `s v` to jump to the  `Value:` section:  Of course all 4 functions [dpqr]wilcox()  are mentioned;
3 of them are rendered as links (wrongly), and `rwilcox` is not. ... and of course there are many other of such help pages where more than one function / object is documented, and in all those cases there's much too much fontified .. which is confusing, even sometimes for me, and even more if I open help pages for others (including teaching) in order to explain things based on the help page texts...

Now if the \alias{} info could be used, we'd know exactly what is on this page and we should hence not link-fontify...   A cheaper alternative would be to parse the "Usage:" section and typically guess correctly from there which objects are document on the given page....

jabranham:
Some R help pages (e.g. qt) document more than one object. Don't
create links that link to the current page. Do this by parsing the
"Usage:" section of the current help buffer.

Closes #847 

We still make links for some things that shouldn't be linked (e.g. `df` is linked in the `qt` help buffer because it's a function but the `df` there is referring to the `qt` argument, not the `df` function), but that should be tackled separately.

cc @mmaechler

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -2261,6 +2261,31 @@ state.")
         inferior-ess-help-command inferior-ess-r-help-command)
   (ess-r-help-add-links))
 
+(defun ess-r-help-usage-objects ()
+  "Return a list of objects in the usage section for the current help buffer.
+In other words, if in the help buffer for \"qt\", return '(\"dt\"
+\"pt\" \"qt\" \"rt\"). If the current buffer does not have a
+usage section, return nil."
+  (unless (derived-mode-p 'ess-r-help-mode)
+    (error "Not an R help buffer"))
+  (save-excursion
+    (save-restriction
+      (let (usage-objects)
+        (widen)
+        (goto-char (point-min))
+        ;; Narrow the buffer to just the "Usage" section
+        (when-let ((beg-of-usage (re-search-forward "^Usage:" nil t))
+                   (end-of-usage (re-search-forward "^[^[:space:]]")))
+          (forward-line -1)
+          (narrow-to-region beg-of-usage (point))
+          (goto-char (point-min))
+          ;; Match objects until a parens
+          (while (re-search-forward (rx bol (0+ whitespace) (group (1+ (not (any "("))))) end-of-usage t)
+            (add-to-list 'usage-objects (match-string-no-properties 1))
+            ;; Skip past function arguments
+            (forward-list)))
+        usage-objects))))
+
 (define-button-type 'ess-r-help-link
   'follow-link t
   'action (lambda (_) (ess-r-help-button-action)))
@@ -2274,7 +2299,8 @@ state.")
   "Add links to the help buffer."
   (let ((help-topics (when (ess-process-live-p)
                        (ess-help-get-topics ess-local-process-name)))
-        (inhibit-read-only t))
+        (inhibit-read-only t)
+        (usage-objects (ess-r-help-usage-objects)))
     (save-excursion
       ;; Search for fancy quotes only. If users have
       ;; options(useFancyQuotes) set to something other than TRUE this
@@ -2286,7 +2312,8 @@ state.")
                          (substring text nil (- (length text) 2))
                        text)))
           (when (and (member text help-topics)
-                     (not (string= text ess-help-object)))
+                     (not (string= text ess-help-object))
+                     (not (member text usage-objects)))
             (delete-region (match-beginning 0) (match-end 0))
             (insert-text-button text
                                 'ess-r-help-link-text text
--- a/test/ess-test-r.el
+++ b/test/ess-test-r.el
@@ -524,6 +524,39 @@ the_dat <- read.csv(\"foo.csv\")"
   (comint-next-prompt 1)
   (should (equal (line-number-at-pos) 3)))
 
+(ert-deftest ess-r-help-usage-objects-test ()
+  (with-temp-buffer
+    (let ((major-mode 'ess-r-help-mode))
+      (insert "The Student t Distribution
+
+Description:
+
+     Density, distribution function, quantile function and random
+     generation for the t distribution with df degrees of freedom
+     (and optional non-centrality parameter ‘ncp’).
+
+Usage:
+
+     dt(x, df, ncp, log = FALSE)
+     pt(q, df, ncp, lower.tail = TRUE, log.p = FALSE)
+     qt(p, df, ncp, lower.tail = TRUE, log.p = FALSE)
+     rt(n, df, ncp)
+
+Arguments:
+")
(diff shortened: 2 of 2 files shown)


mmaechler:
That was quick.. thank you!   The `df` you mentioned, could be tackled if the new `ess-r-help-usage-objects` function not only returned the name of the functions, but (as a separate list or list of lists) the function argument names... yes, even more work, but in some sense natural for a function that analyzes  / parse()s  the Usage section.    More generally it should probably return a list with one element for each usage entry and the element being something like an alist  with the name and the arguments..

Longer term it would of course be cleanest to work with the already parsed Rd objects available from R anyway  (at least for installed packages).    The advantages of these is that \alias{} are indeed available, also in cases they are not so easy to get from Usage such as  `[<-.data.frame` or  S3 (or S4!) methods or S4 classes, etc..

(and I am not sure if this should rather go into the #847 comments or here ..)

jabranham:
Thanks! I thought about that (returning the arguments as well as the function names), but there might be cases where a function has an argument `X` and also links to the function `X`. I'm not sure if it's better to add links in that case or not. WDYT?

mmaechler:
> Thanks! I thought about that (returning the arguments as well as the function names), but there might be cases where a function has an argument `X` and also links to the function `X`. I'm not sure if it's better to add links in that case or not. WDYT?

Non-pretending, cautious (i.e. scientifically decent behavior IMHO) would be to *not* add links.  
Again - we should keep in mind __ and comment so in the sources __ that this is hack anyway and the proper approach would use the `Rd` object (that *is* availble) and exactly follow its `\link{*}`s [and do so *also* in cases these links are *not* enclosed in `\code{}` sometimes for good and well thought through reasons [typically in the R sources themselves, where more thought has been spent on this than in current roxygenized pages...].

jabranham:
> Non-pretending, cautious (i.e. scientifically decent behavior IMHO) would be to _not_ add links.

I'm not sure what science has to do with anything here. We're not writing a research paper. The question is whether it's better to limit the amount of wrong links by not linking arguments to functions or if it's better to link as much as possible so that we don't miss offering potentially useful information.

> Again - we should keep in mind _and comment so in the sources_ that this is a hack anyway and the proper approach would use the `Rd` object (that _is_ available) and exactly follow its `\link{*}`s

How are Rd objects available from installed packages? Certainly not as a file. Is there some other way to get this information?

mmaechler:
> > Non-pretending, cautious (i.e. scientifically decent behavior IMHO) would be to _not_ add links.
> 
> I'm not sure what science has to do with anything here. ..

I may have to explain another time:  Science is good science if people do not claim things that are wrong.  Adding a wrong link falls into that category.  Rather erring on the cautious by *not* providing a dubious link is "scientifically decent" in that sense ..

> > Again - we should keep in mind _and comment so in the sources_ that this is a hack anyway and the proper approach would use the `Rd` object (that _is_ available) and exactly follow its `\link{*}`s
> 
> How are Rd objects available from installed packages? Certainly not as a file. Is there some other way to get this information?

How do you think
```r
help(df, help_type = "pdf")
```
could work if the `Rd` source was not available.  They are all saved in a "database" at package installation time.   But you are right that has not been part of the official R API yet (because people always thought we should be able to change structures .. but haven't done so for quite a few years!).
Following from above,
```r
utils:::print.help_files_with_topic
```
shows more of the nitty gritty details.

jabranham:
> I may have to explain another time: Science is good science if people do not claim things that are wrong. Adding a wrong link falls into that category. Rather erring on the cautious by not providing a dubious link is "scientifically decent" in that sense ..

Thanks, but you misunderstood what I was saying, I think. We aren't doing science here. There's no hypothesis test or modelling or whatnot. Therefore, I'm confused why you're talking about things being "scientifically decent." 

If you insist on using scientific jargon, though, here's my point put a different way: you're worried about Type I errors (putting a link where there shouldn't be one). I'm saying that may be preferable to committing a Type II error (failing to put a link where there should be one).

> How do you think
> help(df, help_type = "pdf")
> could work if the Rd source was not available

Exactly as it seems to: it accesses some compressed file to get the information it needs. Without R exposing that via an API or something, using it will be worse than what we're currently doing since it can change at any moment and break ESS.

jabranham:
@vspinu @lionel- WDYT ? Here's a TL;DR of the above:

In help buffers, we want to avoid putting links to objects documented in the current buffer (e.g. `?qt` shouldn't provide any links to `rt`. However, sometimes objects have argument names that are also functions. There's no way for us to distinguish whether the buffer is talking about the argument or the function (e.g. the instances of `data` in the `lm` help buffer ideally would not be links, but some help buffers may contain links to functions with the same name as one of their arguments). Without the Rd files (which don't seem to be available), there's no foolproof way to fix this. We can either slightly over-linkify by linking arguments to functions that share their name or slightly under-linkify by not linking to function arguments in the current buffer.

My (slight) preference is to err on the side of providing more information to users, even if it might be irrelevant. Martin would prefer not to link to argument names even though it means we might miss a link or two.

lionel-:
Maybe we could follow Martin's proposal but still link when the name is suffixed by `()`? In the tidyverse and r-lib docs we follow the convention of always referring to functions with the function call syntax.

mmaechler:
Hmm..  You are not providing more information to the users if you create links that go into a completely misleading direction.  And note that it's not just about the argument names.
There are many other misleading links produced by the current algorithm: It puts link whenever
1) something is marked (in help text mode) as "code-like".  
2) There's a help page "around" that is named the same as that "code-like" word.

And that's the basic flaw.. which I've started hating because I do open help quite often in ESS and have seen so many false positives: 

There quite a few other non-object  markups  in *.Rd files which also use a "code-like" font, e.g.,
\pkg{..}  
\file{..}  
\code{<non-R>}   
\command{..} 
\var{..}   and probably even more

Also - some good *.Rd pages have `\link{.}s`  which are *not* wrapped by `\code{.}`  (I think I had mentioned that before) where currently Type 1 errors are unavoidable.

Instead of using the internal *.Rd objects --- which I agree we cannot simply use currently, because they were not documented (I had forgotten about that - sorry), we may also use the html version of the help pages, because these do only translate `\link{.}`s  to hyperlinks and so should very very rarely contain errors of both kinds / types.

lionel-:
I think it's better not to linkify arguments because most of the time a documentation page refers to its own arguments rather than external functions of the same name. It's probably not worth worrying about the rare cases where that is not the case?

jabranham:
OK, I'll go ahead and merge this then. I'll open a new issue to track not linking arguments from functions in the current buffer.

Issue #857, opened 2019-03-08
ess-r-versions (extended version set by user) fails if ESS is started by system `site-start`


mmaechler:
This must be a recent change (well, almost surely):  I have for ages had an extended version of `ess-r-versions` e.g., one  "R-shared"  (as indeed my default versions of self-compiled R are *not* shared), and e.g. have  R-shared-usr  in my PATH , a sym.link going from ~/bin/R-shared-usr  ->   /usr/bin/R   the latter of course *not* a self-compiled version of R.   This has worked till a few days ago .. albeit possibly in an emacs version that ran for a week or so (and hence had used what was "current ESS" when that emacs started).

So  in short:  `R-shared-usr`  works fine in an emacs `*shell*` because  ~/bin of course is in $PATH,
but it is no longer found via   `M-x R-shared-usr`  
.. and yes, I do set `ess-r-versions` before loading ESS...    any ideas ?

It's pretty crippling for me being used to call my many different R-* versions from ESS.

jabranham:
Is `~/bin` in `exec-path`? We also don't make runners for files that return nil for any of the following, so perhaps that's what's wrong?

```elisp
(file-executable-p (expand-file-name "R-shared-usr" "~"))
(not (backup-file-name-p (expand-file-name "R-shared-usr" "~")))
(not (file-directory-p (expand-file-name "R-shared-usr" "~")))
```

mmaechler:
Thank you @jabranham ... all that would have been a problem if it had been one  long ago ... 

I found the solution our IT manages Fedora packages not just at work but also for my laptop and "they" decided unfortunately to install the 'ess'  Fedora package ... even though we had agreed they would not do that...  This is a big PITA: The workaround is to start emacs from a shell explicitly with `--no-site-file`  so my version of ESS is loaded *after* my `~/.emacs` (which sets `ess-r-versions`).     But this really shows how unfortunate the current setup is:  The user must set `ess-r-versions` before ESS is loaded but for > 90% of today's Linux users, they'd install ESS as a package of their distribution (ubuntu, debian, fedora, openSuSe,...) and so that is installed in the system `site-start` directory and startup file and all that loads *before*  `~/.emacs`  and so for all those users,  changing `ess-r-versions` has no effect. 

[...]

PR #858, opened 2019-03-08
Make ess-r-versions a defcustom

jabranham:
This lets us use the :set keyword to monitor when it changes so that we can update the runners. See #857.

### DIFF
--- a/doc/help-s.texi
+++ b/doc/help-s.texi
@@ -48,7 +48,7 @@ R on GNU/Linux systems and other Unix-like systems (macOS): If you have
 R-1.8.1}.  By default, ESS will find versions of R beginning ``R-1'',
 ``R-2'', ``R-3'', ``R-devel'', or ``R-patched''.  If your versions of R
 are called other names, consider renaming them with a symbolic link or
-change the variable @code{ess-r-versions}.  To see which functions have
+change the variable @code{ess-r-runners-prefixes}.  To see which functions have
 been created for starting different versions of R, type @kbd{M-x R-} and
 then hit [Tab].  These other versions of R can also be started from the
 "ESS->Start Process->Other" menu.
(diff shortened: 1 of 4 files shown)


jabranham:
Actually I don't think this will work as is. Will have to take a closer look later

jabranham:
Alright, should be ready now I think.

vspinu:
Shall we change the name on this occasion? I have no good ideas, but something more intuitive, like `executable-names` or `executable-prefixes` could be more appropriate.

jabranham:
I don't have a great idea, but how about any of these?

- ess-r-executable-versions
- ess-r-program-versions
- ess-r-versions-regexp
- ess-r-command-regexps

Though TBH I'm not sure if any of those are better than ess-r-versions...

lionel-:
I like `ess-r-program-versions`. Or perhaps `ess-r-runner-versions` to connect with `ess-r-runners`?

mmaechler:
> I like `ess-r-program-versions`. Or perhaps `ess-r-runner-versions` to connect with `ess-r-runners`?

Good point to bring "runners" into play.  Note that we have `(defun ess-r-define-runners ....)` [which uses `(defun ess-define-runner ...)`  and its result stored/cached in variable  `ess-r-created-runners' .. so I'd came to
`ess-r-runner-prefixes` or  `ess-r-prefix-runners`  [the latter because all current `*runner*` names end in `runner` or `runners`.

Note that "program-versions" is misleading too. This is really the variable with the **prefixes** and defaults such as "R-2" "R-3" ..

mmaechler:
what about `ess-r-runner-prefixes`  ?   A day later it seems the most  consistent and nice..

jabranham:
Sounds good to me. I'll make that change and merge this later today unless
I hear otherwise from someone.

vspinu:
Good idea. `ess-r-runners-prefixes` seems to nail it.

PR #860, opened 2019-03-12
Simplify indentation functions

jabranham:
Two psudo-related commits. The first is an under-the-hood change (i.e. won't affect users) where we use `indent-tab-function` rather than ess-specific things. The second obsoletes a few ESS-specific tab completion related variables in favor of the Emacs-wide tab-always-indent. I think it (for now) leaves behavior unchanged (unless they've changed tab-always-indent to 'complete without also updating ess-tab-complete-in-script) but users should migrate their settings.


lionel-:
👍

PR #870, opened 2019-03-20
Add Quit and Reload process to menu

lionel-:


lionel-:
@mmaechler I think PRs are only needed in these cases:

* To get a full CI check with Travis (when making non trivial changes).
* When changing the interface or internals in significant ways so others can weigh in.

No worries for pushing small stuff like this directly to master! Just wanted a review from you here because I don't really use the menu so I'm not sure how it should be organised.

PR #873, opened 2019-03-21
Update julia manual retrieval

nverno:
`ess-julia-manual-lookup-function` stopped working, it looks like julia's manual page was restructured. 
This adds simple modifications to `ess-julia--retrieve-topics` and `ess-julia-manual-lookup-function`
regexps to produce the possible topics.


vspinu:
Thanks!

PR #874, opened 2019-03-23
Extract `inferior-ess--get-proc-buffer-create` from `inferior-ess`

lionel-:
Simplifies control flow.

In a next PR I'll probably change `set-buffer` by `with-buffer-current` to avoid a side effect that makes it hard to reason about which buffer is current from the runners.


jabranham:
Thanks! I like the sound of it, should have time to look at the code
tomorrow.

vspinu:
I am a bit confused by your commits. Is this still WIP?

lionel-:
What is confusing about the commits? I plan to do other changes but this PR is finished.

lisp/ess-inf.el
```diff
@@ -172,97 +172,56 @@ This may be useful for debugging."
...
-                    (get-buffer-create buf-name-str)))
-        (setq method 3)))
-
-      (ess-write-to-dribble-buffer
-       (format "(inf-ess 2.0) Method #%d start=%s buf=%s\n" method cur-dir buf))
-
-      (set-buffer buf)
+           (proc-name (let ((ntry 1))
+                        (while (get-process (ess-proc-name ntry temp-dialect))
+                          (setq ntry (1+ ntry)))
```
jabranham: Maybe `(cl-incf ntry)` is shorter, but that's a really minor point.

lisp/ess-inf.el
```diff
@@ -172,97 +172,56 @@ This may be useful for debugging."
...
 
-      (ess-write-to-dribble-buffer
-       (format "(inf-ess 2.1): ess-language=%s, ess-dialect=%s buf=%s \n"
-               ess-language  ess-dialect (current-buffer)))
-
-      (let* ((infargs (or ess-start-args
-                          inferior-ess-start-args))
-             (proc (get-process procname)))
+      (let ((inf-args (or ess-start-args
+                          inferior-ess-start-args)))
```
jabranham: This variable (inferior-ess-start-args) should probably just be removed. It's a little awkward, and disconnected from e.g. inferior-R-args, which is awkward.

vspinu:
> this PR is finished.

Ok. Then it better be squashed. I see one change here not 13.

lisp/ess-inf.el
vspinu: I think it's never reached, at least for R as ess-start-args is always nil. I agree, it's bad design. Let's remove.

lisp/ess-inf.el
```diff
@@ -172,97 +172,56 @@ This may be useful for debugging."
...
-          (setq buf
-                (inferior-ess-make-comint buf-name-str
-                                          procname
-                                          infargs))
+          (setq inf-buf
+                (inferior-ess--make-comint inf-buf proc-name inf-args))
 
-          (set-buffer buf)
+          (set-buffer inf-buf)
           (set 'default-directory cur-dir)
```
vspinu: `setq` maybe, for consistency

lionel-:
> Ok. Then it better be squashed. I see one change here not 13.

I don't mind squashing this one as the overall diff is not that large and I don't want to have a discussion now, but I disagree squashing is better. There's nothing more annoying when navigating diff history to always hit the same large uninformative diffs. If I see a "Rename x to y" commit, I immediately now that's a trivial change that I can ignore and can navigate deeper without a second thought. If I see "Let-bind x earlier so it doesn't look like it depends on y", it instantly gives me better understanding of the source code. It seems to me that squashing only helps make the commit log shorter, which doesn't have much point.

lisp/ess-inf.el
lionel-: Best done in another PR as this seems to affect all languages.

vspinu:
>  same large uninformative diffs

There is a balance of course. What you do is the opposite extreme. 

> It seems to me that squashing only helps make the commit log shorter, which doesn't have much point.

Well, it's a big point for someone like myself who reads a commit history to get a feel of what has been done by contributors.  It's equally annoying to see N trivial commits instead doing one trivial X instead of one commit doing X. It's easy to miss the important stuff this way.

On the coder side, such micro-management is inefficient, mostly because precise and informative commit messages are hard. I have never seen anyone micro-managing their commits this way, I don't plan to do it myself, and I don't want to impose this on contributors. So please, if you can, refrain from exploding things like this. 

> now that's a trivial change 

Trivial changes could be grouped into one bigger trivial change. All your "rename of local-var", "indent", "use let*" are stylistic changes which can be grouped into one "Cleanup xyz" commit. Arguably, this entire PR is an aesthetic change, thus shattering it into so many commits is a bit crazy.

lionel-:
> It's easy to miss the important stuff this way.

Not with linear merge commits.

> mostly because precise and informative commit messages are hard

Yes but I still put in the effort to make them informative.

> stylistic changes which can be grouped into one "Cleanup xyz" commit

ok

>  thus shattering it into so many commits is a bit crazy.

I disagree.

jabranham:
There are ways of marking commits as not important, which makes it easier to pick out what's important in the log. We could adopt one of those conventions. Emacs upstream uses the convention that commits that begin with `;` are unimportant, which seems reasonable to me.

lionel-:
Interesting. This seems useful for both granular commits and cleanup commits.

lionel-:
One downside to the cleanup commits is that it's hard to clean up all at once and we often end up with

* Clean up xxx
* More clean up xxx
* Final clean up xxx

In any case cleanup commits should have function scope and not file scope so the overall diff is not too large.

vspinu:
> is that it's hard to clean up all at once

If one does a PR it's easy to group cleanup commit in one or two commits. If one commits upstream and then fixes the fix, that's unavoidable, so ok, be it. But intentionally splitting a cleanup commit in N cleanups for a PR is not good, sorry. 

>> mostly because precise and informative commit messages are hard
> Yes but I still put in the effort to make them informative.

That's precisely my point. You spend time and energy for nothing. Isolating code and composing precise messages is hard. For instance "rename infarg to inf-arg" still doesn't tell me where the change was made and why the change was made. So all that becomes non-informative noise in the history. 

> begin with ; are unimportant,

I always felt that emacs log is ugly because of that, but ok. If we start the commit message with informative verbs (cleanup, rewrite, rename etc) it goes a long way already.

lionel-:
> You spend time and energy for nothing.

These granular commits reflect my process for refactoring and I'm not going to change that. I have already conceded to merging clean up commits because I don't want to discuss this and I don't mind them as long as they're small enough.

Issue #871, opened 2019-03-20
Inferior reloading always uses default runner


lionel-:
From @mmaechler https://github.com/emacs-ess/ESS/issues/869#issuecomment-474791774

To fix this we'd need to have the `inferior-ess-r-program` variable used to start the inferior stored in the process buffer.

vspinu:
Not just that. Also the entire parameter list as it could be started with C-u.

lionel-:
Good point.

mmaechler:
Good points indeed.  Still should be fairly easily feasible, wouldn't it?

lionel-:
Should also reuse inferior name.

jabranham:
Can we just do something like this for the runner name? 

```diff
diff --git a/lisp/ess-mode.el b/lisp/ess-mode.el
index eb1fcf3c..b1e54d45 100644
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -668,7 +668,8 @@ Function defined using `ess-define-runner'."
             (cond ((string= dialect "R")
                    (let ((inferior-ess-r-program (or path name)))
                      (require 'ess-r-mode)
-                     (run-ess-r start-args)))
+                     (run-ess-r start-args)
+                     (setq-local inferior-ess-r-program inferior-ess-r-program)))
                   ((string= dialect "S")
                    (let ((inferior-S+-program (or path name)))
                      (require 'ess-sp6-d)
```

I'm not sure we need to save the parameter list; we don't do this for M-x R do we?

lionel-:
yup I was planning to do something like this.

I think it makes sense to save the parameter list though. I'm not sure what you mean about `M-x R`.

mmaechler:
If I use  `C-u M-x R-3.5.0`   and then specify extra command line arguments to R, say "--no-save",  we'd want  the "restart R" to use those (and not ask the user again about them).
... or did I misunderstand what you were asking?

jabranham:
Hmm... apparently I wasn't clear. I meant if you do `C-u M-x run-ess-r` and add some command line argument, then later restart that R, we don't re-apply that argument.

In other words, R's started via the runners and R started with run-ess-r should behave the same from the user's point of view. 

Is that clearer?

lionel-:
hmm, maybe I'm still missing something, but I would expect we reapply the arguments specified by the user on startup.

jabranham:
[...]

PR #875, opened 2019-03-25
Always start a new process when `inferior-ess` is called

lionel-:
`inferior-ess--get-proc-buffer-create` now always returns a process-less buffer. Then we can assume we're starting a process in `inferior-ess`, making things simpler.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -172,7 +172,6 @@ This may be useful for debugging."
                                temp-ess-dialect)
                            temp-ess-dialect))
            (inf-buf (inferior-ess--get-proc-buffer-create temp-dialect))
-           (inf-name (buffer-name inf-buf))
            (proc-name (with-current-buffer inf-buf ess-local-process-name))
            (cur-dir (inferior-ess--maybe-prompt-startup-directory proc-name temp-dialect))
            (default-directory cur-dir))
@@ -186,8 +185,6 @@ This may be useful for debugging."
       (let ((inf-args (or ess-start-args
                           inferior-ess-start-args)))
         (ess--inferior-major-mode ess-dialect)
-        (with-current-buffer inf-buf
-          (rename-buffer inf-name t))
         ;; Show the buffer
         ;; TODO: Remove inferior-ess-own-frame after ESS 19.04, then just have:
         ;; (pop-to-buffer inf-buf)
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -223,18 +223,18 @@ This may be useful for debugging."
             (ess-wait-for-process proc)))
         inf-buf))))
 
-(defun inferior-ess--get-proc-buffer-create (temp-dialect)
+(defun inferior-ess--get-proc-buffer-create (name)
   "Get a process buffer, creating a new one if needed.
 This always returns a process-less buffer. The variable
 `ess-local-process-name' is set in the buffer with the name of
 the next process to spawn. This name may be different from the
 buffer name, depending on how `ess-gen-proc-buffer-name-function'
-generated the latter."
+generated the latter from NAME."
   (let* ((proc-name (let ((ntry 1))
                       ;; Find the next non-existent process N (*R:N*)
-                      (while (get-process (ess-proc-name ntry temp-dialect))
+                      (while (get-process (ess-proc-name ntry name))
                         (setq ntry (1+ ntry)))
-                      (ess-proc-name ntry temp-dialect)))
+                      (ess-proc-name ntry name)))
          (inf-name (funcall ess-gen-proc-buffer-name-function proc-name)))
     (let ((buf (cond
                 ;; Try to use current buffer, if inferior-ess-mode but
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -176,7 +176,6 @@ This may be useful for debugging."
                         (while (get-process (ess-proc-name ntry temp-dialect))
                           (setq ntry (1+ ntry)))
                         (ess-proc-name ntry temp-dialect)))
-           (proc (get-process proc-name))
            (inf-buf (inferior-ess--get-proc-buffer-create proc-name))
            (inf-name (buffer-name inf-buf))
            (cur-dir (inferior-ess--maybe-prompt-startup-directory proc-name temp-dialect))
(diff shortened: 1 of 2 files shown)


lionel-:
Is the transcript stuff really useful? What is the advantage over saving the process buffer to a file?

mmaechler:
> Is the transcript stuff really useful? What is the advantage over saving the process buffer to a file?

Do you mean `transcript` in general, or just here?   I feel `transcript-mode` in general is really useful, and quite a bit under used.

lisp/ess-inf.el
```diff
@@ -250,18 +241,18 @@ This may be useful for debugging."
...
      (ess-ask-about-transfile
       (let ((transfilename (read-file-name
                             "Use transcript file (default none):" nil "")))
         (if (string= transfilename "")
             (get-buffer-create inf-name)
-          (find-file-noselect (expand-file-name transfilename)))))
-     ;; Create a new buffer
+          (let ((buf (find-file-noselect (expand-file-name transfilename))))
+            (if (get-buffer-process buf)
+                (error "Can't start a transcripted session if it already exists")
```
vspinu: I don't think we should blindly throw errors here. I think if the dead-proc buffer exists, even if ess-ask-about-transfile is t, we should not load new transfile, just like the old logic did.

lionel-:
@mmaechler I mean just `ess-ask-about-transfile`.

vspinu:
> Is the transcript stuff really useful?

I don't think so.  `ess-ask-about-transfile` is a useful UI for sure. I believe the transcript should be loaded on request only. That is, start R, call ess-load-transcript to insert the session into the current inferior if needed.

lisp/ess-inf.el
lionel-: This error is when there is a live proc, not a dead proc. What do you mean by "not load new transfile"? The old logic just exited from creating a new process. This behaviour is no different and makes control flow simpler.

vspinu: The old logic would just return a file with "inf-name" if it exists, live process or not. This is changed now in case `ess-ask-about-transfile` is non-nil. Throwing if `ess-ask-about-transfile` is non-nil doesn't seem right. At the very least one should throw regardless of that variable.

lisp/ess-inf.el
```diff
@@ -250,18 +241,18 @@ This may be useful for debugging."
      ((and (not (comint-check-proc (current-buffer)))
            (derived-mode-p 'inferior-ess-mode))
       (current-buffer))
-     ;; Take the *R:N* buffer if already exists (and contains dead proc!)
-     ;; fixme: buffer name might have been changed, iterate over all
-     ;; inferior-ess buffers.
```
vspinu: This `fixme:` said something useful. Was that issue addressed?

lisp/ess-inf.el
lionel-: If there were a live process, this if branch would exit: https://github.com/emacs-ess/ESS/blob/661daee75deace95ccecb37014bde6b475004e31/lisp/ess-inf.el#L194

Hmm I think I'm missing a `comint-check-proc` actually. Is this what you mean?

lisp/ess-inf.el
```diff
@@ -241,7 +231,8 @@ This may be useful for debugging."
             (ess-wait-for-process proc)))))))
 
 (defun inferior-ess--get-proc-buffer-create (proc-name)
-  "Get a process buffer, creating a new one if needed."
+  "Get a process buffer, creating a new one if needed.
+This always returns a process-less buffer."
```
vspinu: I don't think that this function does that. It just returns the buffer `inf-name`. If `inf-name` buffer has process, it doesn't check it.

lionel-: yes but `inf-name` will never have a live process in it. I wanted to pull the loop that checks for this into this function, but `proc-name` is needed in the caller. I can add a redundant check just for readability.

lisp/ess-inf.el
vspinu: >  if branch would exit:

It doesn't matter what the consumer of this function does. If a logic has been extracted into a function it should be self-contained and behave as documented. 

> comint-check-proc actually.

I don't have a good picture of what the new code does, but I think it's not only that. One needs to ensure that the name returned by that func-gen doesn't name an existing buffer, right?

lisp/ess-inf.el
lionel-: hmm or is it possible that `ess-gen-proc-buffer-name-function` may return a buffer name with a live process?

But then is there something wrong with that loop?

```lisp
           (proc-name (let ((ntry 1))
                        (while (get-process (ess-proc-name ntry temp-dialect))
                          (setq ntry (1+ ntry)))
                        (ess-proc-name ntry temp-dialect)))
```

Should it call the name generating function before checking for process?

lisp/ess-inf.el
lionel-: >  If a logic has been extracted into a function it should be self-contained and behave as documented.

Agreed, though that's an internal function so we're allowed to get there gradually. I'll try to find a solution.

> One needs to ensure that the name returned by that func-gen doesn't name an existing buffer, right?

Yes I think so, cf other thread. I think I know how to fix this.

lisp/ess-inf.el
lionel-: oops. I don't understand this fixme note so I don't know if it's still relevant, I'll add it back.

lisp/ess-inf.el
vspinu: eturn a buffer name with a live process 

Potentially yes. But, that loop is for process name, which always stays the same. The gen function is for buf names. As our default buf-generators include proc-name (right?) the buf names are probably unique unless someone changed them by hand.

I cannot help more as I write from memory of that code. Just do whatever feels right but please make that new function self-contained somehow, even if that requires throwing.

lisp/ess-inf.el
vspinu: It's terse and it's highly relevant to our other thread. The idea is that if there was a process "R:2" and it's buffer name was change from `*R:2*` to `*G:2*` manually (rename-buffer) and the process died, then this logic will not detect that buffer. Consequently the process restart breaks if buffer name changed (when called from the script buffer).

Would be great if it could be fixed now as you are on it. BTW, instead of iterating over all buffers I believe we can iterate over all our processes.

lionel-: This seems complicated, especially with the name generating function. If the user renamed `R:2` to `G:2`, it's not obvious to me that we should recycle that buffer. I would expect we don't actually.

lionel-:
`inferior-ess--get-proc-buffer-create` is now self-contained and always throws if a process already exists in the selected buffer. A unit test covers this by let-binding the name generating function to return a constant.

lisp/ess-inf.el
```diff
@@ -241,30 +223,53 @@ This may be useful for debugging."
...
-      (let ((transfilename (read-file-name
-                            "Use transcript file (default none):" nil "")))
-        (if (string= transfilename "")
-            (get-buffer-create inf-name)
-          (find-file-noselect (expand-file-name transfilename)))))
-     ;; Create a new buffer
-     (t
-      (get-buffer-create inf-name)))))
+(defun inferior-ess--get-proc-buffer-create (temp-dialect)
+  "Get a process buffer, creating a new one if needed.
```
jabranham: Checkdoc will complain about temp-dialect, can you mention it in the docstring please?

lisp/ess-inf.el
vspinu: I think we should. At the very least for the restarts. If I want to restart the process in G:2 buffer, it should stay there. 

Also for consistency. If you start a new R session from an non-proc G:2 buffer, it will reuse the G:2 buffer.

vspinu:
Looks good to me. Much nicer this way. Thanks!

lisp/ess-inf.el
lionel-: Agreed that reload should keep the same buffer. Let's keep the current logic and think about inferior restart as part of #871.

lisp/ess-inf.el
lionel-: oops, thank. I should get more familiar with checkdock.

lionel-:
Also allowed to simplify `inferior-ess--start-process` as the check for live process is no longer necessary: https://github.com/emacs-ess/ESS/commit/e0e632c384f9ae4214ed217ec4ed6c53cf0dd5c7

And I've moved the proc initialisation in that function: https://github.com/emacs-ess/ESS/commit/a11a34fa952b28a7b7d783d57af9c31213594ba8

lisp/ess-inf.el
jabranham: Flymake in Emacs 26+ uses it as one of its backends: `(add-hook 'emacs-lisp-mode-hook #'flymake-mode)`

PR #821, opened 2019-01-25
Remove many unused files

jabranham:
I don't see any of these used anywhere. Is there a reason to keep some/all/none of them?

```
* .gitignore: Remove *NEWS

* doc/archive/OONEWS:
* doc/archive/rsn.txt: Rename from top-level

* fontlock-test/baseball.sas:
* fontlock-test/elem.sas:
* fontlock-test/runif.Rcxx:
* fontlock-test/testing.sas:
* install-sh:
* lisp/old/ess-debug.el:
* lisp/old/ess-install.el:
* lisp/old/ess-old-contrib.el:
* lisp/old/ess-old-s.el:
* lisp/old/ess-old-site.el:
* lisp/old/ess-send.el:
* lisp/old/ess-send2.el:
* lisp/tmp/ess-dump.el:
* lisp/tmp/ess-rlib.R:
* lisp/tmp/ess-rlib.el:
* mkinstalldirs: Remove
```

PR #882, opened 2019-03-30
Archive historical versions of ESS in the repository

lionel-:
I would like to record the historical versions of ESS into the repository, for archiving purposes. The earliest revision of ESS is April 4th 1997, so we're missing at least 5 years of development. This PR creates a parallel git history for versions that were not revisioned. It includes the very first two versions of S-mode which are still available in the statlib archive at http://ftp.uni-bayreuth.de/math/statlib/S/. 

The first commit includes S-mode 2.1 by Doug Bates, Ed Kademan, and Frank Ritter. The second updates to 3.4 by David Smith. The commits were manufactured to reflect the original authors and release dates.

We're still missing 3.5, 4.0, 4.2, 4.3, 4.4, 4.5, 4.6, 4.7, 4.8, and 4.8.MM6.XE3. The version after that is 5.0 which can be found at https://github.com/emacs-ess/ESS/commit/8918012bc. We can add these versions on top of the parallel history if we ever come across them. @mmaechler perhaps you still have some old tarballs somewhere?

When merged, this will be saved in the `archive` branch. We should then push release tags to record them properly, and the branch can be deleted.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,0 +1,124 @@
+This file documents the changes from version 2.1 of S.el (also known
+as gnuemacs3 on Statlib) and the current version.
+
+For current users of S-mode, here are some of the incompatible
+changes and features new to version 3.3 of S-mode:
+
+   * Command-line completion of object names, and faster completion in
+     other situations.
+
+   * `Hot Keys' for the commonly-used functions `objects()',
+     `search()' and `attach()' and a facility to add your own hot keys
+     with keyboard macros.
+
+   * Simultaneous multiple function editing, with integrated
+     error-checking and parsing. Mnemonic names for edit buffers.
+
+   * Debugging features: facility for stepping through S code and
+     evaluating portions of code with the output appearing as if the
+     commands has been typed in manually.
+
+   * S can now be run from a different directory each session.
+
+   * A dedicated mode for viewing S help files. Individual help
+     buffers are maintained for quick repeated access. Completion for
+     help files without a corresponding object.
+
+   * Facility for maintaining organised backups of S source code.
+
+   * Indenting and formatting commands for editing S source code.
+
+   * Special handling of the S graphics facilities, including an
+     experimental Tek graphics mode.
+
+   * Better handling of temporary files and buffers.
+
+
+   * Some keybindings have changed to conform to GNU guidelines.
+
+Here are the changes to the keybindings from version 2.1:
+
+INFERIOR-S-MODE:
+
+C-c h    S-display-help-on-object    MOVED TO	   C-c C-h         [*]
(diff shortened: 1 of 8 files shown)


jabranham:
Thanks for looking into this! Out of curiosity, what prompted it? 

> When merged, this will be saved in the archive branch. We should then push release tags to record them properly, and the branch can be deleted.

So these commits will just not ever exist on the `master` branch?

lionel-:
> Thanks for looking into this! Out of curiosity, what prompted it?

I had the idea when we discussed about removing unused historical files: https://github.com/emacs-ess/ESS/pull/821#issuecomment-462021451

I've already collected commit hashes for all ESS releases. The next step is to push the missing tags on the repo. I think annotated tags are best, with the NEWS of the release documented in the tag. These annotated tags have an author and an authordate, which should be set to the person who did the release and the date they did it. They probably don't need to be signed.

> So these commits will just not ever exist on the master branch?

That's right. Unfortunately it's not possible to include them in the master history without rehashing the whole thing.

jabranham:
> I've already collected commit hashes for all ESS releases. The next step is to push the missing tags on the repo. I think annotated tags are best, with the NEWS of the release documented in the tag. These annotated tags have an author and an authordate, which should be set to the person who did the release and the date they did it. They probably don't need to be signed.

Sounds good to me. I wonder if we should start including the release news in our more contemporary tags...

> Unfortunately it's not possible to include them in the master history without rehashing the whole thing.

That's what I figured. Still will be nice, I suppose. Thanks for looking into this!

mmaechler:
Thank you @lionel- for doing this. I've now searched a bit, and did not find earlier ESS versions than 5.1, so you got even further back than I could have ...

vspinu:
How will this work for the future releases? You want to include the code, right? Does this mean swashing the whole release into the archive?

lionel-:
This history will be completely independent from new releases. It's only for releases up to 5.0.

lionel-:
These releases can now be visited at:

https://github.com/emacs-ess/ESS/tree/v2.1
https://github.com/emacs-ess/ESS/tree/v3.4

I have also added them to the release page for discoverability. They can be found in the very last page: https://github.com/emacs-ess/ESS/releases?after=v12.09-2

jabranham:
Thanks! Can we delete the `archive` branch?

lionel-:
We could, the tags themselves prevent garbage collection.

jabranham:
Sorry, I guess what I should've asked is if you're OK removing it. Is there any reason to keep it around (if, e.g. you were planning on adding more versions later on)?

lionel-:
no worries I just deleted it, I guess we can easily add it back if needed to add new versions.

PR #884, opened 2019-03-30
Return inferior buffer from runners

lionel-:
To avoid the side effect of `(set-buffer)` which is hard to reason about.

The R runner still sets the buffer because `ob-R` relies on this undocumented side effect: https://code.orgmode.org/bzg/org-mode/src/a9d3eaa9f341f57189c3e676b9bacff52f440fbf/lisp/ob-R.el#L283

I have prepared a patch for org-mode:

```patch
From 83481be6103bbaeadffc09f592e3bfee643c914a Mon Sep 17 00:00:00 2001
From: Lionel Henry <lionel.hry@gmail.com>
Date: Sat, 30 Mar 2019 14:54:44 +0100
Subject: [PATCH] Don't assume current buffer is process buffer

---
 lisp/ob-R.el | 22 ++++++++++++----------
 1 file changed, 12 insertions(+), 10 deletions(-)

diff --git a/lisp/ob-R.el b/lisp/ob-R.el
index 9e738a8a5..bb0419b9c 100644
--- a/lisp/ob-R.el
+++ b/lisp/ob-R.el
@@ -270,17 +270,19 @@ This function is called by `org-babel-execute-src-block'."
 	    ;; Session buffer exists, but with dead process
 	    (set-buffer session))
 	  (require 'ess) (R)
-	  (let ((R-proc (get-process (or ess-local-process-name
-					 ess-current-process-name))))
+	  (let* ((R-proc (get-process (or ess-local-process-name
+					  ess-current-process-name)))
+		 (R-buf (process-buffer R-proc)))
 	    (while (process-get R-proc 'callbacks)
-	      (ess-wait-for-process R-proc)))
-	  (rename-buffer
-	   (if (bufferp session)
-	       (buffer-name session)
-	     (if (stringp session)
-		 session
-	       (buffer-name))))
-	  (current-buffer))))))
+	      (ess-wait-for-process R-proc))
+	    (with-current-buffer R-buf
+	      (rename-buffer
+	       (if (bufferp session)
+		   (buffer-name session)
+		 (if (stringp session)
+		     session
+		   (buffer-name)))))
+	    R-buf))))))
 
 (defun org-babel-R-associate-session (session)
   "Associate R code buffer with an R session.
-- 
2.19.0
```


lisp/ess-inf.el
```diff
@@ -175,38 +175,39 @@ This may be useful for debugging."
...
-            (ess-set-working-directory cur-dir))
-          (setq-local font-lock-fontify-region-function #'inferior-ess-fontify-region)
-          (setq-local ess-sl-modtime-alist nil)
-          (run-hooks 'ess-post-run-hook)
-          ;; User initialization can take some time ...
-          (unless no-wait
-            (ess-write-to-dribble-buffer "(inferior-ess 3): waiting for process after hook")
-            (ess-wait-for-process proc)))
-        inf-buf))))
+      (setq-local ess-local-process-name proc-name)
```
jabranham: This means that ess-local-process-name gets set in whatever buffer R gets called from.

In other words, before this change if you did M-x R from the scratch buffer, `ess-local-process-name` would still be nil in the scratch buffer. After this change, it'll be `R:1` or something like that.

I'm not sure if we should be associating non-R buffers with processes?

lionel-: Good question. I think it still makes sense? You might not be sending R code to the REPL but you still could interact with it in other ways.

jabranham: I worry that a change like this might have unintended/unforeseen consequences. It would be good to test it more. So perhaps with ESS 19.04 approaching, we should split off an ess-19.04 branch and then commit this to master so it's not in 19.04? WDYT?

lionel-: Sounds good.

vspinu:
All looks good to me. Much better this way indeed.

lisp/ess-inf.el
mmaechler: Gut feeling: I also don't like that non-R buffers would be associated with ESS processes.
People who use Emacs for many things, not just R,  will not at all like such a behavior.

lionel-: I think we should set the local variable because this gives a way to be compatible with both the old `set-buffer` effect and the new non-side-effecty return value.

- With old ESS, `set-buffer` is called with the inferior buffer whose `ess-local-process-name` is set to itself. You can thus get the process or its buffer via `ess-local-process-name`.

- With new ESS, the local variable will be set to the current buffer. You can thus get the process or its buffer via `ess-local-process-name`.

Cf my patch for org-mode, it's compatible with both ESS versions because of this.

> Gut feeling: I also don't like that non-R buffers would be associated with ESS processes.
People who use Emacs for many things, not just R, will not at all like such a behavior.

Can you come up with a downside of having a buffer associated with an ESS process? I can't think of any. I don't thing people will dislike this behaviour, they will probably not even notice it. And if they do, it's in a positive way. They can directly interact with their process instead of ESS asking which one to use. If they ever need to use a different process right after having started one, they can `M-x ess-switch`.

Given the importance of being compatible with old and new ESS versions during the transition, I recommend we set the local process on startup.

mmaechler: It just sounds crazy that mail mail reading,  or snake playing :-) , or dired or ... buffer should be connected to an ESS process.
Using R and editing R scripts may be only a small part of what I do at given moment in Emacs.
Isn't it still true that die hard emacs users start one emacs and never leave it for the day and do everything from inside that one instance ?

jabranham: I think I agree with Martin. I oftentimes do M-x R to do a quick calculation or whatnot, not because I want `*scratch*` (or whatever buffer happens to be current) to be associated with an ESS process. 

I think the practical downside would happen if e.g. you did M-x R in `*scratch*` then later wanted to  do some julia stuff. `*scratch*` would think that you wanted an R session still. Like I said above, I'm not sure of all the implications of this, but we can test it out for the post-19.04 release and see if it's disruptive at all.

> Isn't it still true that die hard emacs users start one emacs and never leave it for the day and do everything from inside that one instance ?

Yes, definitely. Emacs is my window manager, so literally everything I do is inside Emacs. :-)

lionel-: I do everything in R as well. I just don't see any practical downside to having a local buffer variable defined. Especially with the current trend of forcing the inferior buffer by default.

> I think the practical downside would happen if e.g. you did M-x R in *scratch* then later wanted to do some julia stuff. *scratch* would think that you wanted an R session still.

Good point, maybe we should hook some inferior forcing/updating in the language mode hooks?

PR #889, opened 2019-04-05
Obsolete `inferior-ess-start-args`

lionel-:
And use language-specific variables instead.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -1781,6 +1781,9 @@ for editing and then to be returned to the process.")
 The other variables ...-program should be changed, for the
 corresponding program.")
 
+(make-obsolete-variable 'inferior-ess-start-args
+                        "Use the language specific variables like `inferior-R-args'"
+                        "ESS 19.04")
 (defvar inferior-ess-start-args ""
   "String of arguments passed to the ESS process.
 If you wish to pass arguments to a process, see e.g. `inferior-R-args'.")
--- a/lisp/ess-stata-mode.el
+++ b/lisp/ess-stata-mode.el
@@ -83,7 +83,6 @@
     (inferior-ess-primary-prompt   . "[.:] \\|--more--") 
     (inferior-ess-secondary-prompt . "--more--")
     (comint-use-prompt-regexp      . t)
-    (inferior-ess-start-args       . inferior-STA-start-args)
     (inferior-ess-search-list-command   . "set more off\n search()\n")
     (comment-start                . "/\* ")
     (comment-end                  . " \*/")
--- a/lisp/ess-gretl.el
+++ b/lisp/ess-gretl.el
@@ -486,7 +486,6 @@ end keywords as associated values.")
     (ess-smart-operators		. ess-r-smart-operators)
     (inferior-ess-exit-command		. "exit\n")
     ;;harmful for shell-mode's C-a: -- but "necessary" for ESS-help?
-    (inferior-ess-start-args		. "")
     (inferior-ess-language-start	. nil)
     (ess-STERM		. "iESS")
     )
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -125,14 +125,13 @@ If `ess-plain-first-buffername', then initial process is number-free."
                         (= n 1))) ; if not both first and plain-first add number
               (concat ":" (number-to-string n)))))
 
-(defun inferior-ess (&optional ess-start-args customize-alist no-wait)
+(defun inferior-ess (&optional start-args customize-alist no-wait)
   "Start inferior ESS process.
-
 Without a prefix argument, starts a new ESS process, or switches
-to the ESS process associated with the current buffer.  With
-ESS-START-ARGS (perhaps specified via \\[universal-argument]),
-starts the process with those args.  The current buffer is used
-if it is an `inferior-ess-mode' or `ess-transcript-mode' buffer.
+to the ESS process associated with the current buffer. With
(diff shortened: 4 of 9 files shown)


lisp/ess-custom.el
```diff
@@ -1781,6 +1781,9 @@ for editing and then to be returned to the process.")
 The other variables ...-program should be changed, for the
 corresponding program.")
 
+(make-obsolete-variable 'inferior-ess-start-args
+                        "Use the language specific variables"
```
jabranham: The message can probably be a bit more helpful like, "Use language-specific variables instead like inferior-R-args".

lisp/ess-stata-mode.el
```diff
@@ -83,7 +83,6 @@
     (inferior-ess-primary-prompt   . "[.:] \\|--more--") 
     (inferior-ess-secondary-prompt . "--more--")
     (comint-use-prompt-regexp      . t)
-    (inferior-ess-start-args       . inferior-STA-start-args)
```
jabranham: Doesn't inferior-STA-start-args neeed to be used somewhere?

lionel-: This definition in the cust-alist was redundant, `inferior-STA-starts-args` is already being passed to `inferior-ess`. This shows how this variable was confusing and it's a good thing to obsolete it.

lisp/ess-custom.el
lionel-: I considered this as well, but I don't like the non-standard namespace here. Should we alias to `inferior-ess-r-start-args`?

lisp/ess-stata-mode.el
jabranham: Ah, I see that now. Thanks!

lisp/ess-custom.el
jabranham: Might as well rename them all while we're at it so that they're more standard. I'd do something like `ess-<lang>-start-args`.

lionel-: Should we have the convention that "inferior" is a technical term for ESS developers that shouldn't leak into the UI?

Dev oriented: `inferior-ess-r-force`
User oriented: `ess-r-start-args`

As a particle, `inferior` suggests we're interacting with the associated process of the current buffer.

jabranham: Makes sense to me :+1:

lionel-: I'm now wondering if we should take the renaming opportunity to update the interface so that `start-args` variables should be `nil` or a list of strings. This might simplify things and make it easier to combine arguments from different sources.

For instance we'd set the default to `'("--no-readline")` and users would add new arguments in their config with `add-to-list`. Then we no longer need `ess-R-readline` in our config UI.

vspinu: I never liked inferior either, but I think it's useful to distinguish those variables if they are strictly relevant to the process. `repl-` would have been better.

vspinu: Completely agree on the argument list BTW.

lionel-: Right, `repl-` or `iess-` would be shorter. But I think for the UI `ess-` should mean "interaction with REPL". For other functionality (help, package, etc) we have other namespaces.

jabranham: `inferior-` has the advantage of being shared across other Emacs packages (e.g. inferior-python-mode)

lionel-: Do we agree that `inferior` should not be exposed to end users in UI variables and functions? For instance, I was planning to rename `inferior-ess-reload` to just `ess-reload`. It is shorter to type, less scary, and more easily searchable on `M-x`.

jabranham: I don't think so. That would require renaming inferior-ess-r-mode and the others since they generate e.g. inferior-ess-r-mode-hook, -map, etc. I'm open to arguments to change my mind, though.

vspinu: I am fine with user commands not containing `inferior`, but something should be there in this particular case. `ess-reload` is too generic. `ess-reload-process` or `ess-reload-repl` maybe?

There are 2 more commands which we might consider renaming then: `inferior-ess-force` and `inferior-ess-send-input`.

vspinu: @jabranham, I think Lionel didn't  mean the mode functions. Renaming those would be troublesome indeed.

lionel-: Good point about the hooks, though I think it would be ok to just alias the hooks without renaming everything.

> There are 2 more commands which we might consider renaming then: inferior-ess-force and inferior-ess-send-input.

I would think these commands are more on the programming side, so I would not rename them.

Let's keep thining about this then.

jabranham: FWIW `inferior-ess-force` is already aliased to `ess-force-buffer-current`

jabranham: (which is also not a great name, if you ask me) :-)

vspinu: We could stick with `-process` in our names for such functionality: `ess-reload-process`, `ess-force-current-process`.

lionel-: hmm maybe. In the particular case of reloading, I think we should follow the existing naming scheme so it's consistent with `ess-quit`.

vspinu: I would generally prefer to be more explicit, but ok. If we leave it without `-process` then how about `-restart` rather than `-reload`? It's a bit more specific and sesman (which I will bring to ESS at some point) already has `sesman-restart` command.

lionel-: `ess-restart` sounds good to me.

PR #890, opened 2019-04-05
Set local process in current buffer on startup

lionel-:
From discussion at https://github.com/emacs-ess/ESS/pull/884#discussion_r272488587

I think we should set the local variable because this gives a way to be compatible with both the old `set-buffer` effect and the new non-side-effecty return value.

- With old ESS, `set-buffer` is called with the inferior buffer whose `ess-local-process-name` is set to itself. You can thus get the process or its buffer via `ess-local-process-name`.

- With new ESS, the local variable will be set in the current buffer. You can thus get the process or its buffer via `ess-local-process-name`.

Cf my patch for org-mode, it's compatible with both ESS versions because of this. Given the importance of being compatible with old and new ESS versions during the transition, I recommend we set the local process on startup.

I think this also makes ESS more useful because users can directly interact with their process instead of ESS asking which one to use. If they ever need to use a different process right after having started one, they can `M-x ess-switch`.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -175,6 +175,7 @@ This may be useful for debugging."
            (proc-name (with-current-buffer inf-buf ess-local-process-name))
            (cur-dir (inferior-ess--maybe-prompt-startup-directory proc-name temp-dialect))
            (default-directory cur-dir))
+      (setq-local ess-local-process-name proc-name)
       (with-current-buffer inf-buf
         (setq-local default-directory cur-dir)
         ;; TODO: Get rid of this, we should rely on modes to set the
--- a/test/ess-test-inf.el
+++ b/test/ess-test-inf.el
@@ -68,6 +68,11 @@
                            (format-message error-msg))))
       (kill-process proc))))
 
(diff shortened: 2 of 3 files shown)


lionel-:
Of relevance, many ESS commands set the local process in the current buffer (search the repo for `setq ess-local-process-name`).

lionel-:
From https://github.com/emacs-ess/ESS/pull/884#discussion_r272645086

Language mode startup should nilify `ess-local-process-name` if the language of the process does not match the language of the mode being started. This way changing mode in the scratch buffer resets the inferior association, and we're not sending Julia code to the R interpreter. This actually solves the language switching problem in general, which current ESS already has.

vspinu:
@lionel- ping. What's holding this one?

lionel-:
Martin and Alex were not enthusiastic about merging this.

vspinu:
I see. Those are valid points.  Let's close then. ess-local-process will be removed with sesman anyhow.

lionel-:
I guess org-mode can special-case based on ESS version.

PR #891, opened 2019-04-05
Don't link function arguments in help buffers

### LINKED ISSUE #855: Don't link arguments from functions in help buffers
e.g. `?lm` shouldn't link `data` since it is probably referring to the argument, not the data function.

See #848

jabranham:
Closes #855

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -2261,9 +2261,14 @@ state.")
 
 (defun ess-r-help-usage-objects ()
   "Return a list of objects in the usage section for the current help buffer.
-In other words, if in the help buffer for \"qt\", return '(\"dt\"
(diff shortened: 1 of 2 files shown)


jabranham:
Travis doesn't like this for whatever reason, but works fine locally for me, both in batch and interactively. I'll take a look later I guess.

vspinu:
`Process R<1> is not running, but others are. Switch? (y or n)` probably means that process exited earlier than expected and there are other processes. Looks like we are leaking processes in our tests somehow.

jabranham:
Indeed. But since I can't reproduce it locally it's hard to track down.

Also, not sure if it's related but `ess-r-package-auto-no-activation-in-shell-test` fails locally for me, but only in batch, not interactively.

jabranham:
"Fixed" the tests by marking that one to run only if interactive. Merged to master.

Issue #738, opened 2018-11-08
Remove customize-alists


jabranham:
Usually this means setting variables in major modes, but may be on a case-by-case basis.

- [ ] remove ess-setq-vars-local 
- [ ] ess-custom.el:2857:(defvar ess-customize-alist nil
- [ ] ess-custom.el:2862:;; (defvaralias 'ess-local-customize-alist 'ess-customize-alist)
- [ ] ess-custom.el:2864:(defvar ess-local-customize-alist nil
- [ ] ess-gretl.el:460:(defvar gretl-customize-alist
- [ ] ess-inf.el:116:(defun inferior-ess (&optional ess-start-args customize-alist no-wait)
- [ ] ess-julia.el:303:(defvar ess-julia-customize-alist
- [ ] ess-r-gui.el:110:(defvar Rgui-customize-alist
- [ ] ess-r-mode.el:333:(define-obsolete-variable-alias 'R-customize-alist 'ess-r-customize-alist "ESS 18.10.2")
- [ ] ess-r-mode.el:334:(defvar ess-r-customize-alist
- [ ] ess-r-package.el:39:(defvar ess-r-customize-alist)
- [ ] ess-rd.el:39:(defvar ess-r-customize-alist)
- [ ] ess-sas-a.el:32:(defvar SAS-customize-alist)
- [ ] ess-sas-d.el:178:(defvar SAS-customize-alist
- [ ] ess-sas-d.el:298:  (setq-default ess-customize-alist SAS-customize-alist)
- [ ] ess-sp6-d.el:69:(defvaralias 'S+6-customize-alist 'S+-customize-alist)
- [ ] ess-sp6-d.el:70:(defvar S+-customize-alist
- [ ] ess-stata-mode.el:66:(defvar STA-customize-alist
- [ ] essd-els.el:53:(defvar S+elsewhere-customize-alist
- [ ] grep through the source code for stuff like `cust alist`

vspinu:
Before we proceed further with this goal, lets compile a list of relevant variables that represent our "state". This will be needed in order to move the state from one buffer to another, which was historically one function of customize alist. 

We move state in a few places already (mainly process name). I think we should replace that with a generic ess-move-state or something.

PR #892, opened 2019-04-05
Remove ess-customize-alist

jabranham:
This removes ess-customize-alist (the non-buffer-local variable).

The changes here will trivially conflict with #889, so I'll hold off on merging until that gets merged so as to not create conflicts for you @lionel-

Progress toward #738

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -101,8 +101,8 @@ been initialised."
         (t
          (progn
            (setq-local inferior-ess-mode-syntax-table
-                       (eval (or (alist-get 'inferior-ess-mode-syntax-table ess-customize-alist)
-                                 (alist-get 'ess-mode-syntax-table ess-customize-alist))))
+                       (eval (or (alist-get 'inferior-ess-mode-syntax-table ess-local-customize-alist)
+                                 (alist-get 'ess-mode-syntax-table ess-local-customize-alist))))
            (inferior-ess-mode)))))
 
  ;;*;; Process handling
@@ -125,7 +125,7 @@ If `ess-plain-first-buffername', then initial process is number-free."
                         (= n 1))) ; if not both first and plain-first add number
               (concat ":" (number-to-string n)))))
 
-(defun inferior-ess (&optional start-args customize-alist no-wait)
+(defun inferior-ess (start-args customize-alist &optional no-wait)
   "Start inferior ESS process.
 Without a prefix argument, starts a new ESS process, or switches
 to the ESS process associated with the current buffer. With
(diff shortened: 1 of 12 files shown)


lisp/ess-inf.el
```diff
@@ -101,8 +101,8 @@ been initialised."
         (t
          (progn
            (setq-local inferior-ess-mode-syntax-table
-                       (eval (or (alist-get 'inferior-ess-mode-syntax-table ess-customize-alist)
-                                 (alist-get 'ess-mode-syntax-table ess-customize-alist))))
+                       (eval (or (alist-get 'inferior-ess-mode-syntax-table ess-local-customize-alist)
```
lionel-: I think this would be cleaner if `inferior-ess--set-major-mode` took `cust-alist` as argument.

lisp/ess-sas-a.el
```diff
@@ -809,14 +809,14 @@ optional argument is non-nil, then set-buffer rather than switch."
   "And now for something completely different."
   (interactive)
   ;;(ess-sas-file-path)
-  (setq ess-customize-alist SAS-customize-alist)
+  (setq ess-local-customize-alist SAS-customize-alist)
```
lionel-: This is redundant and can be removed I think, because `ess-local-customize-alist` is itself part of the cust-alists. Then `ess-setq-vars-local` can be passed `SAS-customize-alist` directly as in the R runner.

lisp/ess-inf.el
```diff
@@ -125,7 +125,7 @@ If `ess-plain-first-buffername', then initial process is number-free."
                         (= n 1))) ; if not both first and plain-first add number
               (concat ":" (number-to-string n)))))
 
-(defun inferior-ess (&optional ess-start-args customize-alist no-wait)
+(defun inferior-ess (ess-start-args customize-alist &optional no-wait)
```
lionel-: In the other PR I have renamed `ess-start-args` to `start-args`. Similarly I think we can use an abbreviation here, e.g. `cust-alist`. This is easier to read and a clearer signal that it's not a global variable like `ess-r-customize-alist` (since global variables have long names).

lionel-:
Should we try to move the `ess-setq-vars-local` until the very end? This will reveal any parameters of process startup that relies on local variables. Then we can add these as explicit arguments of `inferior-ess`.

However I noticed that `inferior-ess--set-major-mode` resets all non-permanent local variables, so if we move it at the end we'll have a different set of variables defined.

lisp/ess-inf.el
jabranham: That's all fine with me. Like I said, I can rebase this once you merge that other PR and resolve the conflicts. :+1:

lionel-: Already merged :)

jabranham: Great, updated here.

lisp/ess-inf.el
jabranham: I'm avoiding using the customize-alists too much since our eventual goal is to remove them.

Really, this last part should never happen. If it does we should implement a proper mode.

PR #893, opened 2019-04-06
Reload process with same process name and start arguments

### LINKED ISSUE #871: Inferior reloading always uses default runner

lionel-:
Closes #871.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -2255,11 +2255,9 @@ START-ARGS gets passed to the dialect-specific
   (interactive)
   (let* ((inf-buf (inferior-ess-force))
          (inf-proc (get-buffer-process inf-buf))
-         (start-name (with-current-buffer inf-buf
-                       inferior-ess--start-name))
+         (start-name (buffer-local-value 'inferior-ess--start-name inf-buf))
          (start-args (or start-args
-                         (with-current-buffer inf-buf
-                           inferior-ess--start-args))))
+                         (buffer-local-value 'inferior-ess--start-args inf-buf))))
     ;; Interrupt early so we can get working directory
     (ess-interrupt)
     (save-window-excursion
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -2253,24 +2253,29 @@ is run automatically by \\[ess-quit]."
 START-ARGS gets passed to the dialect-specific
 `inferior-ess-reload-override'."
   (interactive)
-  (inferior-ess-force)
-  ;; Interrupt early so we can get working directory
-  (ess-interrupt)
-  (save-window-excursion
-    ;; Make sure we don't ask for directory again
-    ;; Use current working directory as default
-    (let ((project-find-functions nil)
-          (ess-directory-function nil)
-          (ess-startup-directory (ess-get-working-directory))
-          (ess-ask-for-ess-directory nil)
-          (proc (ess-get-process)))
-      (ess-quit 'no-save)
-      (inferior-ess--wait-for-exit proc)
-      (with-current-buffer (process-buffer proc)
-        (inferior-ess-reload--override start-args)))))
+  (let* ((inf-buf (inferior-ess-force))
+         (inf-proc (get-buffer-process inf-buf))
+         (start-name (with-current-buffer inf-buf
+                       inferior-ess--start-name))
+         (start-args (or start-args
+                         (with-current-buffer inf-buf
+                           inferior-ess--start-args))))
+    ;; Interrupt early so we can get working directory
+    (ess-interrupt)
+    (save-window-excursion
+      ;; Make sure we don't ask for directory again
+      ;; Use current working directory as default
+      (let ((project-find-functions nil)
+            (ess-directory-function nil)
+            (ess-startup-directory (ess-get-working-directory))
+            (ess-ask-for-ess-directory nil))
+        (ess-quit 'no-save)
+        (inferior-ess--wait-for-exit inf-proc)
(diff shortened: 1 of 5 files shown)


lisp/ess-inf.el
```diff
@@ -125,6 +125,13 @@ If `ess-plain-first-buffername', then initial process is number-free."
                         (= n 1))) ; if not both first and plain-first add number
               (concat ":" (number-to-string n)))))
 
+(defvar-local inferior-ess--start-name nil
+  "Program name used to start the buffer.
+Buffer-local in inferior buffers.")
```
jabranham: No need to say that it's buffer local if you're defining it with defvar-local. It'll already have that in the docstring.

lisp/ess-inf.el
```diff
@@ -2269,24 +2280,29 @@ is run automatically by \\[ess-quit]."
...
-      (ess-quit 'no-save)
-      (inferior-ess--wait-for-exit proc)
-      (with-current-buffer (process-buffer proc)
-        (inferior-ess-reload--override start-args)))))
-
-(cl-defmethod inferior-ess-reload--override (start-args)
-  (user-error "Reloading not implemented for %s %s" ess-dialect start-args))
+  (let* ((inf-buf (inferior-ess-force))
+         (inf-proc (get-buffer-process inf-buf))
+         (start-name (with-current-buffer inf-buf
```
jabranham: `buffer-local-value` is /much/ faster than `with-current-buffer`, if the only thing you need is the value of a variable.

test/ess-test-inf.el
```diff
@@ -68,6 +68,24 @@
                            (format-message error-msg))))
       (kill-process proc))))
 
+(ert-deftest ess-test-inferior-local-start-args ()
+  (with-r-running nil
+    (with-current-buffer *inf-buf*
```
jabranham: Here's another instance where `buffer-local-value` would be better, I think.

lisp/ess-r-mode.el
```diff
@@ -1363,11 +1363,12 @@ selected (see `ess-r-set-evaluation-env')."
...
   :group 'ess-R)
 
-(cl-defmethod inferior-ess-reload--override (start-args
+(cl-defmethod inferior-ess-reload--override (start-name start-args
                                              &context ((string= ess-dialect "R") (eql t)))
   "Call `run-ess-r' with START-ARGS.
 Then run `inferior-ess-r-reload-hook'."
-  (run-ess-r start-args)
+  (let ((inferior-ess-r-program start-name))
+    (run-ess-r start-args))
```
jabranham: Can't we use `inferior-ess--start-args` here and avoid needing another argument to the function?

lionel-: Did you mean `--start-name`?

lisp/ess-inf.el
lionel-: Ah thanks, I didn't know about this one.

lisp/ess-r-mode.el
lionel-: If `inferior-ess-reload--override` becomes a public customisation point (i.e. moves to `inferior-ess-reload-override`), it can't use private data. At that point it might make sense to bundle inferior metadata in an alist so we only need to pass a single argument. Until then it's probably best to keep the data explicit.

But I think the program name will eventually be passed as explicit argument to `inferior-ess` rather than through the cust-alist, so this reload signature would mirror the startup signature nicely.

jabranham:
I've been testing this out and noticed two things:

1) I've occasionally received an error when trying to start a process. I can't reproduce it reliably yet, but it looks like nil is getting passed to make-process somehow.

2) If you open a script, do M-x R-3.2.1 to start a specific R version, then do C-c C-q to stop that process, then do C-RET to start the process again, it switches to using R instead of R-3.2.1 in the *R:1* buffer.

lionel-:
Thanks for trying! I'll look into the `C-c C-q` behaviour. I haven't seen any errors on process startup. Is this on normal startup or on reload startup?

lisp/ess-r-mode.el
vspinu: Do we really need this overwrite? How about saving the actual runner as a local as well and re-call it?

lisp/ess-inf.el
```diff
@@ -125,6 +125,11 @@ If `ess-plain-first-buffername', then initial process is number-free."
                         (= n 1))) ; if not both first and plain-first add number
               (concat ":" (number-to-string n)))))
 
+(defvar-local inferior-ess--start-name nil
+  "Program name used to start the inferior process.")
```
vspinu: Shall we put this into one internal variable?

lionel-: Done.

lisp/ess-r-mode.el
lionel-: This could probably work, but I prefer to keep this PR small.

lionel-:
> If you open a script, do M-x R-3.2.1 to start a specific R version, then do C-c C-q to stop that process, then do C-RET to start the process again, it switches to using R instead of R-3.2.1 in the R:1 buffer.

What is `C-RET` bound to in your config? It isn't in mine.

lionel-:
Merging this so we can make progress, let me know of any further issues.

jabranham:
C-RET should be bound to `ess-eval-region-or-line-visibly-and-step` in ess-r-mode buffers.

PR #895, opened 2019-04-11
Unbundle julia-mode

jabranham:
This removes the awkward julia-mode download from our build system and just assumes julia-mode is present. It gets around issues like we have in the ess-18.10 branch which is currently unusable since julia-mode changed from a single file to multiple files.

The cost is of course that people who previously installed ESS with a tarball will now need to install julia-mode before doing so, see #780.

It doesn't affect MELPA users, as they never used our build system; they get julia-mode from MELPA.

### DIFF
--- a/Makeconf
+++ b/Makeconf
@@ -1,30 +1,39 @@
-####======= Common Declarations for *all* ESS -*-Makefile-*-s ==========
-
 ## To install ESS for all users on your unix system:
-## 1. Edit Section 1 and 2 below
-## 2. Execute: gmake install
+## 1. Check variables below, edit as necessary
+## 2. make
+## 3. make install
 
 ## Section 1
-## Installation variables for your emacs variant
+## Installation variables for your Emacs variant
 ##
 ## Variable        Description
 ## EMACS           Path to GNU Emacs
-## EMACSBATCH      Emacs command for compiling elisp files
-## SITELISP        Destination of site-lisp; but some variants have no site-lisp
-## LISPDIR         Destination of compiled elisp files
-## INFODIR         Destination of info files
-## ETCDIR          Destination of script and icon files
-## PREFIX(DESTDIR) Directory prepended to SITELISP, INFODIR, DOCDIR & ETCDIR
-##                 Specify either PREFIX or DESTDIR to over-ride /usr
-DESTDIR=/usr
-PREFIX=$(DESTDIR)
+## EMACSBATCH      How to run Emacs as batch
+## SITELISP        Destination of site-lisp
+## ESSDIR          Destination of ESS
+## JULIA_DIR       Path to julia-mode
+DESTDIR ?= /usr
+PREFIX ?= $(DESTDIR)
 
 ##__ GNU Emacs __
 EMACS ?= emacs
-SITELISP=$(PREFIX)/share/emacs/site-lisp
-LISPDIR=$(SITELISP)/ess
-INFODIR=$(PREFIX)/share/info
-ETCDIR =$(PREFIX)/share/emacs/etc/ess
+SITELISP ?= $(PREFIX)/share/emacs/site-lisp
+ESSDIR ?= $(SITELISP)/ess
+EMACSBATCH ?= $(EMACS) -batch -Q -L $(JULIA_DIR)
+
+# Load path:
+ifndef JULIA_DIR
+
+ELPA_DIR ?= ~/.emacs.d/elpa
+
+JULIA_DIR ?= $(shell \
+  find -L $(ELPA_DIR) -maxdepth 1 -regex '.*/julia-mode-[.0-9]*' 2> /dev/null | \
+  sort | tail -n 1)
+ifeq "$(JULIA_DIR)" ""
+  JULIA_DIR = $(SITELISP)
+endif
+
+endif
 
 ##__ GNU Emacs __  for Mac OS X with NeXTstep (Cocoa or GNUstep)
 # PREFIX=/Applications/Emacs.app/Contents/Resources
@@ -42,13 +51,14 @@ ETCDIR =$(PREFIX)/share/emacs/etc/ess
 # INFODIR=/Applications/Aquamacs.app/Contents/Resources/info
 # ETCDIR=$(PREFIX)/ess-mode/etc
 
-EMACSBATCH ?= $(EMACS) -batch -Q
+## COMPILE-FLAGS is set in tests
+COMPILE ?= $(EMACSBATCH) $(COMPILE-FLAGS) --directory . --directory ./obsolete -f batch-byte-compile
+COMPILE-SIMPLE ?= $(EMACSBATCH) --directory . --directory ./obsolete -f batch-byte-compile
 
 ## Section 2
 ## Installation variables for your unix variant
 ##
 ## Variable        Description
-## SHELL           Bourne shell or XPG4 compliant shell
 ## MAKEINFO        program to convert .texi{nfo} to .info
 ## MAKEHTML        program to convert .texi{nfo} to .html
 ## MAKETXT         program to convert .texi{nfo} to .txt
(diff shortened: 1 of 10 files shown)


Makeconf
```diff
@@ -1,30 +1,39 @@
...
-ETCDIR =$(PREFIX)/share/emacs/etc/ess
+SITELISP ?= $(PREFIX)/share/emacs/site-lisp
+ESSDIR ?= $(SITELISP)/ess
+EMACSBATCH ?= $(EMACS) -batch -Q -L $(JULIA_DIR)
+
+# Load path:
+ifndef JULIA_DIR
+
+ELPA_DIR ?= ~/.emacs.d/elpa
+
```
vspinu: This will break all those for whom elpa is not there, including myself. 

How about creating local directory .ELPA and fetch julia there if not found? I have a simple [setup](https://github.com/polymode/poly-R/blob/master/targets/melpa.el) in polymode to do so. Only a fraction of that is needed here.

Makeconf
```diff
@@ -1,30 +1,39 @@
...
+PREFIX ?= $(DESTDIR)
 
 ##__ GNU Emacs __
 EMACS ?= emacs
-SITELISP=$(PREFIX)/share/emacs/site-lisp
-LISPDIR=$(SITELISP)/ess
-INFODIR=$(PREFIX)/share/info
-ETCDIR =$(PREFIX)/share/emacs/etc/ess
+SITELISP ?= $(PREFIX)/share/emacs/site-lisp
+ESSDIR ?= $(SITELISP)/ess
```
vspinu: Can we move this to makefile proper not in makeconf? Would be nice to get rid of makeconf at some point.

Makeconf
```diff
@@ -59,29 +69,15 @@ EMACSBATCH ?= $(EMACS) -batch -Q
...
-UNINSTALL = rm -f
-
-DOCDIR=$(PREFIX)/share/doc/ess
-
-# N.B. $(shell) is a GNU-ism: we need a workaround to be UNIX make friendly
-DOWNLOAD=$(shell which wget > /dev/null && echo 'wget -qO -' || which curl)
-##DOWNLOAD = wget -O -
-##DOWNLOAD = curl
+MAKEINFO ?= LC_ALL=C LANG=en makeinfo
+MAKEHTML ?= $(MAKEINFO) --html --no-split --css-include=atouchofstyle.css -o
```
vspinu: Same here. Why not moving all these to Makefile proper?

Makeconf
vspinu: Actually we can unify it with the travis dependency fetcher.

 I think relying on global .emacs.d/elpa is not good idea because people can test with multiple version of emacs and will run into byte compilation issues.

jabranham: > This will break all those for whom elpa is not there, including myself.

I don't have julia-mode at ~/.emacs.d/elpa either, it's just what I figured was most common. You can tell the Makefile where julia-mode is at by passing the variable JULIA_DIR. 

I really do not like the idea of accessing the network. I oftentimes work without internet access and it's annoying not to be able to run ESS's tests just because I can't download julia.

Makeconf
jabranham: They're variables that users/distros can override, and we also need them in other Makefiles (e.g. lisp/Makefile, doc/Makefile). We can't include the top-level makefile in those sub-makefiles. What do you suggest, exactly?

vspinu: > We can't include the top-level makefile

Right. Forgot about those.

Makeconf
vspinu: Well, either way without the network one won't be able to test/compile ESS, am I right? Except that in the proposed solution one has to install julia manually and emacs version conflicts are likely. Do I understand your proposal correctly?

With what I propose one would need to have network access once, as the .ELPA dir is never deleted unless requested.

> You can tell the Makefile where julia-mode is at by passing the variable JULIA_DIR.

Sorry, but this is a no-go AFAIC. Currently one just has to `export EMACS=...` before `make ...`. I don't think we want to complicate that.

jabranham: > Well, either way without the network one won't be able to test/compile ESS, am I right? Except that in the proposed solution one has to install julia manually and emacs version conflicts are likely. Do I understand your proposal correctly?

I don't think so? My solution doesn't require network access, it does assume you've got julia-mode somewhere. 

By Emacs version conflicts, are you talking about issues with loading elc files compiled with e.g. Emacs 26 on Emacs 25? That's possible, I guess. Not sure how likely it is, and we can fix it if/when it becomes an issue.

> With what I propose one would need to have network access once, as the .ELPA dir is never deleted unless requested.

But we also have 0 control over what version of julia-mode is in there. We also have to keep maintaining some mechanism of downloading julia-mode, compiling it, putting it on load-path, etc.

mmaechler:
We spent a bit of an effort to make this (download, adding to tarball,..) working conveniently with the Makefile.  I'd be happy for *improving* that, rather than *removing* it :
  ESS users have learned that ESS is about "Emacs Speaks Statistics" not just "emacs speaks R" (or "emacs speaks to an API which can be used to speak to stats packages)..  Julia is close to R in some sense and FOSS.
Only we as ESS developers need to have internet connection once to automatically get julia-mode.el.
Once you have it there,  you only need to remember to occasionally update it.

I don't see the big deal here.   Assuming that all ESS users are fluent about ELPA, MELPA, etc is just very restrictive... and does not apply to potential newcomers, e.g. our students who are introduced to R and Rstudio and to some of whom I'd be interested to show the ESS alternative.
Similarly to colleagues and staff out there in companies.  Quite a few of them would start using emacs only because they are told to do so for using ESS.    For all those, unpacking a zip file is normal and hence acceptable.
Having to know about 'M-x' or even emacs lisp expressions to enter somewhere in emacs is out of the question for such newcomers... and the will never "come".

lionel-:
julia-mode will be packaged with ESS+ right? Maybe we should wait until we actually have an ESS+ repo from which to make releases?

Makeconf
vspinu: > I don't think so? My solution doesn't require network access, it does assume you've got julia-mode somewhere.

This does mean that the user needs network to download julia-mode for the first time. Just like with my proposal. 

>  Not sure how likely it is, and we can fix it if/when it becomes an issue.

Very likely if I installed julia-mode on emacs 27 and then test with emacs 25, which I do all the time. 

> But we also have 0 control over what version of julia-mode is in there

Melpa version. Each emacs version will have a sub-dir in ELPA dir with it's own compiled julia-mode. It's a full-control AFAIC. Big advantage of this is that if we have other dependencies (and we will have for ESS+) then those will be fetched there as well. It's a mini-cask in 20 lines of code if you want. 

> We also have to keep maintaining some mechanism of downloading julia-mode

It's the same mechanism that you use on travis - melpa.

vspinu:
This is indeed very much related to ESS+. I propose that we fetch all dependencies into `./lisp/contrib/` from MELPA and include contrib into the ESS+ tar-ball.

I do agree with @jabranham  that once ESS+ is there, we should not bundle julia-mode with the plain ESS anymore.

Makeconf
jabranham: > This does mean that the user needs network to download julia-mode for the first time. Just like with my proposal.

No, they could get it from a flash drive or other media.

>> Not sure how likely it is, and we can fix it if/when it becomes an issue.

> Very likely if I installed julia-mode on emacs 27 and then test with emacs 25, which I do all the time.

Isn't this an issue with the current setup, though?

>> But we also have 0 control over what version of julia-mode is in there

> Melpa version. Each emacs version will have a sub-dir in ELPA dir with it's own compiled julia-mode. It's a full-control AFAIC

It's not full control at all. Let's say I'm trying to track down a bug in ess-julia and I want to test against different versions of julia-mode. That's not easy/possible if we just rely on MELPA. 

Using MELPA is also a big security risk since it's downloading and running unreviewed code from the internet. I'd really prefer a solution with at least the option of using a user-specified julia-mode.

vspinu: Arh. We are just crazy.  We just do git submodules for our deps and be done with all the complexities.

slyrus:
Unsolicited input from a mostly-lurker, so feel free to ignore. I'm often puzzled by all of this discussion of MELPA, ELPA, tarballs, etc... I expect to get the source via git, place it in my .emacs.d/site-lisp directory, do "make clean all" and have things magically work. If I wanted ESS to work with julia I'd expect that the ESS makefile stuff would look for it in my .emacs directory, do the right thing if it finds it, and not build ESS support for julia otherwise. Having the makefile wget/curl download the julia mode seems blatantly "wrong" IMO and means I can't build ESS (without going through my usual hoop of deleting the julia stuff) when I'm offline -- and continuing to have access to a local repo via git is one of its big wins. Just my two cents.

jabranham:
> I'd expect that the ESS makefile stuff would look for it in my .emacs directory, do the right thing if it finds it, and not build ESS support for julia otherwise.

That sounds like we'd need autoconf or something like it, and I'm not sure we want that added complexity. It's much easier (and more normal in the land of elisp packaging) to just assume all dependencies are available.

vspinu:
@jabranham 

Could you please merge all the unrelated cleanups from this one? 

Regarding the core issue I think the git submoduels or subtrees are the right way to go about this one, just like everyone else does it.  Also for ESS+. Let's make that change as part of ESS+ proposal.

lionel-:
If we go the submodule way, I really think ESS+ should be in a separate repo so all the submodules (including ESS) are gathered there.

PR #899, opened 2019-04-19
Ensure ess-dialect is set in the inferior buffer

### LINKED ISSUE #897: inferior R use .history, not .Rhistory in Window 10
When I start R, it reads and writes .history as history file, instead of .Rhistory
comint-input-ring-file-name in the inferior R buffer is ".../.history".
If I run the code below, it gets ".Rhistory". It seems the ess-dialect not correctly set before the comint-input-ring-file-name.
```elisp
(expand-file-name (if (eql t ess-history-file)
                                (concat "." ess-dialect "history")
                              ess-history-file)
                            ess-history-directory)
```

R version 3.5.3 (2019-03-11)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 15063)

jabranham:
Adds a second argument to 'inferior-ess--get-proc-buffer-create' and
sets 'ess-dialect'. Cleans up inferior-ess a bit, mostly by removing a
few unneeded function calls and a now-unneeded nested let-binding.

Closes #897

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -158,60 +158,54 @@ This may be useful for debugging."
   ;; If no transfile, use buffer *S*
   ;; This function is primarily used to figure out the Process and
   ;; buffer names to use for inferior-ess.
-  (let* ((temp-ess-dialect (eval (cdr (assoc 'ess-dialect customize-alist)))))
-    (run-hooks 'ess-pre-run-hook)
-    (ess-write-to-dribble-buffer
-     (format "(inf-ess 1): lang=%s, dialect=%s, tmp-dialect=%s, buf=%s\n"
-             ess-language ess-dialect temp-ess-dialect (current-buffer)))
-    (let* ((process-environment process-environment)
-           ;; Use temp-ess-dialect if not R, R program name otherwise
-           (temp-dialect (if ess-use-inferior-program-in-buffer-name ;VS[23-02-2013]: fixme: this should not be here
-                             (if (string-equal temp-ess-dialect "R")
-                                 (file-name-nondirectory inferior-ess-r-program)
-                               temp-ess-dialect)
-                           temp-ess-dialect))
-           (inf-buf (inferior-ess--get-proc-buffer-create temp-dialect))
-           (proc-name (buffer-local-value 'ess-local-process-name inf-buf))
-           (cur-dir (inferior-ess--maybe-prompt-startup-directory proc-name temp-dialect))
-           (default-directory cur-dir))
-      (with-current-buffer inf-buf
-        (setq-local default-directory cur-dir)
-        ;; TODO: Get rid of this, we should rely on modes to set the
-        ;; variables they need.
-        (ess-setq-vars-local customize-alist)
-        (inferior-ess--set-major-mode ess-dialect)
-        ;; Show the buffer
-        ;; TODO: Remove inferior-ess-own-frame after ESS 19.04, then just have:
-        ;; (pop-to-buffer inf-buf)
-        (pop-to-buffer inf-buf (with-no-warnings
-                                 (when inferior-ess-own-frame
-                                   '(display-buffer-pop-up-frame))))
-        (let ((proc (inferior-ess--start-process inf-buf proc-name start-args)))
-          (ess-make-buffer-current)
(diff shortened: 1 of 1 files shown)


jabranham:
After thinking about this a bit, I'm not sure this fixes the issue. Will try a different approach.

I'll still push the small cleanup bit though.

lisp/ess-inf.el
```diff
@@ -158,60 +158,54 @@ This may be useful for debugging."
...
+         (inf-buf (inferior-ess--get-proc-buffer-create program-name dialect))
+         (proc-name (buffer-local-value 'ess-local-process-name inf-buf))
+         (cur-dir (inferior-ess--maybe-prompt-startup-directory proc-name dialect)))
+    ;; Show the buffer and select its window.
+    ;; TODO: Remove inferior-ess-own-frame after ESS 19.04, then just have:
+    ;; (pop-to-buffer inf-buf)
+    (pop-to-buffer inf-buf (with-no-warnings
+                             (when inferior-ess-own-frame
+                               '(display-buffer-pop-up-frame))))
+    (setq default-directory cur-dir)
```
lionel-: oh you renamed all the `setq-local` to the more opaque `setq` :(

jabranham: I find setq-local much more opaque than setq here. default-directory /has/ to be a buffer-local variable so it seems really odd to use setq-local. I don't think I'm alone here - in the whole Emacs codebase, I can't find a single instance of `setq-local default-directory`, but there's plenty of `setq default-directory`.

That said, I haven't pushed this change yet.

lionel-: This isn't surprising since `setq-local` is a recent addition to Emacs. The value of using `setq-local` increases when the variable is less familiar to the reader. Of course everyone knows `default-directory` is buffer-local, but consistency is important.

I don't understand how `setq-local` is more opaque, even with a well known buffer-local variable like `default-directory`. It's an extra indication of the reach of the side effect, so I think it is less opaque.

jabranham: >  setq-local is a recent addition to Emacs.

I'm not sure ~7 years counts as "recent" :-)

> consistency is important.

Sure, but so is efficiency

> I don't understand how setq-local is more opaque

setq-local is equivalent to doing `make-local-variable` then `setq`. But calling `make-local-variable` on a variable that is already local is confusing; why would you want to do that?

> It's an extra indication of the reach of the side effect

What side effect? There's none here, since it's already buffer-local.

If we do move towards using setq-local for buffer-local variables, we should at least use it everywhere, consistently. And remove the defvar-local calls we have as well since they're meaningless then.

lionel-: > I'm not sure ~7 years counts as "recent" :-)

I meant in comparison to Emacs and `default-directory` (I should have added the comment below another variable though). Old habits die hard, code isn't generally changed just for the sake of consistency, new conventions take time to emerge, and people need convincing ;)

> What side effect?

The side effect of changing a variable. `setq-local` has clearly less reach than `setq`, even if you don't know anything about the variable being changed.

I'm not convinced by the efficiency argument here, it should only matter in very tight loops / performance critical functions. Readability is more important in most cases.

>  And remove the defvar-local calls we have as well since they're meaningless then.

They're not meaningless, I see those as a safeguard and as self-documentation for the expected scope of the variable.

vspinu: One issue with setq-local is that one cannot use multiple assignments with it. I know that @lionel-  doesn't like multi-setqs either, but I personally see those as a cleaner alternative.

lionel-: Let's settle for `setq` then since I'm in the minority. One alternative is to prefix variables that are local only with `local-`.

vspinu: It might not be a crazy idea to fontify buffer-local variables slightly differently. I wouldn't be surprised if someone already wrote a package for that.

vspinu: I gave it a bit more thought and I think it's important to reserve `setq-local` to signal that a non-local variable is made local with make-local-variable the exact replacement of `(set (make-local-variable ...`. It surely was the original intent of the author of that function and it's how it used inside emacs. So let's stick with setq indeed.

PR #900, opened 2019-04-19
Make 'ess-process-name-list' a list of strings

jabranham:
Also rename 'update-ess-process-name-list' to
'ess-update-process-name-list'.

Tagging @lionel- since you initially added the FIXME

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -489,10 +489,8 @@ This marks the process with a message, at a particular time point."
            (format "\nProcess %s %s at %s\n"
                    (process-name proc) message (current-time-string))))))))
 
-;; FIXME: This list is structured as '(("R:2") ("R")). It doesn't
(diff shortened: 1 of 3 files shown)


lionel-:
If we're changing the internals without thinking about backward compat, we should namespace these symbols in `ess--`. I'm sligthly worried that people might be using those for managing and querying inferiors, either in packages or in scripts. Perhaps we should also think about whether a list of inferior buffers might not be a more appropriate representation. In general it feels like such a refactoring should be part of a larger rethinking of inferior management, otherwise it's not necessarily worth it.

jabranham:
I wonder if it makes sense to obsolete that variable and just dynamically generate that list as needed?

vspinu:
I would say we either make it go away completely or don't touch it for now. That's a rudimentary process tracking which is done by sesman and will go away anyhow.

jabranham:
OK, I will close this then.

PR #901, opened 2019-04-22
use with-current-buffer when initializing inferior julia

nverno:
`run-ess-julia` is failing because it seems to be running code in the source
buffer that should be running the the inferior buffer.  This just wraps the initialization
code in a `with-current-buffer` to run in the inferior.

### DIFF
--- a/lisp/ess-julia.el
+++ b/lisp/ess-julia.el
@@ -432,20 +432,20 @@ always be passed to julia, put them in the variable
                                  " ? "))
 		              nil))))
       (let ((inf-buf (inferior-ess jl-start-args ess-julia-customize-alist)))
-        (ess--tb-start)
-        ;; Remove ` from julia's logo
-        (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


lionel-:
oops, thanks! We definitely should add Julia to our CI checks.

jabranham:
I've thought (briefly) about adding Julia to the CI checks, but I fear/assume Travis has an old, super outdated version of Julia.

nverno:
I think you could just get a 1.0 or greater binary from https://julialang-s3.julialang.org/bin/linux/x64/1.0/julia-1.0.2-linux-x86_64.tar.gz.  It would probably speed things up to have a docker image with R/tex/julia or something, but I haven't experimented with docker much at this point.

jabranham:
@nverno Neat, thanks.

This would probably also necessitate making our R tests conditional so that they skip if R isn't installed. I'll open a new issue to track this.

PR #904, opened 2019-04-23
Refactor R-newest into ess-run-newest-r

jabranham:
- Rename R-newest to ess-run-newest-r. Refactor so that instead of
  setting its own function slot it simply let-binds
  inferior-ess-r-program to the newest version of R found.

- Remove ess-newest-r. It has a confusing name and it only used in one
  place.

- Remove ess-current-R-version. Unused.

- Remove ess-current-R-at-least. Unused.

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -478,7 +478,7 @@ will be prompted to enter arguments interactively."
               ;; TODO: Once we drop Emacs 26 support, can probably
               ;; just use the REMOTE argument of `executable-find'.
               (executable-find inferior-ess-r-program))
-    (display-warning 'ess (format "%s could not be found on the system. Try running `R-newest' instead, which searches your system for R." inferior-ess-r-program) :error)
+    (display-warning 'ess (format "%s could not be found on the system. Try running `run-ess-r-newest' instead, which searches your system for R." inferior-ess-r-program) :error)
     (user-error "%s program not found" inferior-ess-r-program))
   (let* ((r-always-arg
           (if (or ess-microsoft-p (eq system-type 'cygwin))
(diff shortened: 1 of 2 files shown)


lionel-:
hmm, should this be `run-ess-r-newest`? The `run-` namespace is normally leading, e.g. `run-python` or `run-julia`.

mmaechler:
Just because of "cleanness" I don't think it is feasible to rename such a widely used user function.  The 
`M-x R-<foo>`  are used much and by many people.

jabranham:
@lionel- I'm always in favor of prefixing everything with ess-. I think it aids in discoverability. But since we've already got run-ess-r and run-ess-julia, I suppose run-ess-r-newest is fine here.

@mmaechler - R-newest is aliased to the new function; I'll update the commit message to mention that.

lionel-:
> I think it aids in discoverability. 

The `run-` namespace also aids in discoverability though.

jabranham:
> The `run-` namespace also aids in discoverability though.

Fair enough. The new commit I pushed uses that name.

lionel-:
Thanks! Sorry to ask again, but shouldn't it be `run-ess-r-newest` rather than `run-ess-newest-r`? The latter doesn't seem properly namespaced.

jabranham:
Indeed it should. Done, thanks!

lisp/ess-r-mode.el
```diff
@@ -870,26 +855,6 @@ returned."
       (setq date (match-string 2 ver-string)))
     (cons date rver)))
 
-(defun ess-current-R-version ()
-  "Get the version of R currently running in the ESS buffer as a string."
-  (ess-make-buffer-current)
-  (car (ess-get-words-from-vector "as.character(.ess.Rversion)\n")))
-
-(defun ess-current-R-at-least (version)
-  "Is the version of R (in the ESS buffer) at least (\">=\") VERSION ?
```
vspinu: Why are these functions removed? These are public functions designed to retrieve versions from R. They might be used by people's code.

jabranham: I thought that they were probably used at one point by ESS but weren't any longer. If they're meant to be external, I can bring them back easily enough. Just let me know if you want me to (or feel free to do so yourself).

vspinu: I would say we keep the first two. At some point even we might need them and that would mean re-inventing the code.

PR #905, opened 2019-04-25
Add a command to insert a pipe

jabranham:
Current keybinding in ess-r-mode-map is C-c RET. I wanted to use C-c C-m (m for magrittr) but I forgot C-m redirected to RET. C-c RET seems reasonable though. I'm open to others' suggestions.

The exact behavior is governed by ess-r-insert-pipe-behavior. It's a list that can contain elements like eol (for end-of-line) and newline (which also indents).

I hope this behavior is satisfactory to everyone, let me know what you think.

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -251,6 +251,7 @@ value by using `ess-r-runners-reset'."
 (defvar ess-r-mode-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-c C-=") #'ess-cycle-assign)
+    (define-key map (kbd "C-c RET") #'ess-r-insert-pipe)
     map))
 
 (defvar ess-r-mode-syntax-table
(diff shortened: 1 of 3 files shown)


vspinu:
I still think that eol and newline should not be the default. For sure not eol; it's too non-standard.

I would be fine with the newline being the default if `C-c RET RET` would cycle immediately to the plain version. I can push a cycling version on top of this PR if you agree.

jabranham:
> For sure not eol; it's too non-standard

Removed. 

> I would be fine with the newline being the default if C-c RET RET would cycle immediately to the plain version

Done.

>  I can push a cycling version on top of this PR if you agree.

Thanks. I had to clean up some merge conflicts anyway so I just went ahead and put that in.

lisp/ess-r-mode.el
```diff
@@ -941,6 +942,42 @@ use \"bin/Rterm.exe\"."
...
+checking for multiple pipes), use \\[universal-argument].
+
+This command sets a transient keymap so that you may easily
+repeat it. Repeated invocations call `delete-indentation'. So,
+for example, if bound to \\<ess-r-mode-map>\\[ess-r-insert-pipe],
+and `ess-r-insert-pipe-behavior' is (newline), you can press the
+last key twice to delete the newline easily."
+  (interactive "P")
+  (cond (literally (insert "%>%"))
+        ((equal this-command last-command) (delete-indentation))
```
vspinu: This is not quite what I had in mind. C-c RET RET will not work. 

I would like `ess-r-insert-pipe-behavior` to be a list of lists which users can customize as they want. It might contain more than 2 entries. In fact i think renaming into `ess-r-insert-pipe-style` and using explicit keywords :eol-newline-%>%, :eol-%>%. maggritr has other pipes as well. So this could be refactored into a generic cycling inserted of R %ops%.

jabranham: Ah. Well, that's much different than this implementation. I'll close this then.

vspinu: Not really. I just wanted to put an extra commit on top of it. Can you re-add it?

Issue #394, opened 2017-01-27
more julia REPL features?


phaverty:
It would be great to get more of the julia REPL features working in the ESS julia prompt.  Going to the shell with ; or RCall with $ or tab completion of unicode like \mu are the first things that come to mind.

antoine-levitt:
Here's for latex completion: https://github.com/emacs-ess/ESS/pull/420. I second the other features.

PR #909, opened 2019-04-30
Set comint-prompt-regexp directly in inferior-ess-julia-mode

### LINKED ISSUE #908: M-p breaks command prompt in inferior ESS for Julia
M-p (comint-previous-input) breaks command prompt in inferior ESS for Julia.

Steps to reproduce:

1. Launch an inferior ESS terminal for Julia
2. Evaluate some expression (e.g. type `1` and hit `RET`)
3. Press `M-p` (instead of getting `julia> 1` I get `1julia>`)

jabranham:
M-p (comint-previous-input) misbehaves unless the entire prompt is
matched. Move the setting into inferior-ess-julia-mode to minimize
reliance on the various -customize-alist variables.

Fixes #908

### DIFF
--- a/lisp/ess-julia.el
+++ b/lisp/ess-julia.el
@@ -291,7 +291,7 @@ to look up any doc strings."
 
 ;;; CORE
 (defvar ess-julia-customize-alist
-  '((inferior-ess-primary-prompt   . "a> ") ;; from julia>
+  '((inferior-ess-primary-prompt   . "\\w*> ")
     (inferior-ess-secondary-prompt . nil)
     (inferior-ess-prompt           . "\\w*> ")
     (ess-local-customize-alist     . 'ess-julia-customize-alist)
@@ -387,6 +387,7 @@ It makes underscores and dots word constituent chars.")
   :group 'ess-Julia
   (ess-setq-vars-local ess-julia-customize-alist)
   (setq-local comint-use-prompt-regexp t)
+  (setq comint-prompt-regexp (concat "^" inferior-ess-prompt))
   (setq ess-dialect "julia")
   ;; eldoc
   (add-function :before-until (local 'eldoc-documentation-function)


vspinu:
Those ess-inferior-prompt are there for a reason. Not sure if they are used for julia, though. This is likely to break some generic functionality now or in the future.

jabranham:
OK, I kept those there and merged.

nverno:
It would be nice not to rely on prompt regexps in case the julia repl is expanded to incorporate its other modes - eg pkg, help, etc.  This may never happen though, so probably not important.

nverno:
I suppose a possible alternative (probably just messier) would be an advice around `comint-previous-input`

lisp/ess-julia.el
```diff
@@ -387,6 +387,7 @@ It makes underscores and dots word constituent chars.")
   :group 'ess-Julia
   (ess-setq-vars-local ess-julia-customize-alist)
   (setq-local comint-use-prompt-regexp t)
+  (setq comint-prompt-regexp (concat "^" inferior-ess-prompt))
   (setq ess-dialect "julia")
```
vspinu: This one is not necessary, we do it already in `inferior-ess-mode`.

jabranham:
> It would be nice not to rely on prompt regexps in case the julia repl is expanded to incorporate its other modes - eg pkg, help, etc.  This may never happen though, so probably not important.

This would require significant changes to comint too. I'm not sure the best way to handle julia's other modes.

nverno:
Couldn't it be achieved using text properties?  The REPL additions would be a bit of work, I had been considering using conditional keys using a menu filter so `?`, `]`, etc. are only active after the prompt and could be used to change REPL modes.

jabranham:
> Couldn't it be achieved using text properties?  

I'm not sure what you mean by "it". If you mean julia switching to Pkg mode when you press "]" at the prompt, no, I don't think it can. Julia isn't activating ] detection in Emacs's shell (e.g. M-x shell julia RET then try ]).

> The REPL additions would be a bit of work, I had been considering using conditional keys using a menu filter so `?`, `]`, etc. are only active after the prompt and could be used to change REPL modes.

Yes, this could work.

nverno:
No, by 'it' I meant using text properties/overlays to mark comint prompts. 

The REPL mode switching I was thinking about using conditional keys - comint mode wouldn't be sending character input like term mode, so the mode switching would be handled by the conditional bindings.  Anyway, this was just spitballing, maybe I'll work on it if I start actively using Julia again.

jabranham:
> No, by 'it' I meant using text properties/overlays to mark comint prompts. 

Ah, OK. This is basically what comint does by default; ESS overrides this to use regular expressions for prompt detection (see 'comint-use-prompt-regexp').

> The REPL mode switching I was thinking about using conditional keys - comint mode wouldn't be sending character input like term mode, so the mode switching would be handled by the conditional bindings.  Anyway, this was just spitballing, maybe I'll work on it if I start actively using Julia again.

Yeah, I've thought of this too. I don't think it would be too terribly hard to implement, just haven't found the time yet. 

#394 is relevant

nverno:
Yea I think it wouldn't be too hard, and more efficient than adding all the checks to an input filter I think.

jabranham:
> Yea I think it wouldn't be too hard, and more efficient than adding all the checks to an input filter I think.

You can check out 'ess-smart-comma', which does something similar if point is at the R prompt.

vspinu:
> so the mode switching would be handled by the conditional bindings.

Such smart commands at REPL have very long history in emacs. SLIME, CIDER and probably others have it.

> This is basically what comint does by default;

It's not very good at it as it marks only those which you interactively enter. All other prompts which are printed during the normal output need to be tackled manually.

PR #915, opened 2019-05-10
[Fix #911] Don't overwrite local user indentation settings

### LINKED ISSUE #911: indentation of basic if-else broken
```R
x <-
  if (a) b
else c
```

I am very surprised this doesn't pop in our tests. Do you see this?

vspinu:
The idea is to explicitly track which user style settings have been set as buffer local variables.

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -311,17 +311,9 @@ indentation style. See `ess-style-alist' for predefined styles."
   ;; completion
   (add-hook 'comint-dynamic-complete-functions 'ess-complete-filename nil 'local)
   (delq t comint-dynamic-complete-functions)
-  (set (make-local-variable 'comint-completion-addsuffix)
-       (cons "/" ""))
-  (add-hook 'hack-local-variables-hook #'ess--set-style-in-buffer nil t)
+  (setq-local comint-completion-addsuffix (cons "/" ""))
+  (add-hook 'hack-local-variables-hook (lambda() (ess-set-style ess-style t t)) nil t)
   (add-hook 'ess-idle-timer-functions 'ess-synchronize-dirs nil 'local))
-
-(defun ess--set-style-in-buffer ()
-  "Set `ess-style' taking into account file and directory local variables.
-This is added to `hack-local-variables-hook'."
-  (setq-local ess-style (or (alist-get 'ess-style file-local-variables-alist)
-                            ess-style))
-  (ess-set-style ess-style t))
 
 (defun ess--get-mode-line-indicator ()
   "Get `ess--mode-line-process-indicator' from process buffer.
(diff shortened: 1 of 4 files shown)


lisp/ess-mode.el
```diff
@@ -311,18 +311,10 @@ indentation style. See `ess-style-alist' for predefined styles."
   ;; completion
   (add-hook 'comint-dynamic-complete-functions 'ess-complete-filename nil 'local)
   (delq t comint-dynamic-complete-functions)
-  (set (make-local-variable 'comint-completion-addsuffix)
-       (cons "/" ""))
-  (add-hook 'hack-local-variables-hook #'ess--set-style-in-buffer nil t)
+  (setq-local comint-completion-addsuffix (cons "/" ""))
+  (add-hook 'hack-local-variables-hook (lambda() (ess-set-style ess-style t t)) nil t)
```
jabranham: Can we keep the named function? Named functions are much easier to debug in hooks than lambdas.

lisp/ess-mode.el
```diff
@@ -429,32 +421,40 @@ current function."
...
-    (if (not quiet)
-        (message "ESS-style: %s" ess-style))
-    ;; finally, set the indentation style variables making each one local
-    (mapc (lambda (ess-style-pair)
-            (make-local-variable (car ess-style-pair))
-            (set (car ess-style-pair)
-                 (cdr ess-style-pair)))
-          (cdr (assq ess-style ess-style-alist)))
-    ess-style))
+  (setq style (or style
```
jabranham: style has to be set unless this is called interactively, right? Can't we move it to a non-optional argument and then put this stuff about ess-completing-read in the call to `(interactive)`?

lisp/ess-mode.el
```diff
@@ -429,32 +421,40 @@ current function."
...
+                  (let ((styles (mapcar (lambda (x) (symbol-name (car x)))
+                                        ess-style-alist)))
+                    (intern (ess-completing-read
+                             "Set ESS mode indentation style"
+                             styles nil t nil nil ess-style)))))
+  (let ((style-alist (or (cdr (assq style ess-style-alist))
+                         (format "Bad ESS style: %s" style)))
```
jabranham: Why isn't this an error anymore?

lisp/ess-mode.el
```diff
@@ -429,32 +421,40 @@ current function."
...
+                           unless (local-variable-p var)
+                           collect var))))
+    (unless quiet
+      (message "ESS-style: %s" style))
+    (mapc (lambda (var)
+            (make-local-variable var)
+            (set var (cdr (assq var style-alist))))
+          (if no-overwrite
+              vars
+            (mapcar 'car style-alist)))
```
jabranham: car should be sharp quoted, right?

lisp/ess-mode.el
```diff
@@ -429,32 +421,40 @@ current function."
...
+                         (format "Bad ESS style: %s" style)))
+        (vars (or (cdr ess--installed-style-vars)
+                  (cl-loop for (var . _) in (cdr (assq 'DEFAULT ess-style-alist))
+                           unless (local-variable-p var)
+                           collect var))))
+    (unless quiet
+      (message "ESS-style: %s" style))
+    (mapc (lambda (var)
+            (make-local-variable var)
+            (set var (cdr (assq var style-alist))))
```
jabranham: `setq-local`?

test/ess-test-indentation.el
```diff
@@ -86,7 +88,9 @@ where the edit took place. Return nil if E represents no real change.
   (with-current-buffer buffer
     (setq buffer-undo-list nil)
     (indent-region (point-min) (point-max))
-    (not (buffer-modified-p buffer))))
+    (if (buffer-modified-p buffer)
+        (progn (switch-to-buffer buffer) nil)
+      t)))
```
jabranham: Why this change? The previous way seems more straightforward.

lisp/ess-mode.el
vspinu: I know but it seems like a waste for a single call.

lisp/ess-mode.el
vspinu: Good idea.

lisp/ess-mode.el
vspinu: Because I deleted it by accident. Good catch!

lisp/ess-mode.el
vspinu: Right.

lisp/ess-mode.el
vspinu: Cannot be because `var` needs to be evaluated.

test/ess-test-indentation.el
vspinu: Just convenience for the interactive use. In case of error the buffer with the wrong indentation will be brought to the front.

lisp/ess-mode.el
vspinu: @jabranham I think I have just by mistake overwritten some of your commits from today in master. Could you please check and force push if so? Have no idea idea how that could have happened but that's is of less importance now.

vspinu: Ok. I think it's a bug in magit. I have just updated it today and now on a non-master branch force-push is pushing to the master. Be careful everyone.

jabranham: We could turn on github's protected branch stuff to disable force-pushing to master. I'd be in favor of that.

vspinu: Yes. Let's do that.

PR #920, opened 2019-05-11
[Fix #911] Don't overwrite local user indentation settings

### LINKED ISSUE #911: indentation of basic if-else broken

vspinu:
This is a followup on #915 which I inadvertently messed up. 

@jabranham I have found a nice compromise. The extra var and tracking of locals are no longer needed. Nor an extra function for the hook.

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -311,17 +311,9 @@ indentation style. See `ess-style-alist' for predefined styles."
   ;; completion
   (add-hook 'comint-dynamic-complete-functions 'ess-complete-filename nil 'local)
   (delq t comint-dynamic-complete-functions)
-  (set (make-local-variable 'comint-completion-addsuffix)
-       (cons "/" ""))
-  (add-hook 'hack-local-variables-hook #'ess--set-style-in-buffer nil t)
+  (setq-local comint-completion-addsuffix (cons "/" ""))
+  (add-hook 'hack-local-variables-hook 'ess-set-style nil t)
   (add-hook 'ess-idle-timer-functions 'ess-synchronize-dirs nil 'local))
-
-(defun ess--set-style-in-buffer ()
-  "Set `ess-style' taking into account file and directory local variables.
-This is added to `hack-local-variables-hook'."
-  (setq-local ess-style (or (alist-get 'ess-style file-local-variables-alist)
-                            ess-style))
-  (ess-set-style ess-style t))
 
 (defun ess--get-mode-line-indicator ()
   "Get `ess--mode-line-process-indicator' from process buffer.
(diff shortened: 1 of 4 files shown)


lisp/ess-mode.el
```diff
@@ -429,32 +421,39 @@ current function."
...
 

 ;;; Formatting / indentation
+(defvar-local ess--installed-style-vars nil
+  "A cons of the form (STYLE . VARS).
+VARS is a list of all style vars which were not set explicitly to
+buffer local values by the user in mode hooks.")
 
-(defun ess-set-style (&optional style quiet)
+(defun ess-set-style (&optional style _quiet)
```
jabranham: Is style mandatory now? It should go before `&optional`, right?

lisp/ess-mode.el
```diff
@@ -429,32 +421,39 @@ current function."
...
+         (style-alist (or (cdr (assq style ess-style-alist))
+                          (error (format "Bad ESS style: %s" style))))
+         (vars (if keep-local
+                   ;; Install, but Keep user's buffer-local settings.
+                   (cl-loop for (var . _) in (cdr (assq 'DEFAULT ess-style-alist))
+                            unless (local-variable-p var)
+                            collect var)
+                 (mapcar #'car style-alist))))
+    (when (called-interactively-p 'any)
+      (message "ESS-style: %s" style))
```
jabranham: This is nit-picky, but this message could probably be more informative. Something like "ESS style set to %s."

lisp/ess-mode.el
```diff
@@ -311,18 +311,10 @@ indentation style. See `ess-style-alist' for predefined styles."
   ;; completion
   (add-hook 'comint-dynamic-complete-functions 'ess-complete-filename nil 'local)
   (delq t comint-dynamic-complete-functions)
-  (set (make-local-variable 'comint-completion-addsuffix)
-       (cons "/" ""))
-  (add-hook 'hack-local-variables-hook #'ess--set-style-in-buffer nil t)
+  (setq-local comint-completion-addsuffix (cons "/" ""))
+  (add-hook 'hack-local-variables-hook 'ess-set-style nil t)
```
jabranham: ess-set-style should be sharp-quoted.

lisp/ess-mode.el
```diff
@@ -429,32 +421,39 @@ current function."
...
+  (interactive
+   (list (let ((styles (mapcar (lambda (x) (symbol-name (car x)))
+                               ess-style-alist)))
+           (intern (ess-completing-read
+                    "Set ESS mode indentation style"
+                    styles nil t nil nil ess-style)))))
+  (let* ((keep-local (null style))
+         (style (or style ess-style))
+         (style-alist (or (cdr (assq style ess-style-alist))
+                          (error (format "Bad ESS style: %s" style))))
```
jabranham: The string argument of error gets passed to `format-message`, so no need to call `format` here, I think.

lisp/ess-mode.el
vspinu: No, it's not. When it's nil, it means install `ess-style` but don't overwrite the user settings. This means we can used `ess-set-style` in hack-local-variables-hook.

lisp/ess-mode.el
vspinu: It doesn't matter in this case and I think both we and core emacs mostly uses simple quote. I find it cleaner this way. Probably just a habit though.

lisp/ess-mode.el
vspinu: Makes sense. Fixed.

lisp/ess-mode.el
vspinu: I am leaving this with simple quote to match the majority style. Judging from the info file for #' it's not useful for symbols in general. Remembering to add #' in elisp is just an annoyance not existing in other lisps.

PR #921, opened 2019-05-16
[WIP] Inferior process misbehavior with individualized prompt

dickmao:
Like a lot of people, I set my R prompt to something other than the
default "> ".

Commit 1c2a5e95 causes `test-org-ob-R-session-output-test` to hang (this doesn't happen on
travis because the prompt remains the default).

The bug also manifests by simply running `M-x R` (assuming your prompt is not
the default).  You will find the prompt echoes unexpectedly with every
space typed.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -66,6 +66,7 @@
 (declare-function ess-dump-object-into-edit-buffer "ess-mode" (object))
 
 (defvar add-log-current-defun-header-regexp)
+(defvar-local ess--make-local-vars-permanent nil)
 
 ;; The following declares can be removed once we drop Emacs 25
 (declare-function tramp-file-name-method "tramp")
@@ -176,8 +177,9 @@ This may be useful for debugging."
       (setq-local default-directory cur-dir)
       ;; TODO: Get rid of this, we should rely on modes to set the
       ;; variables they need.
-      (ess-setq-vars-local customize-alist)
-      (inferior-ess--set-major-mode ess-dialect)
+      (let ((ess--make-local-vars-permanent t))
+        (ess-setq-vars-local customize-alist)
+        (inferior-ess--set-major-mode ess-dialect))
       ;; Read the history file
       (when ess-history-file
         (setq comint-input-ring-file-name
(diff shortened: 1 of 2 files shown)


vspinu:
> Like a lot of people,
Not a lot. Very few till they realized that ESS relies on the shape of the prompt. 

> I set my R prompt to something other than the default "> ".

This will break a lot of things, so many that we cannot even tell which.

dickmao:
I've used ESS for years with "R >" as my prompt, and the `make test` hanging and prompt echoing only started happening after the aforementioned commit.

> This will break a lot of things, so many that we cannot even tell which.

I am skeptical, but will close this PR as the customize-alist code is fraught (presumably a historical accident which no one reading this would have written).

vspinu:
> R >

This one should actually have worked, at least that was always an intent. The trailing "> " is what matter. 

Not sure of what could have caused the change. The commit that you have referenced is largely stylistic. 

Let me check at  least. It  could be indicative of a bigger problem.

mmaechler:
I agree that it would make sense to keep `R> "` working as prompt.  That's been the default prompt e.g. for R code to be submitted to the JSS (Journal of Statistical Software) and similar situations.  Notably, Kurt Hornik has liked that forever (or at least "for many years" ?) and I think used also "privately".

jabranham:
Ideally ESS would support any sort of customized prompt, though that's a ways off.

Related to #831

vspinu:
I have set my prompt to "R>  " and things work as expected. The test suite is ok on all versions of emacs that I have tried. So I guess this issue is no longer there. In any case I will keep the prompt for interactive sessions and if I notice something amiss will be back.

dickmao:
Indeed!  A bisect shows 14f6618 fixed it!

vspinu:
Still, it's a bad user experience setting. For instance, the background pinging (for eldoc etc) normally remove an extraneous "> " prompt which it produces, but with "R> " an extra "R> " pops up. The bol navigation (C-a) does not understand those multiple "R> " at the prompt. I bet there are other issues. 

I am actually thinking of enforcing "> " prompt at startup. What do people think? @lionel- @jabranham @mmaechler

lionel-:
I think in the long term we need to find a solution that allows custom prompts, and in the short term I don't mind letting people customise their prompts. Prompt issues do not seem to pop up so often as to justify a restriction.

lionel-:
That sounds good.

PR #923, opened 2019-05-16
Rewrite ess-rutils

jabranham:
This is a complete re-write of the functionality in ess-rutils so that we can keep this file when we contribute ESS to GNU Emacs.

The commit messages should be clear, let me know if there's something I missed.


vspinu:
How much of these could be used on a routine basis by a generic user? Would it make sense to move those (all?) into the core?

jabranham:
> How much of these could be used on a routine basis by a generic user? 

Most/all could be

> Would it make sense to move those (all?) into the core?

Yes, I was going to propose this later, but I can move it all into `ess-r-mode.el` if everyone's fine with that. Assuming we want to do that, should we remove `ess-rutils.el` or keep it as an empty file (in `lisp/obsolete`?) in case people `(require 'ess-rutils)` in their config?

vspinu:
> or keep it as an empty file (in lisp/obsolete?) in

Yes. In the lisp for now as MELPA still doesn't fetch obsolete dir. Will try to PR a fix to them this week.

jabranham:
OK, I've done that. Will merge soonish.

PR #925, opened 2019-05-20
Skip tests that rely on R when R isn't available

jabranham:


vspinu:
Someone running tests should have an R installed. This change puts an extra burden for those who write R tests in the future. I will surely forget about it at some point.

jabranham:
This is part of an effort to extend our testing facilities to other languages (e.g. Julia, see #901). If we're testing multiple languages, it makes sense not to install everything everywhere.

vspinu:
I  am not sure this is the right strategy. Someone modifying ESS for language X should make sure that the changes don't break language Y. Requiring all developers to have all languages is not such a big deal, right?

jabranham:
I don't think it's unreasonable to skip tests we know will fail if R isn't available. Our CI service will hopefully have all the languages (or at least the free languages) installed, so the dev working on language X will get notified if they break something in language Y. No need for everyone to install R, python, julia, SAS, stata, etc.

vspinu:
People developing for julia might break R functionality and that's big enough of an issue to enforce all devs having the open source languages (R, julia, python). It's trivial to install those and it's non trivial to hunt every single test and insert the protector. An extra complication with considerable negative consequences and questionable marginal benefits. Relying on travis for a core developer is just not right. 

Another story for occasional contributors, but those could rely on Travis, which I guess they do anyhow. 

More importantly, I am not convinced that it's better not to to see tests for, say julia, when julia is not installed than to see them explicitly falling. Falling tests will raise markers and will force people either install or rely on travis, and that's good.

jabranham:
> It's trivial to install those

Not if you don't have an internet connection at the time you're coding. Which is pretty common.

> and it's non trivial to hunt every single test and insert the protector.

This commit took me about a minute to put together. It's pretty trivial.

> I am not convinced that it's better not to to see tests for, say julia, when julia is not installed than to see them explicitly falling.

To me, a failing test should mean "you've broken something". If you're just missing some dependency we should skip the test.

nverno:
How about just adding ert tags to tests for a given language like julia that might not be installed?  Then, you could selectively run tests for a language that might not be installed locally, but still require all tests passing on travis.

vspinu:
@jabranham 
> Not if you don't have an internet connection at the time you're coding.

You need it once, just like with ESS git repo itself. In order to push the changes you need the internet connection anyhow, so you would run one final `make test` before the push and install all the dependencies. 

> This commit took me about a minute to put together. 

It's not about how much you put into it, but rather about maintaining the hierarchy in what follows. Every test writer in the future will have to keep in mind this peculiarity of our test suit. Some might not even be aware of it, and all test will run fine for them even if they broke something. 

> To me, a failing test should mean "you've broken something".

Exactly. And when you "skip the test" you don't know if you broke something. 

@nverno 
> Then, you could selectively run tests for a language

We actually have such mechanism already. One can run `make r` from `test` folder and run tests selectively. We could have extra targets `test-r`, `test-julia` etc. from the root.

jabranham:
>> Not if you don't have an internet connection at the time you're coding.

> You need it once, just like with ESS git repo itself. In order to push the changes you need the internet connection anyhow, so you would run one final make test before the push and install all the dependencies.

If I have ESS and I'm in a plane without internet, I'd like to be able to run the test suite regardless of what software is installed on my computer. Right now that's not possible. One of the reasons is that we assume some software is available (e.g. R), which might not be. Another is that `make clean test` removes julia-mode, which then makes it impossible to compile ESS.

>  Every test writer in the future will have to keep in mind this peculiarity of our test suit. 

This wouldn't be a peculiarity of our test suite. The ERT manual says this is exactly what skip-unless is for:

> Sometimes, it doesn’t make sense to run a test due to missing
> preconditions.  A required Emacs feature might not be compiled in, the
> function to be tested could call an external binary which might not be
> available on the test machine, you name it.  In this case, the macro
> ‘skip-unless’ could be used to skip the test:


>> To me, a failing test should mean "you've broken something".

> Exactly. And when you "skip the test" you don't know if you broke something.

Yes, that's my point.

vspinu:
> If I have ESS and I'm in a plane without internet, I'd like to be able to run the test suite regardless of what software is installed on my computer. 

Then install it before going on the plain. No need to bother everyone else with this extremely unlikely scenario and by all means a tiny inconvenience. 

R is a hard pre-requisite for ESS development. We can discuss for other languages, but R everyone must have. 

> Another is that make clean test removes julia-mode, which then makes it impossible to compile ESS.

That's unrelated issue and will be fixed with git submodules at some point. 

> > Exactly. And when you "skip the test" you don't know if you broke something.

> Yes, that's my point.

Ok, then we agree.  "skipping" is a bad idea. 

More generally, before making a potentially controversial PR like this one, it's a good idea to open an issue and discuss it first.

vspinu:
>  ERT manual says this is exactly what skip-unless is for:

If a hammer is there, it doesn't mean you have to use it. I understand that skipping a test or two could be reasonable and useful sometimes, but skipping tens and potentially hundreds of tests of what consitutes the core functionality is a very different story.

lionel-:
In the current state of things, I tend to agree with Vitalie. ESS is mostly tested via the R backend, so developing without it seems dangerous. On the other hand I don't think it really matters much as long as Travis checks pull requests with R at all times.

jabranham:
I think this is the wrong approach. ESS is Emacs Speaks Statistics, not Emacs Speaks R. python or julia people shouldn't have to install R just to run tests. And as long as we have a CI service, broken R tests will be caught easily.

But I'm clearly in the minority here, so I'll close this PR.

nverno:
But why not just have as Vitalie suggested earlier targets for languages you want to test locally, eg. a test-python target?  It seems like it would solve your problem.

You could simply implement a target like

```
test-%:
        ./run-tests $*
```

lionel-:
> But I'm clearly in the minority here, so I'll close this PR.

I don't think this is true. FWIW, I think this is a step in the right long term direction. I agree with Vitalie that developing ESS currently requires R, but as I said I don't think it's a big deal to allow skipping R unit tests since pull requests are fully checked. I also disagree with Vitalie's assessment that this PR was controversial a priori. Finally, if this has an influence on development of Python support, this should be taken into account.

jabranham:
@nverno That's an option, but at that point there's no advantage to including it in ESS than just distributing it as a separate package. There's been rumblings for some time to split up the languages ESS supports into separate packages anyway, so perhaps python is a good place to start.

nverno:
I guess the main disadvantage would just be discoverability of the python extension.  It would just be a bummer if something trivial stifled your efforts to extend language support.

vspinu:
> ESS is Emacs Speaks Statistics, not Emacs Speaks R.

Let's not confuse the two. ESS is for users. Tests are for developers.

htc502:
interesting discussion. as an end user, my question is that have you guys come up with a solution that brings python to ESS?  I am eager to try that. R is definitely my favorite. but python is doing more and more in some machine learning senario.

PR #930, opened 2019-05-29
Avoid downloading an ESSR binary on remote computers

### LINKED ISSUE #429: Potential security issue in ESSR

jabranham:
I've done some testing with this by using M-x shell to ssh to a remote computer on the local network. This seems fast enough and seems to work reliably. It would be helpful if others could test it though.

Closes #429

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -1328,10 +1328,10 @@ selected (see `ess-r-set-evaluation-env')."
   "Load/INSTALL/Update ESSR."
   (let* ((pkg-dir (expand-file-name "ESSR" ess-etc-directory))
          (src-dir (expand-file-name "R" pkg-dir)))
-    (if (not (or (bound-and-true-p ess-remote)
-                 (file-remote-p (ess-get-process-variable 'default-directory))))
-        (inferior-ess--r-load-ESSR--local src-dir)
-      (inferior-ess-r-load-ESSR--remote src-dir))))
+    (if (or (bound-and-true-p ess-remote)
+            (file-remote-p (ess-get-process-variable 'default-directory)))
+        (inferior-ess-r-load-ESSR--remote src-dir)
+      (inferior-ess--r-load-ESSR--local src-dir))))
 
 (defun inferior-ess--r-load-ESSR--local (src-dir)
   (let ((cmd (format "local({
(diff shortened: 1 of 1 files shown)


vspinu:
You basically deleted some code and left the plain file loader which was used now only as a fallback. There is a reason things are done the way are done. Testing on local networks is non-representative. Try with a slow remote server on the other side of the planet and then you will discover the reasons this file transfer doesn't work. We discussed this before. Sending big files through stdin doesn't work well. I believe you have tried it yourself even. 

If you want you can add a temporary variable to control this for now, but this solution is a no-go.

jabranham:
> Testing on local networks is non-representative

How do we know that? I figure that's the way ess-remote is most often used.

Regardless, I think this is still an improvement over the current situation where we're just serving a binary blob without any sort of authentication.

vspinu:
> I figure that's the way ess-remote is most often used.

No idea frankly. Maybe. I never used it on local networks.

> I think this is still an improvement

That blob is part of ESS repo, so it's as secure as doing git clone and using ESS itself. And more secure than using ESS from elpa. So not really an improvement AFAIC.

I will try to find time to fix this reliably in June. It's not hard as we already discussed.

jabranham:
> That blob is part of ESS repo, so it's as secure as doing git clone and using ESS itself

Only if you trust the service hosting ESS (github/microsoft at the moment).

Would you be OK with adding a boolean but flipping the current behavior (e.g. defaulting to `ess--inject-code-from-file` but with the option to fallback to the download)?

vspinu:
> Would you be OK with adding a boolean but flipping the current behavior 

This has been there for many years (in fact a much more insecure version). So I would prefer to leave it for a few more weeks with a boolean (if you really need it now). 

I will make it a high priority. Will try to fix by the end of the week.

vspinu:
@jabranham I just pushed a change to master inadvertedly. Instead of creating and pushing to new branch on origin magit now pushes to master. Will have to dig what's going on. 

Could you please review the history? 

The problem with `ess-remote` still percists with new emacses. I have implemented a chunked injection which does solve the problem on my machines. The fetching from github is still available but it's disabled by default. 

Loading the binary through stdin is tricky because it requires converting it to ascii encoding but core R doesn't have support for base64 encoding which is commonly used for this purpose.

vspinu:
> Instead of creating and pushing to new branch on origin magit now pushes to master. Will have to dig what's going on.

It was a [bug](https://github.com/magit/magit/issues/3872) in magit. It's fixed now.

PR #931, opened 2019-05-30
fix: convert to standard filename when dump object

ShuguangSun:
When we want to looked an object in a package not loaded (package::function) or an object not exported (package:::function), it will be not approprate to use the oject (package::function or package:::function) as filename. So convert it to strandard filename first.

### DIFF
--- a/lisp/ess-mode.el
+++ b/lisp/ess-mode.el
@@ -569,7 +569,7 @@ generate the source buffer."
                                                            ess-local-process-name))
                        (ess-setq-vars-local ess-local-customize-alist)
                        (apply ess-source-directory nil)))))
-         (filename (concat dirname (format ess-dump-filename-template object)))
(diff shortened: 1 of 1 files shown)


vspinu:
Good idea indeed. Thanks!

jabranham:
@ShuguangSun We're trying to contribute ESS to GNU Emacs upstream. They require contributors to sign paperwork that transfers their copyright to the FSF. You're approaching the point where we'd need the paperwork for you, so if you'd like to contribute to ESS in the future, you may want to start the paperwork now. There's more instructions here: https://github.com/emacs-ess/ESS/blob/master/doc/CONTRIBUTING.org and feel free to ask me questions.

ShuguangSun:
@jabranham Thanks for reminding me. I'll do it as soon as possible.

PR #935, opened 2019-06-05
Update reference card

frederic-santos:
Hi,
Since the reference card [seems to be outdated](https://ess.r-project.org/index.php?Section=documentation&subSection=reference%20cards), I propose some updates and fixes in the source tex file.
(Maybe I'm not the best candidate for that since I started using Emacs and ESS only one month ago, so there might remain some incorrect or obsolete keybindings.)
Also, I don't know if you wish to keep the current template for the reference card, or switch to the template adopted by [Emacs](https://www.gnu.org/software/emacs/refcards/pdf/refcard.pdf) and [OrgMode](https://www.gnu.org/software/emacs/refcards/pdf/orgcard.pdf).
Thanks!

### DIFF
--- a/doc/refcard/refcard.tex
+++ b/doc/refcard/refcard.tex
@@ -3,6 +3,7 @@
 \usepackage{parskip}
 \usepackage{verbatim}
 \usepackage{fullpage}% +- ok for paper 'A4' as well
+\usepackage{textcomp}% added from \textasciigrave
 %\usepackage{svn}% and {fullpage} are both in debian/ubuntu pkg 'texlive-latex-extra'
 
 \addtolength{\textheight}{20mm}
@@ -33,8 +34,8 @@
 
   \smallskip
 
-  {\small updated for ESS 12.09-1}% {\footnotesize --- needs \em{more} updating!}}
-  \\[1ex] {\tiny September 2012}
+  {\small updated for ESS 18.10-3}% {\footnotesize --- needs \em{more} updating!}}
+  \\[1ex] {\tiny June 2019}
            % \footnotesize --- as of \today
 \end{center}
 % takes a lot of space
@@ -54,13 +55,13 @@
 \Sect{Interacting with the process}
 %%    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-For use in a process buffers ({\small inferior-ess-mode}):
+For use in a process buffer ({\texttt{inferior-ess-mode}}):
 
 \begin{tabI}
   \texttt{\RET   } & Send a command\\
   \texttt{C-c \RET} & Copy old input\\
-  \texttt{\TAB}& Complete object or file name.\\
-  & Also bound to \texttt{M-\TAB}, \texttt{M-C-i}.\\
+  \texttt{\TAB}& Complete object or file name\\
+  %& Also bound to \texttt{M-\TAB}, \texttt{M-C-i}.\\ % seems to be obsolete
   \texttt{C-c C-c }& Break \\
   \texttt{C-g}     & interrupt Emacs' waiting for S\\
   %% SfS-only (others need C-u C-a for comint-bol !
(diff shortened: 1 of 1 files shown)


doc/refcard/refcard.tex
```diff
@@ -237,44 +240,47 @@
...
   (active during debugging):
 
   \begin{tabular}{p{20mm}l}
     \texttt{M-C}&  Continue \\
     \texttt{M-N}&  Next line\\
     \texttt{M-Q}&  Quit\\
     \texttt{M-U}&  Up frame\\
-    \texttt{C-M-S-c }&  Continue Multiple\\
-    \texttt{C-M-S-n }&  Next Multiple\\
+    \texttt{M-C-C}&  Continue Multiple\\
```
vspinu: I think the previous version is more explicit, and thus better. The M-C-C is a bit confusing.

doc/refcard/refcard.tex
```diff
@@ -288,18 +294,57 @@
...
+  
+  Keybindings in \texttt{ess-r-package-mode} (\texttt{C-c C-w})\\
+  (call to \texttt{devtools} functions):
+  
   \begin{tabI}
-    \texttt{C-t, t} & Toggle developer on/off\\
-    \texttt{C-a, a} & Add package to the dev list\\
-    \texttt{C-r, r} & Remove package from dev list\\
+    \texttt{C-b, b} & Build package (\texttt{devtools::build()})\\
+    \texttt{C-c C-c} & Check package (\texttt{devtools::check()})\\
```
vspinu: There are multiple commands on C-c. Would be good to list all.

doc/refcard/refcard.tex
```diff
@@ -288,18 +294,57 @@
...
+    \texttt{C-c C-e} & Insert a template of help file\\
+    \texttt{\TAB} & Indent current line\\
+    \texttt{C-c C-f} & Display list of font specifiers\\
+    \texttt{C-c C-j} & Insert an `\texttt{\textbackslash item}' on next line\\
+    \texttt{C-c C-n} & (in example section) \\
+                     & send current line to ESS process \\
+    \texttt{C-c C-p} & Preview help file in plain text \\
+  \end{tabI}
+
+  For \texttt{Roxygen} users, inside \texttt{*.R} files (\textbf{C-c C-o}):
```
vspinu: shall we make it `(prefix C-c C-o)`. Also elsewhere?

frederic-santos: Thanks, done in this section and elsewhere.

doc/refcard/refcard.tex
frederic-santos: The previous version has been restored.

doc/refcard/refcard.tex
frederic-santos: Indeed, I was wondering which ones should be included. The help in ESS lists three commands (and five keybindings):
```
C-c C-w C-c C-c			ess-r-devtools-check-package
C-c C-w C-c C-w			ess-r-devtools-check-with-winbuilder
C-c C-w C-c c			ess-r-devtools-check-package
C-c C-w C-c h			ess-r-rhub-check-package
C-c C-w C-c w			ess-r-devtools-check-with-winbuilder
```
In fact, `devtools::build_win()` seems to be deprecated (and will be removed soon), so I didn't know if `C-c C-w` should be included here.
For `C-c h`, no problem.

vspinu: Hmm, I found it pretty useful. Any idea why they removed it? Is there a replacement?

frederic-santos: It seems that the old `build_win()` has been replaced by three distinct functions, `check_win_release()`, `check_win_devel()` and `check_win_oldrelease()`: https://github.com/r-lib/devtools/issues/1598
I suppose it might require some changes in ESS.

vspinu: Yes. Your changes are good, but we probably will wait a bit before moving to those functions in order to keep it working on older versions for a bit longer.

vspinu:
Thanks!

PR #944, opened 2019-06-27
Remove old news from NEWS

jabranham:
Any reason to keep it in there?


lionel-:
For ESS I don't know, but for R it's useful to have the old NEWS easily available.

mmaechler:
Good point, @lionel-  We should some also provide a (or several) *rendered* versions of `onewfeat.texi`

vspinu:
I think we need to keep at least 2 years of development in that file, better 5. When people stumble on new behavior they search the news file. Digging the onews.texi, is not easy. Do we even compile it during make?) 

I would also rename the file into `old-news` on this ocassion.

jabranham:
OK, I have reworked this branch so that old-news is put into the info manual. There's a link right at the top of (new) news so that it's easy to find.

I think this makes it easy enough to find the old news that we don't need to keep it around in news, but if people still want to keep the last two or so releases in there let me know and I'll move them back.

vspinu:
I have looked at this again and, to be frank, I like the cleanup, but not so much the new organization. 

Keeping news.texi just for development news is non-standard. I don't know anyone doing that. 

How about keeping in news.texi all news from 2011 (from when we switched to new versioning scheme) and in the old news keep the old stuff up to version 5.x?

mmaechler:
I strongly prefer @vspinu 's proposal  (not the least because I found the scheme we use with R very sensible. Note that there, the old news are even split a bit further; but one reason for that is also that we had changed the source format (from an semi-formal ascii text to *.Rd).

jabranham:
> Keeping news.texi just for development news is non-standard. I don't know anyone doing that.

Well, Emacs, for one.

Seems like I'm in the minority here in wanting to clean this up though so I'll just close this PR.

PR #953, opened 2019-07-28
Obsolete ess-mouse and mouseme

jabranham:
None of this is tested and as far as I can see no one's really using it. You can get a nice menu on clicking by holding control and right clicking. Should we just move it to obsolete?

----

#


vspinu:
I would say we do.

PR #960, opened 2019-09-19
add latex expansion to ess-julia-mode

seregaxvm:
Add expansion of Unicode characters in `ess-julia-mode`.
For example, after typing `\alpha` and running command `completion-at-point`, `\alpha` will be replaced by Unicode character α.
This solves problem described at https://github.com/emacs-ess/ESS/pull/122#issuecomment-168600364


jabranham:
Thanks!

jabranham:
Merged as 9e522405814b3ae7a81853930d7c97d50cdadd4f

Issue #965, opened 2019-10-08
Review debugger implementation, commands and docs


vspinu:
Followup from #964

PR #966, opened 2019-10-09
Use cursor-intangible text properties instead of intangible props

jabranham:
The intangible text property was made obsolete in Emacs 25.1.

Only lightly tested this so would be good if others could check it out.


vspinu:
Let's merge this. This intangibility stuff isn't working quite as expected under some circumstances. Should be tested and reviewed more carefully as part of #965.

PR #969, opened 2019-10-11
Fix help lookup default choice

ilohmar:
When `helpobsj` starts w/ one or more `nil` values, delq removes them from the return value, but does not mutate the list.  Then the CAR will be nil and the default choice missing.

Easy fix is to store `delq`'s return value (the list itself is used nowhere else in the function).


vspinu:
Makes sense. Thanks!

PR #975, opened 2019-12-31
Fix `M-j' on dev emacs

lionel-:
Now bound to `default-indent-new-line'.

See https://github.com/emacs-mirror/emacs/commit/8a11e430ec261c08cc928a7a5b05ee1027f50368 for context. (Thanks to @jabranham for the pointer!)


jabranham:
Thanks! I think the argument should be optional though right?

lionel-:
It's compulsory in the implementations I've seen but could be made optional.

jabranham:
Ah, OK. I was basing my comment on the docstring of `comment-line-break-function` which says: `The function should take a single optional argument, which is a flag indicating whether it should use soft newlines.`

But it probably doesn't matter.

lionel-:
Merged with optional soft argument.

PR #977, opened 2020-01-15
Accept negative numbers in ess-auto-width

### LINKED ISSUE #976: `ess-execute-screen-options` set alternate width
I love how ESS can auto-adjust the R process's width to match the emacs window width (with ess-execute-screen-options), but ... I'd like to set the process to that width **minus 2 characters** (or 3 characters, or 1 character, etc.).

Is this doable with the current ESS implementation?

([stackoverflow post with the same question.](https://stackoverflow.com/questions/59727367/emacs-ess-ess-execute-screen-options-to-width-2?noredirect=1#comment105604922_59727367))

jabranham:
Closes #976

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -191,7 +191,9 @@ lot."
   "When non-nil, set the width option when the window configuration changes.
 When 'frame, set the width to the frame width. When 'window, set
 the width to the window width. If an integer, set the width to
-that integer. Anything else is treated as 'window."
+that integer. If it's a negative integer, set the width to the
+window's width minus that number. Anything else is treated as
+'window."
   :group 'ess
   :type '(choice (const :tag "Do nothing" :value nil)
                  (const :tag "Frame width" :value frame)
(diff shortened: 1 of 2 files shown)


lisp/ess-custom.el
```diff
@@ -191,7 +191,9 @@ lot."
   "When non-nil, set the width option when the window configuration changes.
 When 'frame, set the width to the frame width. When 'window, set
 the width to the window width. If an integer, set the width to
-that integer. Anything else is treated as 'window."
+that integer. If it's a negative integer, set the width to the
```
vspinu: Looks like in the code it's the positive integer and which is negated at the end.

jabranham: I don't think so?

```
(ess-calculate-width 'window)
"options(width=158, length=99999)
"

(ess-calculate-width -2)
"options(width=156, length=99999)
"

(ess-calculate-width 2)
"options(width=2, length=99999)
"
```

PR #981, opened 2020-01-27
Add compile function.

charJe:


lionel-:
I think we should avoid adding small functions piecemeal, and instead implement compilation commands in a consistent way for a range of cases. Connected to #531.

I've been using this for a while.

```lisp
(defun my-ess-r-eval-in-compile-buffer (expr)
  "Evalate R expression EXPR in a standalone `compilation-mode' buffer."
  (let* ((procname inferior-ess-r-program-name)
         (command (format "%s --slave --no-readline -e \"%s\"" procname expr)))
    (compilation-start command nil
                       (lambda (name-of-mode)
                         (concat "*" (downcase name-of-mode) "*"))
                       ess-r-error-regexp-alist)))

(defun my-ess-rmarkdown-render (arg)
  (interactive "P")
  (let ((cmd (if (file-exists-p "_bookdown.yml")
                 (if arg
                     (format "bookdown::render_book('%s')" buffer-file-name)
                   (format "bookdown::render_book('%s', preview = TRUE)" buffer-file-name))
               (format "rmarkdown::render('%s')" buffer-file-name))))
    (my-ess-r-eval-in-compile-buffer cmd)))
```

charJe:
That's fine with me. I just needed the functionality, and I thought other people might need it too. It should be a part of the master branch.

lionel-:
I'm closing this for now because this single command would be better suited in the Emacs wiki or in an SO answer. If we're adding compilation-mode support to ESS, it should be part of a larger design.

PR #985, opened 2020-02-08
eldoc respect eldoc-echo-area-use-multiline-p flag

seregaxvm:
Do not truncate eldoc message lines when `eldoc-echo-area-use-multiline-p` flag is
set to `t`

### DIFF
--- a/lisp/ess-julia.el
+++ b/lisp/ess-julia.el
@@ -278,10 +278,12 @@ to look up any doc strings."
             (setq args (sort args (lambda (s1 s2)
                                     (< (length s1) (length s2)))))
             (setq doc (concat doc (pop args)))
-            (while (and args (< (+ (length doc) (length (car args))) W))
+            (while (and args (or (eq t eldoc-echo-area-use-multiline-p)
+                              (< (+ (length doc) (length (car args))) W)))
               (setq doc (concat doc "  "
                                 (pop args))))
-            (when (and args (< (length doc) W))
+            (when (and args (or (eq t eldoc-echo-area-use-multiline-p)
+                                (< (length doc) W)))
               (setq doc (concat doc " {--}")))
             doc))))))
 
--- a/lisp/ess-r-completion.el
+++ b/lisp/ess-r-completion.el
@@ -71,16 +71,19 @@ to look up any doc strings."
                doc1)
           (when doc
             (setq doc (ess-eldoc-docstring-format funname doc))
-            (when (and margs (< (length doc1) W))
+            (when (and margs (or (eq t eldoc-echo-area-use-multiline-p)
+                                 (< (length doc1) W)))
               (setq doc1 (concat doc (propertize "  || " 'face font-lock-function-name-face)))
-              (while (and margs (< (length doc1) W))
+              (while (and margs (or (eq t eldoc-echo-area-use-multiline-p)
+                                    (< (length doc1) W)))
                 (let ((head (pop margs)))
                   (unless (assoc head bargs)
                     (setq doc doc1
                           doc1 (concat doc1 head  "=, ")))))
               (when (equal (substring doc -2) ", ")
                 (setq doc (substring doc 0 -2)))
-              (when (and margs (< (length doc) W))
+              (when (and margs (or (eq t eldoc-echo-area-use-multiline-p)
+                                   (< (length doc) W)))
                 (setq doc (concat doc " {--}"))))
             doc))))))
 


jabranham:
ESS's eldoc already respects the value of `eldoc-echo-area-use-multiline-p` for me. Can you give an example of how to trigger behavior where it doesn't?

seregaxvm:
I use the following function to show verbose eldoc message in `*Help*` buffer:
```elisp
(lambda () (interactive)
  (when eldoc-documentation-function
    (let (msg oldmult)
      (setq oldmult eldoc-echo-area-use-multiline-p)
      (setq eldoc-echo-area-use-multiline-p t)
      (setq msg (funcall eldoc-documentation-function))
      (setq eldoc-echo-area-use-multiline-p oldmult)
      (if msg
	  (with-help-window (help-buffer)
	    (princ msg))
	(message "Documentation not found")))))
```

In `ess-r-mode` without this patch I get:
r-wo
with patch:
r-w

In `ess-julia-mode`  without this patch I get:
julia-wo
with patch:
julia-w

jabranham:
That's not really how `let` should be used. This would be more correct, I think:

```elisp
(lambda () (interactive)
  (when eldoc-documentation-function
    (let ((eldoc-echo-area-use-multiline-p t)
          (msg))
      (setq msg (funcall eldoc-documentation-function))
      (if msg
	  (with-help-window (help-buffer)
	    (princ msg))
	(message "Documentation not found")))))
```

And seems to work for me. Let me know if that is what you want.

seregaxvm:
Thanks for the function fix, but it is not the issue. Without the submitted patch ess truncates named arguments even if `eldoc-echo-area-use-multiline-p` is set to `t`.
Look at `par()` example above. Without patch ess-eldoc returns only three named arguments, when in reality there are a few dozens of them.

jabranham:
Does it work the way you want if you use a function other than `par`? We treat par specially, try something else like `library` or `qplot`?

seregaxvm:
This will affect all functions that return list of arguments of all methods from [ess-function-arguments](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-inf.el#L2441-L2451).
In particular, functions `par`, `options` and S3 methods (`plot`, `summary`, `predict` etc)

vspinu:
I see all arguments of `par` as long as they fit into one line. 

@seregaxvm I am not entirely sure what original problem is. Have you looked at `ess-eldoc-abbreviation-style`?

seregaxvm:
From `eldoc-echo-area-use-multiline-p` description:
> If value is t, never attempt to truncate messages

So, it is expected that [`ess-r-eldoc-function`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-r-completion.el#L52-L85) and [`ess-julia-eldoc-function`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-julia.el#L263-L286) will not truncate messages when this flag is `t`.
However, ess treats some function arguments specially. While most of the arguments are processed (in case of R) with [`ess-eldoc-docstring-format`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-r-completion.el#L87-L145) ([here](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-r-completion.el#L73)), `par`, `options` and S3 methods are [processed separately](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-r-completion.el#L74-L84) and truncated regardless of  `eldoc-echo-area-use-multiline-p` value.

> I am not entirely sure what the original problem is. Have you looked at ess-eldoc-abbreviation-style?

`ess-eldoc-abbreviation-style` is [applied in `ess-eldoc-docstring-format`](https://github.com/emacs-ess/ESS/blob/master/lisp/ess-r-completion.el#L96-L141), so arguments, that are not being processed by this function, are not formatted properly.

vspinu:
Thanks for clarification, you are right of course. I didn't like that when `ess-eldoc-abbreviation-style` is 'aggressive it overwrites the value of `eldoc-echo-area-use-multiline-p`, so I changed that as well.

lisp/ess-julia.el
```diff
@@ -278,10 +278,12 @@ to look up any doc strings."
...
                                     (< (length s1) (length s2)))))
             (setq doc (concat doc (pop args)))
-            (while (and args (< (+ (length doc) (length (car args))) W))
+            (while (and args (or (eq t eldoc-echo-area-use-multiline-p)
+                              (< (+ (length doc) (length (car args))) W)))
               (setq doc (concat doc "  "
                                 (pop args))))
-            (when (and args (< (length doc) W))
+            (when (and args (or (eq t eldoc-echo-area-use-multiline-p)
+                                (< (length doc) W)))
```
vspinu: This change is redundant, args is nil here if `eldoc-echo-area-use-multiline-p` is t.

PR #989, opened 2020-02-26
Remove buffer switching from ess-request-a-process

### LINKED ISSUE #987: display-buffer-alist reusable-frames does not work as expected
This was originally posted on [emacs so](https://emacs.stackexchange.com/q/55556/21814).

When using the `display-buffer-alist` value from [ess doc](https://github.com/emacs-ess/ESS/blob/master/doc/ess.texi#L632-L642)
```lisp
(setq display-buffer-alist
      '(("*R"
         (display-buffer-reuse-window display-buffer-pop-up-frame)
         (reusable-frames . 0))))
```
and starting R shell from some `ess-r-mode` buffer the following occurs: after creating a new frame displaying the R shell, Emacs displays a second instance of the `ess-r-mode` buffer in this new frame instead of returning to the old instance. 
Everything works correctly when I set `reusable-frames` globally
```lisp
(setq display-buffer-reuse-frames 0)
```
but this should not be necessary.

Quoting [so answer](https://emacs.stackexchange.com/a/55571/21814):

The erroneous behaviour happens only the first time one types `C-c C-z` (`ess-switch-to-inferior-or-script-buffer`) in an `ess-r-mode` buffer, i.e. before the corresponding inferior `*R*` process buffer has been created.

This is because `ess-force-buffer-current` calls `ess-request-a-process` with a non-`nil` `noswitch` argument when an inferior process does not already exist.

`ess-request-a-process`, in turn, does the following as its last step:
```lisp
(if noswitch
    (pop-to-buffer (current-buffer)) ;; VS: this is weird, but is necessary
  (pop-to-buffer (buffer-name (process-buffer (get-process proc)))))
```

This call to `(pop-to-buffer (current-buffer))` is wrong and is what causes your `ess-r-mode` buffer to appear a second time in the new frame. It was added in the following commit from 2012: <https://github.com/emacs-ess/ESS/commit/b29ea8f934f7c08a512c73f14e914bca7229b3c1>

I boldly say it is wrong because popping to the current buffer is quite an intrusive operation (as indicated by the bug in question), and the original intention of the author can almost definitely be written in a better way. I don't know what issue the author originally faced, but perhaps the ESS devs can figure it out.

jabranham:
Use `display-buffer-overriding-function' and
`display-buffer-no-window' to avoid showing the inferior buffer until
the end of the function since we specifically handle whether we show
the buffer or now via a function argument.

Closes #987

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -718,23 +718,24 @@ LANGUAGE is ignored."
 (defun ess-request-a-process (message &optional noswitch ask-if-1)
   "Ask for a process, and make it the current ESS process.
 If there is exactly one process, only ask if ASK-IF-1 is non-nil.
-Also switches to the process buffer unless NOSWITCH is non-nil.  Interactively,
-NOSWITCH can be set by giving a prefix argument.
-Returns the name of the selected process."
-  (interactive
-   (list "Switch to which ESS process? " current-prefix-arg))
-                                        ; prefix sets 'noswitch
-  (ess-write-to-dribble-buffer "ess-request-a-process: {beginning}\n")
+Also switches to the process buffer unless NOSWITCH is non-nil.
+Interactively, NOSWITCH can be set by giving a prefix argument.
+Returns the name of the selected process. MESSAGE may get passed
+to `ess-completing-read'."
+  (interactive (list "Switch to which ESS process? " current-prefix-arg))
   (update-ess-process-name-list)
-
-  (setq ess-dialect (or ess-dialect
-                        (ess-completing-read
-                         "Set `ess-dialect'"
-                         (delete-dups (list "R" "S+" (or (bound-and-true-p S+-dialect-name) "S+")
-                                            "stata" (or (bound-and-true-p STA-dialect-name) "stata")
-                                            "julia" "SAS" "XLS"  "ViSta")))))
-
-  (let* ((pname-list (delq nil ;; keep only those matching dialect
+  (let* ((ess-dialect (or ess-dialect
+                          (ess-completing-read
+                           "Set `ess-dialect'"
+                           (delete-dups (list "R" "S+" (or (bound-and-true-p S+-dialect-name) "S+")
+                                              "stata" (or (bound-and-true-p STA-dialect-name) "stata")
+                                              "julia" "SAS")))))
+         ;; Set `display-buffer-overriding-action' here since we
+         ;; explicitly handle buffer switching below with
+         ;; `pop-to-buffer' which handles the `display-buffer'
+         ;; framework.
+         (display-buffer-overriding-action '(display-buffer-no-window (allow-no-window . t)))
+         (pname-list (delq nil ;; keep only those matching dialect
                            (append
                             (mapcar (lambda (lproc)
                                       (and (equal ess-dialect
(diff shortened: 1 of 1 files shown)


seregaxvm:
I used the following `init.el` file:
```lisp
(add-to-list 'load-path "/tmp/ess-site/share/emacs/site-lisp/ess/lisp")

(require 'ess)
(require 'ess-r-mode)

(setq display-buffer-alist
      '(("*R"
         (display-buffer-reuse-window display-buffer-pop-up-frame)
         (reusable-frames . 0))))
```
Works fine with and without `display-buffer-alist`.
One small annoyance - selection does not return to the original `ess-r-mode` buffer and stays in R repl.

jabranham:
Thanks for testing.

> selection does not return to the original ess-r-mode buffer and stays in R repl.

When you do what? `C-c C-z`?

seregaxvm:
> When you do what?

No, on repl startup from `ess-r-mode` buffer (<C-return>). Selection used to return to `ess-r-mode` original buffer from newly created one.

lisp/ess-inf.el
```diff
@@ -718,23 +718,24 @@ LANGUAGE is ignored."
...
+                          (ess-completing-read
+                           "Set `ess-dialect'"
+                           (delete-dups (list "R" "S+" (or (bound-and-true-p S+-dialect-name) "S+")
+                                              "stata" (or (bound-and-true-p STA-dialect-name) "stata")
+                                              "julia" "SAS")))))
+         ;; Set `display-buffer-overriding-action' here since we
+         ;; explicitly handle buffer switching below with
+         ;; `pop-to-buffer' which handles the `display-buffer'
+         ;; framework.
+         (display-buffer-overriding-action '(display-buffer-no-window (allow-no-window . t)))
```
vspinu: So this is the only change right? All others are cleanups of various kind AFAICS. 

If we move this line all the way down to where pop-to-buffer is used, the comment would be no longer necessary.

jabranham: Yes, though this fix isn't complete yet. It's more complicated than I thought and I haven't had time to look at it yet.

jabranham:
> Selection used to return to ess-r-mode original buffer from newly created one.

@seregaxvm I think this is controlled by `inhibit-switch-frame` in the display buffer alist. So hopefully:

```elisp
(setq display-buffer-alist
      '(("*R"
         (display-buffer-reuse-window display-buffer-pop-up-frame)
         (reusable-frames . 0)
         (inhibit-switch-frame . t))))
```

works for you. However, there's no guarantee whatever window manager you're using will honor that so it may not. 

This PR is ready to go but doesn't work with Emacs 25 so I'll merge whenever we drop support for Emacs 25 (should be happening around the time of Emacs 27 release, which is soon^TM)

vspinu:
@jabranham would be nice to keep 25compatibility till we release. We haven't released for so long that we need to keep 25 in next release. Would be really great find a 25 hack for this, or just keep old behavior on 25th.

jabranham:
@vspinu if/when there's a plan to release, I'll look into it

jabranham:
> What is the behaviour with Emacs 25?

Judging from the travis build (https://travis-ci.org/github/emacs-ess/ESS/jobs/655569672) there seems to be 3 issues, I'm unsure if they're related:

1. one test fails  because at some point a window is no longer live
2. another test fails because a process is no longer live
3. the test suite hangs

PR #993, opened 2020-03-06
fix(ess-r-mode): fix eldoc in Emacs 28 snapshot

ShuguangSun:
The eldoc-mode can't be turned on in the Emacs version:
       GNU Emacs 28.0.50 (build 2, x86_64-w64-mingw32) of 2020-02-27

The `eldoc-documentation-functions` is nill and not sure whether it is a bug of Emacs. The fix makes eldoc work with ESS again.


jabranham:
Thanks! I updated this for julia and the inferior modes as well.

vspinu:
I don't see eldoc on Emacs 28 though.

PR #994, opened 2020-03-07
[Fix #990] Use ESSDESTDIR instead of ESSDIR for the installation dir

### LINKED ISSUE #990: build ESSDIR configuration gets overwritten
Current Makefile from `master` branch does not respect `ESSDIR` value from Makeconf file.
It gets unconditionally overwritten [here](https://github.com/seregaxvm/ESS/blob/master/Makefile#L8).

vspinu:


jabranham:
LGTM

PR #1006, opened 2020-04-26
Update copyright and license

jabranham:
This is in preparation to donate ESS to the FSF so it can be hosted from GNU ELPA.

- Standardize all header information so that the FSF has copyright
- Move to GPLv3

At this point some work still remains to be done before going on GNU ELPA. There are a few files we've been unable to get copyright assignment for:

- ess-stata-lang.el
- ess-stata-mode.el
- obsolete/ess-noweb-font-lock-mode.el
- obsolete/ess-noweb-mode.el
- obsolete/ess-noweb.el
- obsolete/ess-r-args.el
- obsolete/ess-rutils.el
- obsolete/ess-swv.el

so we'll need to decide if we want to remove or rewrite those.


rsparapa:
We can live without stata-mode.  But, will we be OK without noweb?  Since polymode is not part of ESS, then what will happen without noweb?

lionel-:
noweb has fundamental and critical bugs that cause issues in other Emacs packages as soon as it's loaded. No matter the status of polymode, it should be deprecated and its use discouraged, because we are not being good citizens right now :(

vspinu:
So we have no choice but to move ess-stata into a separate package then? 

The `obsolete/*` should be deleted of course.

jabranham:
We can either move it to a separate package or rewrite stata support.

A rewrite probably wouldn't be that much work. I imagine font-lock is the most complicated bit, everything else should be pretty simple (famous last words...)

lionel-:
Is there any evidence that we have Stata users? hmm I guess there's https://github.com/emacs-ess/ESS/labels/lang%3Astata

I would just move to a separate package https://github.com/emacs-ess/stata-mode/ and mention in the README that we're looking for someone to rewrite it with FSF copyright. And perhaps a hint at @tslumley that he's still welcome to sign the FSF papers for his elisp code ;)

rsparapa:
Good suggestion.  The mode is actually helpful for UNIX/Linux stata.  But, stata on those platforms seems to be fairly unreliable which makes for a frustrating experience.

vspinu:
I used it a few times a few years ago. It wasn't terrible (in fact pretty helpful) but indeed linux support doesn't seem to be flourishing. 

I would vote for what @lionel- proposed. Create a new repo `ess-stata-mode` in our org and move all the stuff there. No need to push to melpa I guess, or what do you think?

jabranham:
Just finished up this branch, think it's ready to go. Once merged, we can put the next release of ESS on GNU ELPA.

Of course, you'll still have to have MELPA available for julia-mode. Anyone know if someone's contacted the authors to see if they can go through the copyright process?

jabranham:
the julia-mode package, not ess-julia.

rsparapa:
Ok, then I would say that is not part of ESS and I would get rid of it.  Maybe it could be part of ESS+.  It is time for this to happen instead of constantly going after signatures.  Keep in mind that they will also have to install polymode, markdown-mode, etc. so julia-mode is no different.

jabranham:
I got in touch with Stefan Monnier who suggested we remove julia-mode from Package-Requires and helped out with how to successfully compile ess-julia without julia-mode. Just pushed a commit that does that.

vspinu:
Guys, let's not delete stata and other major things before we release. Let's make next release the last release of ESS as we know it before GNU-ELPA thing.

rsparapa:
I don't think that is necessary.  Keep in mind these are practically unsupported and haven't changed much in years.  The current release that we have out is there is adequate if anyone wants these features.

jabranham:
The only thing being delete is Stata support. Julia-mode won't be in `Package-Requires` anymore so it won't get automatically pulled in. This is a little annoying for some users but enables installing with just GNU ELPA (i.e. no MELPA).

I don't want to wait any longer on this. We've already been more than 2 years without a release so some of the changes I'm most excited about haven't been released yet. I absolutely do not want to wait another 2 years to put this on GNU ELPA.

rsparapa:
+1

jabranham:
Cleaned up the commits on this branch, will merge later this week.

jabranham:
Done, I think next ESS release we'll be good-to-go to put on ELPA.

PR #1007, opened 2020-05-02
Update `default-directory` of process buffer after setting wd

### LINKED ISSUE #1004: `ess-set-working-directory` should not change `default-directory`
I've been confused for years with the behaviour of Emacs regarding the active directories of the files I'm editing. I just realised it's `ess-set-working-directory` that changes it to the new working directory of R.

It seems that this behaviour is intended: https://github.com/emacs-ess/ESS/commit/40ebf0f71eb0677895506201fac6da4d6119ffa2. But I'm not sure why changing the REPL directory should affect the buffer I'm switching from.

Should we change this to preserve the active directory? The active directory influences UI commands like `M-x find-file`. I suspect I'm not the only one who is confused by this behaviour.

lionel-:
Instead of updating the caller's buffer.
Closes #1004.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -2760,8 +2760,8 @@ NO-ERROR prevents errors when this has not been implemented for
                         (with-parsed-tramp-file-name path v v-localname)
                       path)))
         (ess-eval-linewise (format ess-setwd-command lpath))
-        ;; use set instead of setq to take effect even when let bound
-        (set 'default-directory (file-name-as-directory path)))
+        (ess-set-process-variable 'default-directory
+                                  (file-name-as-directory path)))
     (unless no-error
       (error "Not implemented for dialect %s" ess-dialect))))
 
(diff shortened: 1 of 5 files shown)


lisp/ess-inf.el
```diff
@@ -2760,8 +2760,8 @@ NO-ERROR prevents errors when this has not been implemented for
                         (with-parsed-tramp-file-name path v v-localname)
                       path)))
         (ess-eval-linewise (format ess-setwd-command lpath))
-        ;; use set instead of setq to take effect even when let bound
-        (set 'default-directory (file-name-as-directory path)))
+        (ess-set-process-variable 'default-directory
```
jabranham, line 2763: I don't think this call is needed, right?

lionel-: I'm not sure what you mean. This call sets the process default directory as we discussed.

jabranham:
I think it gets set without this though. Unless I'm missing something?

lionel-:
Where do you think it is getting set?

jabranham:
Black magic sorcery?

I think we've got a watcher in the inferior buffer to automatically sync
the working directory of the buffer and process

lionel-:
ok let's keep the non-magic line then ;)
Also in my tests the default-directory of the inferior was not updated without it.

PR #1014, opened 2020-05-20
Set default value of `ess-eval-visibly` to `'nowait`

### LINKED ISSUE #998: Change default of ess-eval-visibly?
What do people think about changing the default of ess-eval-visibly? I think the current default, which blocks Emacs, comes as a surprise to a lot of people, especially newcomers. IME it's one of the first things people change or they get frustrated by.

I'd probably chose `nil` over `nowait` since our output isn't perfect.

lionel-:
Closes #998.

This default is better for most people because the process reveals boundaries between rounds of evaluation, which makes it much easier to figure out what an output is about.

Once https://github.com/lionel-/ESS/tree/fix-prompt is merged, the behaviour will be identical to `nil` except for the echoing the input commands at the start of evaluation.

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -1823,9 +1823,9 @@ Otherwise, they get their own temporary buffer."
 
 (defcustom ess-eval-visibly 'nowait
   "Non-nil means ess-eval- commands display commands in the process buffer.
-If 'nowait (the default), ESS shows input commands in the process
-buffer, but doesn't wait for the process. Thus all the output is
-printed after the input lines.
+If 'nowait, ESS shows input commands in the process buffer, but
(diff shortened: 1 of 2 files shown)


lisp/ess-custom.el
```diff
@@ -1821,17 +1821,18 @@ Otherwise, they get their own temporary buffer."
 
 (defvaralias 'ess-eval-visibly-p 'ess-eval-visibly)
 
-(defcustom ess-eval-visibly nil
+(defcustom ess-eval-visibly 'nowait
   "Non-nil means ess-eval- commands display commands in the process buffer.
+If 'nowait (the default), ESS shows input commands in the process
```
jabranham, line 1826: Please don't say (the default), it's already put in there automatically by Emacs.

PR #1015, opened 2020-05-20
Improve xref lookup

vspinu:
Xref wasn't working for me in 80% of the cases. Things seem to be always working with this patch as long as `ess-r-package-library-paths` is correctly setup.

I wonder how we can popularize `ess-r-package-library-paths` a bit more. I expect few people know that in order to make xref work that one is necessary.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,6 +4,10 @@
 Changes and New Features in 19.04 (unreleased):
 @itemize @bullet
 
+@item ESS[R]: Improved @code{xref} lookup (@kbd{M-.}).
+Function locations are now always detected for package libraries listed
+in @code{ess-r-package-library-paths}.
+
 @item ESS[R]: Evaluated lines starting with the Roxygen prefix
 are now always stripped from the prefix, so they can be sent to the
 process easily. Previously, this was only the case inside the
--- a/etc/ESSR/R/completion.R
+++ b/etc/ESSR/R/completion.R
@@ -21,7 +21,7 @@ local({
 
 .ess_eval <- function(str, env = globalenv()) {
     ## don't remove; really need eval(parse(  here!!
-    tryCatch(base::eval(base::parse(text=str), envir = env),
+    tryCatch(eval(parse(text=str), envir = env),
              error=function(e) NULL) ## also works for special objects containing @:$ etc
 }
 
@@ -38,7 +38,7 @@ local({
     }
     fn <- .ess_eval(name, env)
     if (is.null(fn)) {
-        objs <- getAnywhere(name)$objs
+        objs <- utils::getAnywhere(name)$objs
         for (o in objs) {
             if (is.function(o)) {
                 fn <- o
@@ -55,13 +55,13 @@ local({
             out <- sprintf("(\"%s\" %d %d)\n", file, line, col - 1)
         }
     }
-    base::cat(out)
+    cat(out)
 }
 
 
 .ess_fn_pkg <- function(fn_name) {
     objs <- utils::getAnywhere(fn_name)
-    base::print(base::sub("(package|namespace):", "", objs$where))
+    print(sub("(package|namespace):", "", objs$where))
 }
 
 .ess_funargs <- function(funname) {
--- a/etc/ESSR/R/completion.R
+++ b/etc/ESSR/R/completion.R
@@ -37,6 +37,15 @@ local({
         env <- globalenv()
     }
     fn <- .ess_eval(name, env)
+    if (is.null(fn)) {
+        objs <- getAnywhere(name)$objs
+        for (o in objs) {
+            if (is.function(o)) {
+                fn <- o
+                break;
+            }
+        }
+    }
     out <- "()\n"
     if (is.function(fn) && !is.null(utils::getSrcref(fn))) {
         file <- utils::getSrcFilename(fn, full.names = TRUE)
@@ -46,21 +55,13 @@ local({
             out <- sprintf("(\"%s\" %d %d)\n", file, line, col - 1)
         }
     }
-    cat(out)
+    base::cat(out)
 }
 
 
 .ess_fn_pkg <- function(fn_name) {
-    fn <- .ess_eval(fn_name)
-    env_name <- base::environmentName(base::environment(fn))
-    out <- if (base::is.primitive(fn)) { # environment() does not work on primitives.
-               "base"
-           } else if (base::is.function(fn) && env_name != "R_GlobalEnv") {
-               env_name
-           } else {
-               ""
-           }
-    base::cat(base::sprintf("%s\n", out))
+    objs <- utils::getAnywhere(fn_name)
+    base::print(base::sub("(package|namespace):", "", objs$where))
 }
 
 .ess_funargs <- function(funname) {
--- a/lisp/ess-r-xref.el
+++ b/lisp/ess-r-xref.el
@@ -93,19 +93,19 @@ DEFAULT-PKG is the name of the package where presumably SYMBOL is located."
                          ((listp ess-r-package-library-paths)
                           ess-r-package-library-paths)
                          (t (user-error "Invalid value of `ess-r-package-library-paths'"))))
-         (pkg.dir (or (when (= (length pkgs) 1)
-                        (cons (car pkgs) (assoc-default (car pkgs) ess-r-xref-pkg-sources)))
-                      (cl-loop for pkg in pkgs
-                               for d in lib-dirs
-                               for p = (expand-file-name pkg d)
-                               when (file-exists-p p) return (cons pkg p))))
-         (file (when pkg.dir (expand-file-name src-file (cdr pkg.dir)))))
+         (loc (or (when (= (length pkgs) 1)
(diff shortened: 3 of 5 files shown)


doc/newfeat.texi
```diff
@@ -4,6 +4,8 @@
 Changes and New Features in 19.04 (unreleased):
 @itemize @bullet
 
+@item ESS[R]: Improved @code{xref} lookup (@kbd{M-.})
```
lionel-, line 7: This needs an explanation of what was improved.

> I wonder how we can popularize ess-r-package-library-paths a bit more. I expect few people know that in order to make xref work that one is necessary.

Maybe mention it here? I don't use this variable personally.

etc/ESSR/R/completion.R
```diff
@@ -37,6 +37,15 @@ local({
         env <- globalenv()
     }
     fn <- .ess_eval(name, env)
+    if (is.null(fn)) {
+        objs <- getAnywhere(name)$objs
```
lionel-, line 41: This seems a bit sloppy to me, in case of name conflict the behaviour will be unpredictable.

lionel-:
```suggestion
        objs <- utils::getAnywhere(name)$objs
```

lisp/ess-inf.el
```diff
@@ -165,15 +165,13 @@ This may be useful for debugging."
...
   (let* ((dialect (eval (cdr (assoc 'ess-dialect customize-alist))))
          (process-environment process-environment)
-         ;; Use dialect if not R, R program name otherwise
-         (temp-dialect (if ess-use-inferior-program-in-buffer-name ;VS[23-02-2013]: FIXME: this should not be here
-                           (if (string-equal dialect "R")
-                               (file-name-nondirectory inferior-ess-r-program)
-                             dialect)
-                         dialect))
-         (inf-buf (inferior-ess--get-proc-buffer-create temp-dialect))
+         (inf-buf (inferior-ess--get-proc-buffer-create
```
lionel-, line 168: Is this change related to xref?

lisp/ess-r-xref.el
```diff
@@ -79,27 +79,32 @@ srcrefs point to temporary locations."
...
-                         (expand-file-name pkg ess-r-package-library-paths))
+(defun ess-r-xref--pkg-srcfile (symbol src-file &optional default-pkg)
+  "Look in the source directory of the R package containing symbol SYMBOL for SRC-FILE.
+DEFAULT-PKG is the name of the package where presumably SYMBOL is located."
+  (let* ((pkgs (delq nil
+                     (delete-dups
+                      (or (cons default-pkg
+                                (ess-get-words-from-vector (format ".ess_fn_pkg(\"%s\")\n" symbol)))
+                          (user-error "Can't find package for symbol %s" symbol)))))
+         (libdirs (cond ((stringp ess-r-package-library-paths)
```
lionel-, line 90: Symbol separator:

```suggestion
         (lib-dirs (cond ((stringp ess-r-package-library-paths)
```

lisp/ess-r-xref.el
```diff
@@ -79,27 +79,32 @@ srcrefs point to temporary locations."
...
+                         (list ess-r-package-library-paths))
                         ((listp ess-r-package-library-paths)
-                         (cl-loop for d in ess-r-package-library-paths
-                                  for p = (expand-file-name pkg d)
-                                  when (file-exists-p p) return p))
-                        (t (user-error "Invalid value of `ess-r-package-library-paths'")))))
-         (file (when dir (expand-file-name r-src-file dir))))
+                         ess-r-package-library-paths)
+                        (t (user-error "Invalid value of `ess-r-package-library-paths'"))))
+         (pkg.dir (or (when (= (length pkgs) 1)
```
lionel-, line 95: Symbol separator:

```suggestion
         (pkg-dir (or (when (= (length pkgs) 1)
```

lionel-:
Actually is there any use case where we would like to getAnywhere a symbol which isn't in the package namespace? It seems reasonable to ask the developer to reinstall the package when a symbol is imported (this way we can find it in the `asNamespace()`), or to use `devtools::load_all()` otherwise.

vspinu:
I think we should not take strictly dev prospective. One might want to inspect a code from a package which is not developed, or if developed not loaded with load_all. Or just request to navigate from `print(foo)` code etc. So I would see `asNamespace` a major complication. As long as the namespace is loaded we should get it and if namespace is not loaded and we can infer the package, we load it. 

Next step would be to print the source of the code in temporary buffers for functions/objects which we don't know the location as we discussed on few occasions. This would be a close to 100% solution.

lisp/ess-r-xref.el
vspinu: This is intentional, it's a cons cel `(pkg . dir)` this is why the dot.

etc/ESSR/R/completion.R
vspinu: Which part is slopy? the .ess_eval? or the getAnywhere. I see the problem with .ess_eval because it will take the first occurence in the search path, but getAnywhere is ok, it will find all occurrences of the symbol. 

ess_eval is needed for complexities like `x$foo`. Probably better to revert the order then, first getAnywhere then evel.

lisp/ess-inf.el
vspinu: crap, commuted some unrelated code. Sharp eye!

etc/ESSR/R/completion.R
vspinu: Actually, first eval then getAnywhere might be ok. If the user is inside a package source he/she most likely wants to get symbols from the package first, and that's what first eval will do presumably. Alternatively the output of findAnywhere can be resorted to push current package on top. Seems like complication for little benefit.

lisp/ess-r-xref.el
lionel-: I'm not a fan of this convention because it makes the code look messy, also it's the first time I see this. If it is important to mention it's a cell, maybe `cell-pkg-dir`? Otherwise I'd just use `pkg-info` or something like that.

vspinu: Well, style is 80% what you are used to. I think we should use letters to full extent to convey the semantic meaning of the name, not jus stubbornly stick to `-` or `_` etc. Not a big deal though, I can call it `loc` or something. This `car.cdr` it's my own thing. I think I used it in ESS as well.

lionel-:
> I think we should not take strictly dev prospective. One might want to inspect a code from a package which is not developed, or if developed not loaded with load_all.

I'm not sure I understand. There is no issue in the load-all case even without this patch because all the symbols are in the search path. If we evaluate in the `asNamespace()` for the current package, there will be no issue without load-all as well. (Actually even for the load-all case, it will fix the corner case of another package on the search path masking imported symbols, so that's a win for devtools users as well).

I like the idea of using getAnywhere when the package mode is not active though. But instead of using the first function, it should ideally propose a list of all potential locations to the user. For package development though, I would really prefer to only usie imported symbols. Getting transported to the wrong location in case of duplicates is certainly not an option in this case, and getting a list of potential locations even though there is only one relevant location (the imported one) would not be great either.

So I think we should use `asNamespace()` when package-mode is enabled, which should fix all your xref issues when you're not using load-all. And use `getAnywhere()` otherwise, with a note to improve this in the future to propose a list of locations to the user. What do you think?

etc/ESSR/R/completion.R
lionel-: I like the ESS eval, my concern is with get-anywhere when the package mode is active, for the reasons exposed in https://github.com/emacs-ess/ESS/pull/1015#issuecomment-632587606.

lionel-: Can you add a comment that we're evaluating instead of using `get(inherits = TRUE)` so that it works with namespaced symbols please?

lisp/ess-r-xref.el
lionel-: In my opinion it's unemacsy to use special symbols in names, like `?` for predicates etc. Maybe we should discuss this separately and use the conventional style for now? It's certainly easier to just use Emacs core conventions, when they exist.

lionel-:
> I wonder how we can popularize ess-r-package-library-paths a bit more. I expect few people know that in order to make xref work that one is necessary.

Now I understand why it's necessary. I think it's mostly for users  of `devtools::install()` which first copies the package in a temporary directory before installing. This messes up the srcref paths:

```r
devtools::install("~/Dropbox/Projects/R/tidyverse/dplyr", keep_source = TRUE)

attr(attr(dplyr::bind_rows, "srcref"), "srcfile")
#> /private/var/folders/b9/1vbq6rn93_1fk71sn95dqb8r0000gn/T/RtmpiR1dwr/R.INSTALL11a14ebe7869/dplyr/R/bind.r
```

Setting `ess-r-package-library-paths` circumvents this problem.

vspinu:
>  it should ideally propose a list of all potential locations to the user.

Indeed, that's the plan. This is how xref works in other places.

lisp/ess-r-xref.el
vspinu: I surely understand where you are coming from and that "." does iritate me for the same reasons. I will change it to `loc`. Still I think it's a matter of getting used to.

Issue #1008, opened 2020-05-03
Tagged prompt detection


lionel-:
Prompt detection is needed for these tasks:

- Navigation and font-locking
- Set busy status of process
- Post processing output:
  - Remove successive continuation prompts: `> + + + >`
  - Add newline after intermediate prompts.

The _continuation_ prompts are displayed when incomplete expressions are sent to R. When long paragraphs are evaluated, this results in long lines of `+` which we would like to strip from the output.

The _intermediate_ prompts are displayed when multiple complete expressions are evaluated. They separate the outputs of different expressions separated by newline. Expressions separated by semi-colons do not get an intermediate prompt. These intermediate prompts are annoying because they cause output to be misaligned by 2 characters, which is especially problematic when the first line of output are column names.

With this input:

```r
0 +
    +
    1
3;4
5
```

R outputs two continuation prompts and two intermediate prompts:

```
+ + [1] 1
> [1] 3
[1] 4
> [1] 5
>
```

Prompt detection is important to ESS but it is tricky to get right. Prompt massaging is even trickier and we've gotten extra newlines in outputs lately. Also, the code is getting quite complex. So I'd like to propose a more robust way of detecting prompts, and simpler post-processing behaviour.

I think the way to solve prompt detection it to mark the R prompts with an unlikely sequence of ANSI escapes. We'll use that sequence to robustly detect the prompts created by the REPL. Using ANSI escapes to mark the prompts reduces the chance of leaking the markers where they shouldn't. In most cases, the escapes will be processed out of the output. TRAMP uses a similar strategy for detecting the echo of commands, see `tramp-echo-mark`.

In addition to tagged prompts, I think we should simplify the post-processing.

- Remove continuation prompts without replacement, unless they are trailing. The current replacement isn't very helpful, and instead of `+ . + output` I'd rather just see `output`. I'd prefer to keep things simple, but no strong feeling about keeping the current behaviour and making it customisable though, if you prefer the `.` thing.

- Even if we can detect intermediate prompts robustly, it's not possible to reliably detect when a new line should be inserted. It makes sense when a data frame is printed but not when a string or number is printed. Maybe we should only insert a newline when the output is multiline? However that won't solve the `source(echo = TRUE)` issue, so maybe we should stop trying to insert newlines?

[...]

Issue #1017, opened 2020-05-24
Accumulated linewise evaluation?


lionel-:
This might be an alternative to tagged prompts (#1008) that is potentially more robust with a better console output.

The idea is to send inputs linewise to R, and wait for output with a very short delay. If we see a trailing prompt (continuation or otherwise) after that delay, send the next line immediately. Otherwise, this means the expression was complete and R is busy computing it. We save the rest of inputs in an accumuluation buffer. When we receive new output in the filter, we check for a trailing prompt (continuation or otherwise) and send the next accumulated line if any. If the user sends new inputs but R is busy, this goes in the accumulation buffer. If the user sends an interrupt, we cancel all pending inputs.

This should result in interlaced inputs and outputs like when `ess-eval-visibly` it `t` but without the freezes. Actually we might even deprecate this option and only support linewise evaluation.

Could this work? The only issue I can think of is that if an output batch artificially contains a trailing prompt, we'll send the next line even though R is still busy. This will mess up the interlacing of inputs and outputs but shouldn't otherwise have deleterious effects. Also, I can't imagine this will be a common edge case.

mmaechler:
This definitely preferable to ESS sending its own newlines, and I think conceptually this really fine.
Indeed, you are right the *robust*ness should be our primary concern.  We need something where ESS/Emacs never locks up  (the user should be able to notice when R "locks up", i.e. needs too long, but useR should *never* be lead to see emacs lockup and confusing that with R lock up).

As we have to deal with scenarios where a very large region is sent to R for evaluation, or a medium size region but one that has some parts that take 2-3, even dozens of seconds, etc, the idea of
- try to expect a quick answer, if we get it ... we get correctly formatted output,
- if not leave it to the R "terminal/console"  I/O  how the result will look like... but definitely don't spend CPU waiting for R.

Choosing the "very short delay" may be somewhat crucial: We can afford say 500ms on startup (of the R process) trying to measure speed  [which may differ from the next R process that is started in the same emacs/ess instance].   I think it may be important to "message" to the user about this delay setting and that (s)he can easily customize it.    I think an "optimal" delay is really a tradeoff between user <-> `*R*` interaction  and  "beauty" of the `*R*` process console itself, and e.g. I may want a different delay when "hacking" as opposed to when using ESS interactively when teaching applied statistics.

What I really like about this proposal that we would finally stop doing things in R behind the user's back; which is something I've loathed more and more....

[...]

PR #1018, opened 2020-05-26
Tagged prompts

lionel-:
Experiment with #1008.

Not ready for review yet: the variables and internal dependencies changes are temporary.


lionel-:
This approach is problematic because there are situations where the ansi escapes might not be trimmed. For instance when sinking output of `source()` to a file.

Closing in favour of the async linewise queue proposed in #1017.

PR #1019, opened 2020-05-27
[Fix #956] Add inferior-ess-fix-misaligned-output

### LINKED ISSUE #956: `source("....", echo=TRUE)`  produces prompts wrongly in iESS `*R*`
As the subject says:  The current "prompt searching" and similar hoop jumping really kill the nice `echo = TRUE` behavior of  `source()`  by adding extra newlines the result of which is very ugly.

I have set `ess-eval-visibly` to (non default) `nil` .. but don't really expect it to make a difference

vspinu:
Add new `inferior-ess-fix-misaligned-output` defcustom to control the output offset hackery. 

I am leaving the default to `t` because I think it works in most cases and it has been the behavior for some years now. TBD.

@mmaechler do you have other issues that you would like to be dealt with before the release?

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,6 +4,10 @@
 Changes and New Features in 19.04 (unreleased):
 @itemize @bullet
 
+@item ESS[R]: Automatic offsetting of R process output is now disabled by default
+because it produces undesirable output in some situations. To re-enable,
+set @code{inferior-ess-fix-misaligned-output} to t.
+
 @item ESS[R]: Improved @code{xref} lookup (@kbd{M-.}).
 Function locations are now always detected for package libraries listed
 in @code{ess-r-package-library-paths}.
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -564,6 +564,11 @@ can use `ess--busy-slash', `ess--busy-B',`ess--busy-stars',
 (defvar ess--busy-timer nil
   "Timer used for busy process indication.")
 
+(defcustom inferior-ess-fix-misaligned-output nil
+  "If non-nil, try to correct misaligned process output."
+  :group 'ess-tracebug
+  :type 'boolean)
+
 (defcustom inferior-ess-replace-long+ t
   "Determines if ESS replaces long + sequences in output.
 If 'strip, remove all such instances.  Otherwise, if non-nil, '+
@@ -1306,8 +1311,9 @@ value as it might be a continuation prompt."
          (t (error "Invalid values of `inferior-ess-replace-long+'")))))))
 
 (defun ess--offset-output (prev-prompt str)
-  "Add suitable offset to STR given the preceding PREV-PROMPT."
-  (if prev-prompt
+  "Add suitable offset to STR given the preceding PREV-PROMPT.
+Do nothing if `inferior-ess-fix-misaligned-output' is nil."
+  (if (and inferior-ess-fix-misaligned-output prev-prompt)
       (let ((len (length prev-prompt)))
         ;; prompts have at least 2 chars
         (if (eq (elt prev-prompt (- len 2)) ?+)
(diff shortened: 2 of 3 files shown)


lionel-:
Personally I would prefer to make it internal and turned off by default, since we're going to remove this in the next release anyway, and since it is buggy (we still have issues with tibble output right?). The ideal way to fix misaligned output is to get R to (optionally) output the echo for intermediate outputs (not the first one), this would fix the issue in all IDEs.

vspinu:
> make it internal and turned off by default,

Ok, for internal, but let's see if people still see tibble problems as I don't see any.

vspinu:
@lionel- I think the users who experience the tibble issues are in the minority. So I would say we go with the default on and no-custom var as you have suggested.

lionel-:
Sounds good. But I thought Martin considered this a blocker for the release?

vspinu:
Now that we have an option to disable it, I think it should not be a blocker anymore @mmaechler

mmaechler:
Well, the default of the new option should be the *previous*-ESS correct behavior in those cases that were as broken as #956 (`source(*, echo=TRUE)`).   ==> The default must not be `t` but `nil`.
Once that's changed, then yes, go ahead merging and have one less blocker for release.

mmaechler:
I should also say the variable name is misleading:  Unless I completely misunderstand, it must be set to `nil` in order that a very old basic R function   `source(*,  echo=TRUE)`  works correctly.
So, in this case it is not at all about "fixing misaligned"  it is about "not destroying correctly aligned"...

vspinu:
@mmaechler the behavior in question is about adjustment of te output. Basically each time we send something to the inferior the output would appear immediately after prompt, unless we fix it, like so:

```
> > . + >     Sepal.Length Sepal.Width Petal.Length Petal.Width    Species
1            5.1         3.5          1.4         0.2     setosa
2            4.9         3.0          1.4         0.2     setosa
3            4.7         3.2          1.3         0.2     setosa
4            4.6         3.1          1.5         0.2     setosa
5            5.0         3.6          1.4         0.2     setosa
6            5.4         3.9          1.7         0.4     setosa
```

This is why it's called `fix-misaligned-output`, because ESS fixes the misaligned output produced by R. IMO this is a good an very useful fix, even if it doesn't work in 100% of cases. 

The unfortunate side effect of the adjustment is the interference with `source(..., echo = TRUE)` that you reported. 

Given the above, are you completely sure that you would like to forego the benefits of output adjustment for the benefit of `source(...,echo=T)` by default for all users? Or maybe you could be fine  by setting that variable yourself?

mmaechler:
Yes, I am completely sure:  You know that "everything is possible" with a language like emacs lisp; hence it is possible in theory to do your change in a way that it does not break previously working correct behavior : 
Currently you want to "fix" something that > 25 years of ESS have easily survived, and people have not complained about at least not for the first 23 years.  So I would rather call that "fix" a  "nice to have" / "optional prettification".
On the other hand, your "fix" breaks an R function that has worked blamelessly for ~ 15 years, and continues to do so in all other "consoles" but ESS (for a few months now).      In all open source software one would not introduce such a change unless it's turned off by default, and actually one would keep it an important issue to improve that change such that eventually it does work when it should but does not "bad work" when it should not .. and only then close the issue... and then probably turn the option *on* by default, only then.

vspinu:
This fix is not for a couple of months. It's there for 10 years and it's as old as a tracebug itself. It just became smarter since 2 years , thus detecting more cases. 

In any case, I am changing the default as I am getting tired of the release stalemate. But it's not an ideal outcome. I expect most people to want the non-default behavior. So they all will have now to go and try to figure out what changed, read NEWS and set the variable. Software is constant battle of tradeoffs, and we should strive to choose the ones what are best for the users.

jabranham:
With the `nil` value and `ess-eval-visibly` set to `nil`, output begins on the line of the prompt. Is that an easy fix?

PR #1022, opened 2020-06-18
Only strip roxy prefix if the whole input is prefixed

### LINKED ISSUE #1020: Only strip roxy prefix when the whole input is prefixed
See https://github.com/emacs-ess/ESS/commit/1e18deda5ccad9c0f9c54a4cf0ceb5f6c9c1b77f#commitcomment-39806833

lionel-:
Closes #1020

Currently does not try to detect leading `#' @examples` field or rmarkdown fences.


vspinu:
Does this work with `eval-paragraph`?

lionel-:
Yes it is run by `ess-presend-filter-functions` and so should work with all multi-line evaluation functions, including `ess-eval-paragraph` and `ess-eval-buffer-from-beg-to-here`.

PR #1028, opened 2020-07-20
Clean up tests

lionel-:
- Properly kill the inferior process and buffer after test is finished. The buffer is killed synchronously to avoid it being accidentally reused at a later time in another test.

- Pass session buffer explicitly in org-mode tests, otherwise they reuse `*R*`.

- Don't set transient map unless called interactively.

Fixes broken tests on master.

### DIFF
--- a/test/ess-test-inf.el
+++ b/test/ess-test-inf.el
@@ -89,16 +89,17 @@
   ;; (ess-async-command "{cat(1:5);Sys.sleep(5);cat(2:6)}\n" nil (get-process "R")
   ;;                    (lambda (proc) (message "done"))
   (with-r-running nil
-    (let (semaphore)
+    (let ((inf-proc *proc*)
+          semaphore)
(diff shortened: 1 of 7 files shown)


test/ess-test-inf.el
```diff
@@ -89,16 +89,17 @@
   ;; (ess-async-command "{cat(1:5);Sys.sleep(5);cat(2:6)}\n" nil (get-process "R")
   ;;                    (lambda (proc) (message "done"))
   (with-r-running nil
-    (let (semaphore)
+    (let ((inf-proc *proc*)
```
lionel-, line 92: I now realise a cleaner syntax would explicitly bind the inferior variables like this:

```lisp
(with-r-running (inf-buf inf-proc) file ...)
```

The binding would be optional:


```lisp
(with-r-running (inf-buf) file ...)
(with-r-running nil file ...)
```

PR #1030, opened 2020-07-21
Review visual test framework

### LINKED ISSUE #713: Improve ELT error messages
Is there a way to see where in the "literate" tests an error occurs? All Travis tells me is 

```
     ((should
       (string= src-string output))
```

and then it echos the whole file into the buffer.

### LINKED ISSUE #1013: broken point ("cursor") movement when (eval)ing inside `ess-r-transcript-mode`
In previous versions of ESS, navigating *and* evaluation R code was phantastically simple, thanks to our `ess-r-transcript-mode`  which is turned on automagically, e.g., for files ending in `*.Rout` (including those `*.Rout.save` ones, known from base R and package testing).

Let's take one R's own files, download it locally to open in emacs (om a decent OS, in the shell/terminal:)

`     wget https://svn.r-project.org/R/trunk/tests/reg-tests-2.Rout.save
`

If you open in it in an emacs with a correctly working ESS, I use ESS 18.10.2, also starting R, e.g. by
`
emacs -f R  <your_path_to>/reg-tests-2.Rout.save
`
and go to the first prompt, e.g., by `C-s Reg`  you can press `[Enter]` repeatedly to evaluate things to the `*R*` buffer.  At the beginning this also works nicely with current ESS (from github, MELPA, ..).
Things start differing between correctly working and current ESS dramatically, once you get to line 29 (of `reg-tests-2.Rout.save`):  In current ESS, on `[Enter]` there, point (aka "the cursor") just goes one line further (line 30), which is a "+ ...." continuation line of a multi-line evaluation, I'm appending a small screen shot (using `linum-mode` to show line numbers).
Bildschirmfoto von 2020-05-18 17-09-57

whereas in ESS 18.10.2 (and *all* earlier versions of ESS), it nicely jumps down to line  69  which is exactly where the next prompt is ... so you can continue the evaluations by continuing with `[Enter]`.

Probably much related to the similarly "new" bugous behavior in issues #956 and #973 .

All of these are probably related to prompt detection, and all of these were working correctly in all released versions of ESS and also for quite some time after releasing ESS 18.10.2 (2018-11-10).

lionel-:
Branched from #1029.

This is a complete review of the visual approach to unit testing. This approach is easier to implement and should be more maintainable than the current one. It is inspired by the ERT and use-package interfaces.

Despite recent efforts, ESS still has very few unit tests because it is hard to test side effects. The goal of this framework is to make it easy to test the side effects in Emacs buffers, mainly motion and mark commands, and output in inferior buffers. There is now documentation in `test/etest/README.md`. You can find a rendered version at https://github.com/lionel-/ESS/tree/clean-elt/test/etest. A review of this documentation would be particularly welcome.

For illustration the last commit implements a test for #1013 (closes #1013):

```elisp
(etest-deftest ess-r-transcript-motions-test ()
  "[enter] handles commands, non-commands, and prompts (#1013)."
  :init ((eval . (ess-r-test-transcript-init)))
  :case "
¶R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.
"
  :test "RET"
  :result "
R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.
¶"

  :inf-result "
> "

  :messages "No command at this point"

  :case "
> ¶for(i in 1:3) {
+   print(i)
+ }
[1] 1
[1] 2
[1] 3
"
  :test "RET"
  :result "
> for(i in 1:3) {
+   print(i)
+ }
[1] 1
[1] 2
[1] 3
¶"

  :inf-result "for(i in 1:3) {
+   print(i)
+ }
[1] 1
[1] 2
[1] 3
> ")
```

This tests the effect of the `"RET"` command in a `ess-transcript-mode` buffer. The following effects are checked:

- Cursor motion as user types RET (cursor is represented with ¶).
- Output in the inferior buffer.
- Messages emitted by the command.

With an ERT-like deftest block approach, we no longer test whole files in bulk. Closes #713.

My favourite feature is `M-x etest-update`. To create the unit test above, we start with empty comparison cases:

```elisp
(etest-deftest ess-r-transcript-motions-test ()
  "[enter] handles commands, non-commands, and prompts (#1013)."
  :init ((eval . (ess-r-test-transcript-init)))
  :case "
¶R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.
"
  :test "RET"
  :result ""

  :inf-result ""

  :messages ""

  :case "
> ¶for(i in 1:3) {
+   print(i)
+ }
[1] 1
[1] 2
[1] 3
"
  :test "RET"
  :result ""

  :inf-result "")
```

Then, with point inside the deftest block, call `M-x etest-update`. This runs the test commands and fills all the empty strings with the outputs, producing the test above.

Compared to the current literate approach, we lose the ability of editing the test case in the relevant mode. It should be possible to add polymode support though. Also, it would be trivial to extend it to support case and results files (including with the `etest-update` command to update the results files when they are outdated).

I haven't switched all the tests in `test/literate` to this new approach yet but I plan to do it once this PR is merged.


test/ess-test-r-utils.el
```diff
@@ -98,6 +98,21 @@ inserted text."
...
+
+(defvar ess-r-test-proc-buf
+  (let* ((buf (run-ess-test-r-vanilla))
+         (proc (get-buffer-process buf)))
+    (ess-wait-for-process proc)
+    (with-current-buffer buf
+      (let ((inhibit-read-only t))
+        (erase-buffer)))
+    buf)
+  "Common process buffer for tests.")
```
jabranham, line 110: This creates a process whenever this file is loaded, right? I'm not sure that's the right approach; I'd like to be able to have the file load without side effects like that. (I'm also not sure of the behavior on byte compilation)

test/etest/README.md
```diff
@@ -0,0 +1,256 @@
+
+# Introduction
+
+etest is an ERT extension to make it easier to test Emacs behaviour such as cursor and mark movement and inferior buffer output..
```
jabranham, line 4: British spelling & double period typo at the end. There's also britishism "initialisation" a few times later on

test/etest/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+
+## Defining test cases
+
+At its heart, the `etest-deftest` macro is an interpreter of keywords command. The main keywords to know about are `:case`, `:test`, and `:result`.
+
+
+### `:case`
+
+All etest tests are run in a dedicated buffer. The `:case` takes a string that defines the content of the test buffer. The special characters `¶` and `×` specify the positions of the Emacs point and mark cursors in that buffer. These characters were chosen to be 1 character wide so they don't distract from the general structure of the test case, and outside of the ASCII range so they don't conflict with valid symbols in the tested languages.
```
jabranham, line 32: Specify what happens when those two (or one) are omitted?

test/etest/README.md
```diff
@@ -0,0 +1,256 @@
...
+(etest-deftest my-test ()
+  :case "¶if ¶while ¶for"
+  :test "M-f"
+  :result "if¶ while¶ for¶")
+```
+
+Or for testing properties of many keywords:
+
+```elisp
+(defun face-at-point ()
```
jabranham, line 219: face-at-point is a builtin function

test/ess-test-r-utils.el
lionel-: Good point, we could generate it lazily with a closure instead.

test/etest/README.md
lionel-: Thanks but I think I prefer that etest uses the British/European spelling of English.

test/ess-test-r-utils.el
lionel-: done!

test/etest/README.md
jabranham: Could you reconsider? The rest of Emacs uses American spelling for the sake of consistency.

lionel-: I think this is the sort of things that should be consistent within projects rather than across projects. But since you feel strongly about it, I have changed to American spelling.

lionel-:
> I didn't see an easy way to test, do something, then test again. Did I miss it? I saw the multiple cursor stuff but this is different I think. How to, for example, check that ess-cycle-assign cycles through the assignment operators?

etest tries hard to set the current Emacs state of Emacs as if commands were run interactively, so to test cycling you just repeate the same command multiple times. This should work with multiple cursors too (state is reset for each cursor).

I have changed the insert- and cycle-assign tests to use etest:

```elisp
(etest-deftest ess-insert-assign-test ()
  "Repeated calls cycle between assignment and self-insert."
  :init ((mode . r)
         (eval . (let ((map (make-sparse-keymap)))
                   (define-key map "_" 'ess-insert-assign)
                   (use-local-map map))))
  :case "foo¶"
  :eval "_" :result "foo <- ¶"
  :eval "_" :result "foo_¶"
  :eval "_" :result "foo_ <- ¶")

(etest-deftest ess-cycle-assign-test ()
  "Repeated calls cycle trough assignment operators."
  :init ((mode . r))
  :case "foo¶"
  :eval "C-c C-=" :result "foo <- ¶"
  :eval "C-c C-=" :result "foo <<- ¶"
  :eval "C-c C-=" :result "foo = ¶"
  :eval "C-c C-=" :result "foo -> ¶"
  :eval "C-c C-=" :result "foo ->> ¶"
  :eval "C-c C-=" :result "foo <- ¶"
  :eval "C-c C-=" :result "foo <<- ¶")
```

PR #1034, opened 2020-08-16
Fix #950: Delay stripping of tramp prefix in project path

### LINKED ISSUE #950: ess-r-package-eval-linewise problem on remote R session
The `ess-r-package-eval-linewise` function does not work properly for me when running an R session on a remote machine through Tramp.  The difficulty that I have is that the function uses the Tramp filename such as e.g. `"/ssh:remote.server.name:/home/username/packagename"` rather than `"/home/username/packagename"`, the former of which R obviously doesn't understand.

This issue affects the `ess-r-devtools` family of functions.

The `ess-r-package-eval-linewise` function gets the path string through a call to `ess-r-package-info` which in turn gets the value from a call to `ess-r-package--find-package-path`.  It appears to me that somewhere along the way it might be desirable to drop the Tramp portion of the path string to enable the correct functionality on remote machines.

Two functions that might be helpful for such a task are `tramp-tramp-file-p` and `tramp-dissect-file-name`.  So for example, for a path string `path`, one could do the following:
```
(if (tramp-tramp-file-p path)
    (nth 6 (tramp-dissect-file-name path))
  path)
```
I would be happy to submit a PR inserting such a section of code if it were decided where the best place for it.  One thing that I don't have any understanding of though is how package dependencies (i.e. the dependency on tramp) in Emacs are specified and whether any code would need to be added in that regard.

dapritchard:
Fixes #950.

The issue is that tramp portion of the filepath was being stripped before `ess-r-package--find-package-name` attempts to use the path to find the package's DESCRIPTION file, and consequently was looking in the local filesystem rather than the remote.  This patch takes the responsibility of stripping the tramp portion of the filename out of the hands of `ess-r-package--find-package-path` and leaves it to the calling routine to do so, if desired.

`ess-r-package-info` is the only function that calls `ess-r-package--find-package-path`, so this change shouldn't cause any unintended problems.

### DIFF
--- a/lisp/ess-r-package.el
+++ b/lisp/ess-r-package.el
@@ -122,9 +122,12 @@ efficiency reasons."
     (let* ((path (ess-r-package--find-package-path (or dir default-directory)))
            (name (when path
                    (ess-r-package--find-package-name path)))
+           (local (if (and path (file-remote-p path))
+                      (tramp-file-name-localname (tramp-dissect-file-name path))
+                    path))
            (info (if name
                      (list :name name
-                           :root path)
+                           :root local)
                    '(nil))))
       ;; If DIR was supplied we cannot cache in the current buffer.
       (if dir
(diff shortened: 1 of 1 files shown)


jabranham:
LGTM, thanks!

jabranham:
I merged this in, not sure why github didn't automatically close this PR.

Issue #876, opened 2019-03-26
Drop Emacs 25


jabranham:
How do people feel about announcing dropping support for Emacs 25 after the 19.04 release? Emacs 26.1 has been out about a year already, and Emacs 26.2 should be out any day now. A year is plenty of time to upgrade. And those who are stuck on 25 can continue to use 19.04. WDYT?

lionel-:
I think 1 year is too short because people tend to delay their Emacs upgrades. What features do you need from 26.1?

vspinu:
Way to early indeed. 27 is not released, nor even 26.2. Supporting 25 is not hard at all.

mmaechler:
Yes!   Our users are *not* mainly emacs hackers and PhD students!
If you work in a bigger company (outside of IT), e.g., pharmaceutical (!),
you cannot easily update emacs at all.  Nor in a bank or insurance company, nor at FDA, or .. or ..

jabranham:
@lionel- It'd be more like 1.5 years, since 19.04 would support 25 and the release after that (19.10?) wouldn't. But that's minor. I've already been bitten several times by Emacs 25 stuff, most recently with `cXXr` not existing in elisp until 26. It'd also be nice to be able to simplify the code in several places (especially the hacky way we deal with flymake right now since it requires 26). The optional arguments to `string-trim` would be nice to use too. And we could remove our backport of `ess-display-buffer-reuse-mode-window`. And that's just off the top of my head.

I still haven't heard a convincing case against dropping support for Emacs 25. @vspinu it's not hard, but it does create some uglier code than needed. And @mmaechler those users unable to update Emacs can live with the older ESS version.

But if we're keeping 25 support through at least the release after next, my question is when are we dropping it?

vspinu:
> my question is when are we dropping it?

Shortly after emacs 28 is out. I think we should what most emacs projects do. Support last 2 stables of emacs.

jabranham:
So let's say Emacs 27.1 is released May 2020 and ESS continues our current 6-month XX.04/10 release cycle.

Then ESS 20.10 mentions that next release we drop Emacs 25, then actually dropping support in 2021? Emacs 26 will be 3 years old at that point. I think it's a waste of our limited time to think about supporting people who insist on staying on old Emacses that long. If they want an old Emacs, they can run an old ESS.

vspinu:
[...]

PR #1038, opened 2020-08-18
Require Emacs 26 or newer

jabranham:
With the release of Emacs 27.1 we should remove support for Emacs 25. Emacs 26.1 was released over two years ago which gives users plenty of time to upgrade. The distros I looked through (debian, fedora, nixOS, arch, etc) are all shipping 26.1 or newer. If someone wants to run Emacs 25, they can also run an older version of ESS.

Note to self to merge #989 after merging this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: emacs-lisp
 env:
   # When there's a new major Emacs release, also update
   # byte-compile-error-on-warn to the new release, below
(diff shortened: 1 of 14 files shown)


lionel-:
We may need to finish the discussion in #876.

jabranham:
I thought the conclusion there was to support the last 2 major stable versions?

PR #1039, opened 2020-08-25
add: utils to view data

ShuguangSun:
It is a collection of utils for viewing data and now it implemented the functions for R based on dplyr or data.table. 

I don't have time to maintain it and would like to contribute it to ESS. It will be helpful as doing it in Emacs.

```
Customization:
ess-view-data-backend-list: dplyr (default), dplyr+DT, data.table+magrittr
ess-view-data-print-backend-list: print (default), kable
ess-view-data-save-backend-list: write.csv (default), readr::write_csv, data.table::fwrite kable
ess-view-data-complete-backend-list: jsonlite
ess-view-data-read-string: ess-completing-read (default), completing-read, ido-completing-read, ivy-completing-read

Utils:
NOTE: it will make a copy of the data and then does the following action
ess-view-data-print: the main function to view data
ess-view-data-set-backend: change backend
ess-view-data-toggle-maxprint: toggle limitation of lines per page to print
ess-view-data-filter:
ess-view-data-select / ess-view-data-unselect
ess-view-data-sort
ess-view-data-group / ess-view-data-ungroup
ess-view-data-mutate
ess-view-data-slice
ess-view-data-wide2long / ess-view-data-long2wide
ess-view-data-update
ess-view-data-reset
ess-view-data-unique
ess-view-data-count
ess-view-data-summarise
ess-view-data-overview
ess-view-data-goto-page / -next-page / -preious-page / -first-page / -last-page / -page-number
ess-view-data-save
```

### DIFF
--- a/lisp/ess-view-data.el
+++ b/lisp/ess-view-data.el
@@ -1,0 +1,1955 @@
+;;; ess-view-data.el --- View Data                   -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2019  Shuguang Sun
+
+;; Author: Shuguang Sun <shuguang79@qq.com>
+;; Keywords:
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


lionel-:
> I don't have time to maintain it

We're all a bit pressed regarding time as well, so we don't really have time to maintain this either ;)

Nowadays I think there's little sense in adding large chunks of non-generic functionality in ESS. We have a nice packaging and distribution system with MELPA (and soon, non-GNU ELPA as was just announced on emacs-devel). So probably the way forward would be to create your own Emacs package and declare it in maintenance mode (only bugs fixed). It could be only available on Github if you're afraid having users would create a workload, or you could add it to MELPA to make it easier to install.

I don't know what others think though.

ShuguangSun:
Thanks for the information.
It might not the time. I'll put it in melpa.
You can contact me any time.

lionel-:
Thanks. I think eventually we should think about a place to document where to find ESS-friendly packages like yours.

PR #1040, opened 2020-08-25
Improve buffer switching in `ess-request-a-process`

### LINKED ISSUE #987: display-buffer-alist reusable-frames does not work as expected

lionel-:
Supersedes and closes #989.
Closes #987.

@jabranham I have pushed your cleanups on master so we can see things more clearly :) https://github.com/emacs-ess/ESS/commit/cb75588d972b97e597c42108b3c83c2f67947cac

When I move the `let` around `ess-start-process-specific`, which seems a better place for it, there is no longer CI issues for Emacs 25. Maybe stemmed because the overriding action was still in place when we used `pop-to-buffer` when `no-switch` is non-nil, which was incorrect IIUC, but I'm not sure.

Also use `save-current-buffer` to prevent `pop-to-buffer` from setting it. I think this was the original motivation for Vitalie's change, though I'm unsure as well.

This new setup seems to make sense and I checked that it fixes #987. If I have time I'll try to think about unit tests for this sort of window behaviour.

It might make sense to have our own `ess--pop-to-buffer` wrapper that can be easily disabled with a let-bound global variable because there are still some effects from `pop-to-buffer` that we can't prevent.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -734,6 +734,11 @@ to `ess-completing-read'."
                            (delete-dups (list "R" "S+" (or (bound-and-true-p S+-dialect-name) "S+")
                                               "stata" (or (bound-and-true-p STA-dialect-name) "stata")
                                               "julia" "SAS")))))
+         ;; Set `display-buffer-overriding-action' here since we
+         ;; explicitly handle buffer switching below with
+         ;; `pop-to-buffer' which handles the `display-buffer'
+         ;; framework.
+         (display-buffer-overriding-action '(display-buffer-no-window (allow-no-window . t)))
          (pname-list (delq nil ;; keep only those matching dialect
                            (append
                             (mapcar (lambda (lproc)
@@ -780,8 +785,7 @@ to `ess-completing-read'."
                     (process-name (get-buffer-process buf))
                   (ess-start-process-specific ess-language ess-dialect)
                   (caar ess-process-name-list))))))
-    (if noswitch
-        (pop-to-buffer (current-buffer)) ;; VS: this is weird, but is necessary
+    (unless noswitch
       (pop-to-buffer (ess-get-process-buffer proc)))
     proc))
 
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -719,6 +719,19 @@ LANGUAGE is ignored."
         (error "No ESS processes running; not yet implemented to start (%s,%s)"
                language dialect)))))
 
+(defmacro ess--with-no-pop-to-buffer (&rest body)
+  "Disable some effects of `pop-to-buffer'.
+Prevent `display-buffer' from performing an action and save the
+current buffer to prevent `pop-to-buffer' from setting a new
+current buffer."
+  ;; `pop-to-buffer' might still raise windows and frames so it may be
+  ;; better to have our own configurable `ess--pop-to-buffer' wrapper.
+  (declare (indent 0)
+           (debug (&rest form)))
+  `(let ((display-buffer-overriding-action '(display-buffer-no-window (allow-no-window . t))))
+     (save-current-buffer
+       ,@body)))
+
 (defun ess-request-a-process (message &optional noswitch ask-if-1)
   "Ask for a process, and make it the current ESS process.
 If there is exactly one process, only ask if ASK-IF-1 is non-nil.
@@ -734,11 +747,6 @@ to `ess-completing-read'."
                            (delete-dups (list "R" "S+" (or (bound-and-true-p S+-dialect-name) "S+")
                                               "stata" (or (bound-and-true-p STA-dialect-name) "stata")
                                               "julia" "SAS")))))
-         ;; Set `display-buffer-overriding-action' here since we
-         ;; explicitly handle buffer switching below with
-         ;; `pop-to-buffer' which handles the `display-buffer'
-         ;; framework.
-         (display-buffer-overriding-action '(display-buffer-no-window (allow-no-window . t)))
          (pname-list (delq nil ;; keep only those matching dialect
                            (append
                             (mapcar (lambda (lproc)
@@ -783,7 +791,11 @@ to `ess-completing-read'."
               (let ((buf (ess-completing-read message (append proc-buffers (list "*new*")) nil t nil nil)))
                 (if (not (equal buf "*new*"))
                     (process-name (get-buffer-process buf))
-                  (ess-start-process-specific ess-language ess-dialect)
+                  ;; Prevent new process buffer from being popped
+                  ;; because we handle display depending on the value
+                  ;; of `no-switch`
+                  (ess--with-no-pop-to-buffer
+                    (ess-start-process-specific ess-language ess-dialect))
                   (caar ess-process-name-list))))))
     (unless noswitch
       (pop-to-buffer (ess-get-process-buffer proc)))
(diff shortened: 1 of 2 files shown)


lisp/ess-inf.el
```diff
@@ -719,6 +719,19 @@ LANGUAGE is ignored."
         (error "No ESS processes running; not yet implemented to start (%s,%s)"
                language dialect)))))
 
+(defmacro with-ess--no-pop-to-buffer (&rest body)
```
jabranham, line 722: I don't think we need a whole new macro for a piece of code we only use once. If/when we need to do this more we can implement a macro then. 

I also think we should properly namespace functions/macros/variables with `ess-`.

lionel-: I don't think it's a big deal to have this internal macro, it makes the code simpler to read and understand.

We already discussed the namespacing of `with-` macros, if we are going to change the naming convention it should be in a different PR after a new discussion.

lionel-:
@jabranham Do you agree that it's better for the final pop-to-buffer to be outside the no-pop-to-buffer context? I'm not sure how it could work.

lisp/ess-inf.el
jabranham: I never said it was a big deal :-) I don't see how having a whole new macro is simpler to read or understand than a single let binding though. The name also isn't 100% accurate as we're overriding the display-buffer framework which is slightly different than pop-to-buffer. It should be something like `ess--display-buffer-no-window`.

I also don't recall settling the `with-` conversation one way or the other. The elisp manual is pretty clear that packages should namespace all functions/variables/macros with their name, especially noninteractive ones; see `(info "(elisp) Coding Conventions")`.

jabranham:
@lionel- I'm not sure I understood that question. Doesn't the final call to pop-to-buffer /have/ to be outside the no-pop-to-buffer stuff?

lisp/ess-inf.el
lionel-: > than a single let binding though.

Fully agreed because the let binding is informative through the variable name (provided it's informative). Please note that I created the macro when a single let binding was no longer sufficient. There's a with-current-buffer in there now. Surely you'd wonder what it's doing there so we'd need a comment and then the code gets stuffy and the main intent (starting a new process) no longer stands out because it's mixed with other concerns.

> The name also isn't 100% accurate as we're overriding the display-buffer framework which is slightly different than pop-to-buffer.

pop-to-buffer is in the display-buffer framework IIRC? It just has some other logic for frame, window, and buffer selection that the macro is trying to cancel out. Or is there a purely display-buffer alternative? Is that documented somewhere?

Regarding the name I think it is good because it corresponds to the pop-to-buffer calls we currently have. It is an internal name so can easily change with the implementation and does not need to be abstract, the concrete name reflecting the current implementation is actually better for now. Most of all, since it's internal we shouldn't spend too much time discussing it.

>  also don't recall settling the with- conversation one way or the other. The elisp manual is pretty clear that packages should namespace all functions/variables/macros with their name

It was in the discussion about inferior-ess- IIRC. I like `with-prefix-` because it helps reading the code, `with-` macros set up special contexts and so should stand out. I don't mind going against the elisp manual because that's how you make conventions evolve. I could be convinced to go with `prefix-with-` after surveying what other projects do (I'd rather spend my ESS time worrying about other things though).

lionel-:
> Doesn't the final call to pop-to-buffer /have/ to be outside the no-pop-to-buffer stuff?

I think so. I think that's why it didn't work before on Emacs 25 in #989. I was wondering why it worked (or appeared to) on newer Emacsen.

lisp/ess-inf.el
lionel-: Quick overview of styles used in the wild. `prefix-with-` dominates.

<details>
From MELPA:

clojure-mode: `with-clojure-buffer`
use-package: `use-package--with-elapsed-timer`
evil: `evil-without-restriction` etc
company-mode: `company-capf-with-buffer`
diff-hl: `diff-hl-with-diff-switches`
elpy: `with-elpy-rpc-virtualenv-activated`
flycheck: `flycheck-error-with-buffer`
magit: `magit-with-top-level`
markdown-mode: `markdown--with-gensyms`
projectile: `projectile-with-default-dir`
smartparents: `sp-with-modes`

From Emacs:

gnus: `gnus-icalendar-with-decoded-handle`, `gnus-async-with-semaphore`
subr: `subr--with-wrapper-hook-no-warnings`
tramp: `tramp-with-demoted-errors`, `with-tramp-progress-reporter`
tramp-gvfs: `with-tramp-dbus-call-method`
rcirc: `with-rcirc-process-buffer`
soap: `soap-with-local-xmlns`
eshell: `eshell-with-file-modes`
nxml: `nxml-with-degradation-on-error`
reftex: `reftex-with-special-syntax`
table: `table-with-cache-buffer`
org: `org-with-gensyms`
server: `server-with-environment`
calc: `calc-with-default-simplification`
mode-local: `with-mode-local`
ede: `ede-with-projectfile`
semantic: `wisent-with-context`
ediff: `ediff-with-current-buffer`
vc: `with-vc-properties`
newcomment: `comment-with-narrowing`
json: `json--with-indentation`
isearch: `with-isearch-suspended`
erc: `erc-with-buffer`
js: `with-js`
xref: `xref--with-dedicated-window`
cc: `c-with-syntax-table`
python: `python-shell-with-environment`
package: `package--with-work-buffer`
backtrace: `backtrace--with-output-variables`
generator: `cps--with-value-wrapper`
</details>

lionel-: In exported symbols we have one `with-prefix-` and two `prefix-with-`. The majority of Emacs projects seems to use `prefix-with-`. So let's go with that. I have renamed the symbol.

lionel-:
I'm merging this so we can get some testing before release.

PR #1042, opened 2020-09-02
[WIP] Travis CI coverage for Emacs 27 with Nix

peterbecich:


lionel-:
I think I'd prefer to switch to Github Actions if we're going to review the whole CI. By the way you can experiment with Travis configs on your fork rather than in the main repo.

lionel-:
Here is an example of Github workflow that I've seen in the wild: https://github.com/magnars/dash.el/blob/master/.github/workflows/test.yml

The Github action it uses is at https://github.com/marketplace/actions/set-up-emacs

Not sure what others (@mmaechler, @vspinu, @jabranham) think though.

* Travis-CI lost many of their engineers when they were bought off. It's unclear what their future is.

* Our Travis config uses EVM for retrieving versions of Emacs. It doesn't seem abandoned but it's not proactively maintained either. The snapshot build seems to be stuck at 27.0.50 and there is no 27.1 recipe yet.

* The Github Action for Emacs is maintained by Steve Purcell, the same person who maintains MELPA.

PR #1045, opened 2020-09-05
Make `inferior-ess-replace-long+` process-local

lionel-:
This makes it easier to disable the replacement in unit tests. It must be disabled because the replacement is not deterministic and depends on how the process output is chunked.


mmaechler (approved):
I don't have much of a clue of this inferior R buffer mangling.    But the new code does look better (and shorter --> less maintenance!).. so I'd be happy if it's merged

Issue #1024, opened 2020-07-01
ess-eval-function with tramp fails if local $TMP missing on remote


epruesse:
Trying to `C-c` a function when using Tramp with a remote R session, I get this:

```
Error in file(file, ifelse(append, "a", "w")) : 
  cannot open the connection
In addition: Warning message:
In file(file, ifelse(append, "a", "w")) :
  cannot open file '/var/folders/rq/83r63b9569g91xz05q64vd8h0000gn/T/test.R!PFnUsY': No such file or directory
> 
```

It appears that ESS is using the local value of `$TMPDIR` to create the filename for a remote temp file. Since I'm going from Mac to Linux, the `/var/folders/...` user temp dir does not exist on the remote.

I've tried altering `ess-source-directory`, the only variable I could find containing the tmp dir name, but does doesn't seem to have any effect on the file ESS attempts to write to on the remote computer.
Running `ess-eval-function` with `$TMPDIR` not existing on the remote leads to this:

epruesse:
I worked around this by adding

```lisp
(setq temporary-file-directory "/tmp")
```

to my `init.el`. It's probably still a bug to assume that `localhost:$TMP` will exist on another host.

dapritchard:
The call stack that I observe for the call to `ess-eval-region-or-function-or-paragraph-and-step` is the following, where more recent calls are shown on top.
```
ess-make-source-refd-command
ess-tracebug-send-function
ess-eval-function
ess-eval-function-or-paragraph
ess-eval-region-or-function-or-paragraph
ess-eval-region-or-function-or-paragraph-and-step
```
The problem I believe occurs in the `ess-make-source-refd-command` function. The value that is bound to `proc-dir` is the value of `default-directory` for the inferior buffer, which has a local filepath. As a result `nil` is then bound to `remote`, and then later the value bound to `tmpfile` is constructed using `temporary-file-directory`, rather than `(tramp-get-remote-tmpdir remote)`.

epruesse:
Thank you!

PR #1049, opened 2020-09-14
Reconstruct remote filepath in `ess-make-source-refd-command`

dapritchard:
As noted in #1024, running `ess-eval-region-or-function-or-paragraph-and-step` and other similar functions when the inferior R process is on a remote machine can result in an error. The error appears to be caused by trying to create a file in the temporary directory in the wrong location on the remote. The underlying issue appears to stem from the call to `ess-make-source-refd-command`. In particular, checking whether the process is on a remote is done via the call to `(file-remote-p proc-dir)`, however the string bound to `proc-dir` has already had the tramp remote server information stripped off of it. This patch uses the variable `comint-file-name-prefix` to recover the remote server information prefix.

### DIFF
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -253,10 +253,12 @@ inferior process command for loading the temporary file.  This
 command conforms to VISIBLY."
   (let* ((filename buffer-file-name)
          (proc-dir (ess-get-process-variable 'default-directory))
-         (remote (when (file-remote-p proc-dir)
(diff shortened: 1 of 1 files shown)


dapritchard:
While this fix works, I am beginning to think that maybe it is better to fix the value of `default-directory` itself. I am changing this to draft while I think through this some more.

dapritchard:
This MR is also related to #1050.

dapritchard:
Withdrawing this PR in favor of #1055.

PR #1055, opened 2020-10-11
Synchronize default-directory using remote path

### LINKED ISSUE #1001: inferior-ess-reload with remote R process
Hey all,

when I load an R-file via TRAMP and then execute `ess-eval-region`, a remote R process is started. However, when I call then `inferior-ess-reload`, a local R process is started instead of a remote one. The `setwd` call that tries to re-set the previous working directory then fails.
One has to manually close the R process, then `ess-eval-region` (or similar) starts again a new remote process.

Though I use spacemacs and found the issue there, I can reproduce the issue also with `emacs -Q -l ~/.emacs.d/elpa/26.3/develop/ess-20200328.1547/ess-autoloads.el`

Could be related to #871. I checked the "suggested" similar issues, but they do not mention this problem.

### LINKED ISSUE #1024: ess-eval-function with tramp fails if local $TMP missing on remote

### LINKED ISSUE #1050: The value of `default-directory` for inferior process buffers on a remote machine
I've been digging into the issues for #1001 and #1024, and it is my understanding that both issues stem from the value of `default-directory` for the inferior process buffer when on a remote machine. The value of `default-directory` for the ess-r-mode buffers on a remote machine have a form like `"/ssh:test.stat.internal:/share/projects/P0024/R/"`, but the value of `default-directory` for inferior R buffers on a remote machine have a form like `"/share/projects/P0024/R"` (i.e. the remote machine information is stripped off). Interestingly, if I open a comint buffer on a remote machine, then the value of `default-directory` retains the remote machine information (i.e. has a form like `"/ssh:test.stat.internal:/share/projects/P0024/R/"`).

I have so far been unable to discover what is causing the value of `default-directory` for inferior R buffers on a remote machine to lose the remote machine information. Is there a reason that this information is being stripped away when the inferior buffer is created? Would it be worth trying to figure out why this is the case and changing the behavior?

dapritchard:
Fixes #1001, fixes #1024, and fixes #1050.

In #1001 and #1024 it was noted that both issues occur due to process buffers on a remote machine having a `default-directory` value that has only the local portion of the directory string (e.g. rather than `"/ssh:10.0.1.213:/path/to/dir"` it would have `"/path/to/dir"`). This MR attempts to rectify this issue.

A new utility function `ess-path-update-local-portion` is created that can be used to update the local portion of a (possibly remote) path. As an aside, for the implementation I considered using `file-remote-p` to extract the remote portion of the path since it gave me exactly what I needed for all of the cases I tried, but I ended up using TRAMP's API directly even though it made the function a little bit more complicated since I felt more confident that it would work correctly for all cases.

The function that causes the remote portion of the `default-directory` string to be lost is `ess-synchronize-dirs`. This function is run as part of a timer that is defined in the body of `ess--idle-timer`, with the goal of keeping `default-directory` synchronized to the R process' working directory. The working directory is obtained by querying R which provides a local path, but the remote portion of the path filename is then never reattached. This patch modifies `ess-synchronize-dirs` to make use of `ess-path-update-local-portion` to reconstruct the path string with the remote information.

Additionally, it turned out that `inferior-ess-reload` had one other issue that needed to be addressed. As noted in #1001, the process file handler is determined by the value of `default-directory`. Now `inferior-ess-reload` let-binds the value of `ess-startup-directory` to the (local) path of the current working directory, as obtained by querying R. Eventually `inferior-ess` is called which then sets the value of `default-directory` based on a call to `inferior-ess--maybe-prompt-startup-directory` which in turn calls `inferior-ess--get-startup-directory` which then returns a value based on `ess-startup-directory`. Since we're already keeping `default-directory` synchronized with R's working directory, I propose simply using the value of `default-directory` for `ess-startup-directory`.


dapritchard:
It appears that `tramp-dissect-file-name` and friends have changed their interface somewhere between Emacs 25.1.1 and Emacs 26.1, which was fortunately picked up by the tests. I will have to make some changes to accommodate both versions.

dapritchard:
The latest commits in this pull request are a tale of two stories. Firstly, I added some runtime logic that adapted to the different versions of `tramp-dissect-file-name` between Emacs 25 and 26. But by the time I finished those changes, the code seemed so unnecessarily convoluted that I decided to take a second look at `file-remote-p`. After reading through the TRAMP manual and adding tests to cover what I perceived to be all the main cases, I felt confident in replacing `tramp-dissect-file-name` with `file-remote-p`. The later commits perform this replacement, which cuts `ess-path-update-local-portion` from 6 lines of code plus two helper functions and a `declare-function` to just one line of code. Indeed, we could just inline the body of that function into the necessary places if that would be preferable.

Since the history of pull request has become somewhat of a mess, it might be nice to squash the commits.

lisp/ess-inf.el
```diff
@@ -2811,9 +2812,20 @@ To be used in `ess-idle-timer-functions'."
...
   ;; synchronized automatically.
   (interactive)
-  (let ((dir (car (ess-get-words-from-vector "getwd()\n"))))
+  (let* ((dir (car (ess-get-words-from-vector "getwd()\n")))
+         (new-default-dir (ess-path-update-local-portion default-directory dir)))
     (message "(ESS / default) directory: %s" dir)
-    (setq default-directory (file-name-as-directory dir))))
+    (setq default-directory (file-name-as-directory new-default-dir))))
+
+(defun ess-path-update-local-portion (current-path new-local-path)
```
lionel-, line 2820: This function should be private. I think this signature might be clearer:

```suggestion
(defun ess--derive-connection-path (old new)
```

The docstring could be shortened to simply mention that a new connection directory is derived from `OLD` (a path) and `NEW` (a path), in such a way that the host and connection information in `OLD` is retained in the `NEW` path. `NEW` must be an absolute local path, it can't contain connection information.

lionel- (commented):
Thanks a lot for investigating these issues and coming up with a fix! I think this makes sense, but the util should be private.

lisp/ess-inf.el
dapritchard: I've updated the function signature and the docstring. Your comments made me think a little bit more about the function accepting only a local path for the input to `new`, and so I've updated the function so that this can also be a remote path.

dapritchard:
These latest commits generalize the `ess--derive-connection-path` function (renamed from `ess-path-update-local-portion (current-path`) to accept a remote path as well as a local path for the argument to `new`. When the input to `new` is a remote path the connection information is removed from the string (so that e.g. an input of `"/ssh:10.0.1.213:/path/to/file` for `new` produces the same results as an input of `/path/to/file`).

The tests for `ess--derive-connection-path` have been revamped a little bit to make it more convenient to test both the local and remote input cases for `new`.

lisp/ess-inf.el
```diff
@@ -2246,7 +2246,7 @@ START-ARGS gets passed to the dialect-specific
       ;; Use current working directory as default
       (let ((project-find-functions nil)
             (ess-directory-function nil)
-            (ess-startup-directory (ess-get-working-directory))
+            (ess-startup-directory default-directory)
```
lionel-, line 2249: The scoping isn't correct here I think. We don't want to change the directory of the process to that of the current buffer.

```suggestion
            (ess-startup-directory (ess-get-process-variable 'default-directory))
```

dapritchard: Thanks, updated.

lionel-:
Thanks a lot!

epruesse:
Thank you guys!

PR #1065, opened 2020-11-28
Use sentinel to delimit output in `ess-command`

### LINKED ISSUE #1043: Forming completions hangs emacs -- when the R prompt is `"R> "`
I'm using Emacs 27.1 (or 28) with R 4.0.2 on a Windows 10 machine.  I have  these two possibly relevant configurations:
```
(inferior-ess-r-help-command ".ess.help('%s', help.type='text')\n" )
(ess-use-auto-complete nil)
```
Follow these steps to create the bug:
```
1. Start Emacs
2. M-x R (to start inferior R process)
3. rnorm(9999)
4. ?iris
```

Sometimes, and ONLY SOMETIMES, Emacs will hang with this message "Forming completions for package:devtools".

It's so weird that the crash only happens about 25% of the time, and this makes debugging very, very time consuming and tedious.  Would love to get some suggestions on how to proceed.

Notes:
1. Important to note that the hanging happens only about 20% of the time.  The rest of the time, the help page will correctly display in another buffer.
2. I never had this problem with Emacs 26.  Only with Emacs 27 and 28.
2. I'm not sure if "rnorm(9999)" is necessary, but it seems to increase the chance of failure.
3. A similar issue was reported in the past #44.  I have disabled auto-complete (see the setting above).
4. Hanging happens with "Forming completions", but need not stop at "devtools". The hang can happen with any package in the forming-completions
4. When Emacs hangs, I cannot recover the session in any way (Ctrl-Z, Ctrl-G, Ctrl-C, Ctrl-Break, clicking on the close-window button in the titlebar, etc.)  I have to use Task Manager to kill Emacs.
5. Because of this, I am unable to get a crash dump or other diagnostic (as far as I know).  Please let me know if I'm wrong about this!!!
6. I can run the steps 1-4 several times and it works just fine.
a. If Emacs does NOT hang, I can probably (with 20% probability) repeat the steps without any problem.
b. If Emacs DOES hang, I can kill Emacs, and then (with 20% probability) repeat the steps without any problem.
c. If Emacs DOES hang, and I do NOT kill the hanging Emacs, then opening ANOTHER Emacs session and repeating steps 1-4 WILL hang with 100% probability.  (This will make it much easier to try things for debugging!).
 
7. The "Forming completions" is inside a loop in the function `ess-get-modtime-list` here:
https://github.com/emacs-ess/ESS/blob/cb75588d972b97e597c42108b3c83c2f67947cac/lisp/ess-inf.el
8.  Looking at `ess-get-modtime-list` I see this bit of code
```
                               (prog2
                                       (message "Forming completions for %s..." pack)
                                       (ess-object-names pack index)
                                     (message "Forming completions for %s...done" pack))))))
```
In the `*Messages*` buffer I can see this (when there is no hang):
```
Forming completions for .GlobalEnv...done
Forming completions for ESSR...done
Forming completions for package:stats...done
Forming completions for package:graphics...done
Forming completions for package:grDevices...done
Forming completions for package:utils...done
Forming completions for package:datasets...done
Forming completions for package:httr...done
Forming completions for package:devtools...done
Forming completions for package:usethis...done
Forming completions for .startup...done
Forming completions for package:methods...done
Forming completions for Autoloads...done
Forming completions for package:base...done
Retrieving RDS aliases...done
```
Notice the "done" at the end of each line.  When there IS a hang, the minibuffer shows only
```
Forming completions for package:grDevices"
```
Notice there is no "done" printed when there is a hang. Looking again at `ess-get-modtime-list`, perhaps the problem lies with `ess-object-names`.  When I look at the code for that function, there are 3 FIXMEs.  

In the *ESS* dribble buffer is this (without hang).
```
(R): ess-dialect=R, buf=*Ibuffer*, start-arg=nil
 current-prefix-arg=nil
(inferior-ess: waiting for process to start (before hook)
(inferior-ess 3): waiting for process after hook(R): inferior-ess-language-start=options(STERM='iESS', str.dendrogram.last="'", editor='emacsclient', show.error.locations=TRUE)
(get-object-list R) ..--> (ess-get-modtime-list)
 (ess-search-list ... ) after 'search()
', point-max=306
(nil): created new alist of length 14
 (length alist) : 14
 have re-read pos=1: -> length 0
 (ess-search-list ... ) after 'search()
', point-max=306
Processing RDS files ...
 |--> nil [searching t]
```
If there is a hang, then "point-max=306" is the last thing shown in the dribble buffer.

lionel-:
Builds on #1063.

1. Accumulate `ess-command` output in the output buffer to avoid prompt or sentinel getting truncated by chunking.
2. Use sentinel to signal end of evaluation. This way things looking like final prompts won't cause issues.

This should make it more stable. Hopefully (1) fixes #1043.

### DIFF
--- a/etc/ESSR/R/.basic.R
+++ b/etc/ESSR/R/.basic.R
@@ -133,7 +133,9 @@ if(.ess.Rversion < "1.8")
         unquote(substitute(expr))
     }
 
-.ess.command <- function(expr) {
+.ess.command <- function(expr, sentinel) {
+    on.exit(writeLines(sentinel))
+
     out <- withVisible(expr)
 
     # Print result manually because we can't rely on auto-print
--- a/etc/ESSR/R/.basic.R
+++ b/etc/ESSR/R/.basic.R
@@ -138,12 +138,11 @@ if(.ess.Rversion < "1.8")
 
     out <- withVisible(expr)
 
-    # Print result manually because we can't rely on auto-print
-    # without changing the last value
-    if (out$visible) {
+    ## Print result manually because we can't rely on auto-print
+    ## without changing the last value
+    if (out$visible)
         print(out$value)
-    }
 
-    # Keep `.Last.value` stable
+    ## Keep `.Last.value` stable
     invisible(.Last.value)
 }
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -407,6 +407,18 @@ Return non-nil if the process is in a ready (not busy) state."
                    (string-match (concat "\\(" inferior-ess-secondary-prompt "\\)\\'") string)))
     ready))
 
+(defun inferior-ess--set-status-sentinel (proc output-buf sentinel)
+  (with-current-buffer output-buf
+    (save-excursion
+      ;; The only assumption is that the prompt finishes with "> "
+      (goto-char (- (point-max) 2))
+      (when (looking-at inferior-ess-primary-prompt)
+        (forward-line -1)
+        (when (looking-at (concat sentinel "\n"))
+          (delete-region (match-beginning 0) (match-end 0))
+          (process-put proc 'busy nil)
+          (process-put proc 'ess-output-sentinel nil))))))
+
 (defun inferior-ess-mark-as-busy (proc)
   "Put PROC's busy value to t."
   (process-put proc 'busy t)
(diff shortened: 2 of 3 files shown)


etc/ESSR/R/.basic.R
```diff
@@ -133,7 +133,9 @@ if(.ess.Rversion < "1.8")
         unquote(substitute(expr))
     }
 
-.ess.command <- function(expr) {
+.ess.command <- function(expr, sentinel) {
+    on.exit(writeLines(sentinel))
```
lionel-, line 137: - Should check that no sink is active. Temporarily divert to stdout for writing the sentinel if there's a sink? Note that the sink might be set up by the user before the background command is run. Currently this causes a hang because the sentinel gets sunk.

  Actually we might want to check for active sinks before the background command runs and consider the REPL busy if there is any. The ESS command won't be able to get output anyway, and the user will get parasite output from the background command in their diversion file.

- Would be nice to communicate unexpected exits and bubble this up as errors on the Emacs side. We could use a boolean flag and flip it after evaluation of `expr`, this would be robust (e.g. handle `try(silent = FALSE)` or even restart invokations).

mmaechler: Thank you very, @lionel- ,  for addressing this issue.  As you (all) know, I've been unhappy quite a bit with "ESS fiddling with R behind my back",  and indeed I had noticed  that `sink()`, `capture.output()`, etc have occasionally behaved very badly/randomly/strangely.   Currently, I don't have the time to delve into this (including prompt detection etc) at all,  but eventually it would be quite important and helpful for ESS development if you could write up / document what "we" are doing currently, and why (and where in the emacs lisp sources if that's not obvious)...

lionel-: By the way I now wonder if the extra newlines you've been observing in some contexts are due to background tasks calling `inferior-ess-available-p`. This routine checks whether a process has been marked busy (i.e. there's no final prompt) and also sends a new line to double check. I haven't investigated this yet.

> eventually it would be quite important and helpful for ESS development if you could write up / document what "we" are doing currently, and why

That would be useful eventually, perhaps as part of the switch to the async evaluation queue discussed in #1017. I'm only starting to get a complete picture of what is going on. It will be important to establish what are the safe assumptions a routine can make in different contexts.

Issue #1067, opened 2020-12-01
R process shown in two windows instead of one


tmalsburg:
When I open an `.R` file and execute

```
  M-x ess-eval-region-or-line-visibly-and-step
```

on a line of code when there’s no R process yet, the current window is split and a new R process is started.

Problem: The R process is shown in both windows, not just in the new window.

Environment:
- Emacs version: current master branch with native-comp.
- ESS installed from MELPA.
- Tested with `emacs -q`

mmaechler:
Yes, thank you.  This behavior has started a while ago, but definitely this year ...

tmalsburg:
I had a look.  I think the issue is in `ess-eval-region-or-line-visibly-and-step` where `ess-force-current-buffer` opens a new window and then the subsequent `display-buffer` also switches to the R process in yet another window.  There's an `inhibit-same-window`, but at this point we're in the process window and the `inhibit-same-window` probably doesn't do what it was likely intended to do.

My solution in #1068 solves my immediate problem, but I'm not sure it's the best solution.  It may be better to remove the `inhibit-same-window`. I was also wondering whether `ess-force-current-buffer` actually needs to create a new window.  It's documentation says that it makes sure there is a process.  New windows are not mentioned.  However, `ess-force-current-buffer` is used in many places and some may actually rely on the new window, so I don't feel confident changing anything there.

tmalsburg:
Just checked, removing the `inhibit-same-window` works as well.  Let me know what you prefer and I'll update the PR as needed.

tmalsburg:
The commit that likely introduced the issue is b2778ed23bc765c3370bb33dce8a9a458e3581f7.  From eye-balling it, i get the impression that it may have introduce this issue in a few other places as well.

tmalsburg:
Hm, I think this commit may actually be the culprit: 24da603184ce39246611dd5b8602e769d7ebd5bf  Or some other commit made in this function around that time.  At least this is the code that is responsible for `ess-force-current-buffer` showing the process window. And the time of this commit is more in line with my memory of when this issue first appeared.

tmalsburg:
Last comment for today: I couldn't nail down the root cause but I think the issue is that `ess-force-current-buffer` creates a new window when it shouldn't. There is code from two months ago which aimed at preventing this (21cb413184765ca86138b69f8a3e806a4421bdfe) but it appears not to be working or it misses some cases.  Perhaps @lionel-, author of the commit in question, can help here.

[...]

PR #1068, opened 2020-12-01
Prevent new R process from being shown in two windows

tmalsburg:
Affects `ess-eval-region-or-line-visibly-and-step'.  See #1067.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1616,7 +1616,7 @@ Note that when inside a package and namespaced evaluation is in
 place (see `ess-r-set-evaluation-env') evaluation of multiline
 input will fail."
   (interactive)
-  (ess-force-buffer-current)
+  (save-window-excursion (ess-force-buffer-current))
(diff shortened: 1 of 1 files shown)


lionel-:
Before making more changes to this part of the code I think we need to set up unit tests and define expected behaviour systematically in relation to `display-buffer`, its possible configurations, and the possible call sites.

I've been meaning to do that as part of https://github.com/emacs-ess/ESS/labels/process%3Awindows. If you'd like to look into unit tests, I think (but I'm not sure) we'll need to have a parallel set of tests that are run in interactive rather than batch mode because the window/frame system is not fully available in batch mode. `evil-mode` uses such a setup for their tests: https://github.com/emacs-evil/evil/blob/d6cf6680/evil-tests.el#L119-L123.

tmalsburg:
I wish I could work on these unit tests, but I've already spent more time on this than I can afford at this time.  Sorry.

tmalsburg:
I'm closing this PR as it's almost certainly not the desired solution.

PR #1076, opened 2020-12-18
Add `ess-toggle-trace`

lionel-:
To help our users provide useful debugging information.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,6 +4,11 @@
 Changes and New Features in 19.04 (unreleased):
 @itemize @bullet
 
+@item ESS: New @code{ess-toggle-trace} function.
+Call it once to start tracing all ESS functions with
+@code{trace-function}. Call it again to stop tracing.
+Tracing is useful for debugging background ESS behaviour.
+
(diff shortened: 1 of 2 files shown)


jabranham:
Interesting! I think (in no particular order):

1. We should have either this or the dribble calls, not both
2. It should be a minor mode perhaps?
3. We shouldn't promise "all" ess functions when it won't trace the many functions that don't start with ess-

jabranham:
4. Perhaps this should go in Emacs upstream to trace any given prefix

lionel-:
Thanks for your comment Alex. I have clarified the behaviour regarding prefixed tracing, and added a few non-prefixed functions to the tracing set. I have also switched to a minor mode. An easy way to make this generic to all of Emacs would be to provide a buffer-local variable pointing to a function returning the function names to trace.

Regarding the dribble buffer, I think it's a bit early to rush into removing it. Certainly we can remove some dribble calls now that we have tracing, but the dribble buffer may still be useful to log more structured information about ESS state. Also it's less verbose and less intrusive than tracing and so dribble logging can be kept enabled during a work session, which is not possible for tracing.

Issue #1025, opened 2020-07-13
Emacs hangs when opening help (re-open #954)


hendrikweisser:
I'm experiencing what seems to be the same issue as reported in #954. That has been closed as resolved, but I believe should be re-opened. There are two comments on the issue reporting that the problem persists:
https://github.com/emacs-ess/ESS/issues/954#issuecomment-572740250
https://github.com/emacs-ess/ESS/issues/954#issuecomment-655499551

jabranham:
Thanks. I know what the issue is, should be able to fix it next week. If
it's bothering you until then you can disable adding links in help pages as
a workaround. I think the variable is named something like
ess-help-add-links

hendrikweisser:
@jabranham: That sounds great! I looked for a variable named `ess-help-add-links` or similar, but I only found a function (`ess-r-help-add-links`) and I don't know how I would disable that.

jabranham:
Sorry, I thought there was a variable to control that behavior but apparently I was mistaken. Just redefine ess-r-help-add-links to return nil:

```
(defun ess-r-help-add-links () nil)
```

jabranham:
Just pushed something that hopefully fixes this issue - please test it out and let me know if you still have this problem.

hendrikweisser:
@jabranham: Took me a while to get around to testing it, but it appears that the problem is indeed solved (using ESS version 20200825.829 from Melpa). Thank you very much!

r2evans:
Reopen?

ESS-20201208.1047, hanging again with `?data.table::CJ`, same symptoms as before.

r2evans:
FYI, using 20201221.1551 (on emacs-26.3), while it no longer crashes, it now doesn't open any help *at all* (first first time). Loosely demonstrated:

```
|----------------------------------------------------------------||------------------------------------------------------------------------------|
| > library(data.table)                                          ||                                                                              |
| > ?CJ                                                          ||   Retrieving RDS aliases...                                                  |
|                                                                ||   ess-command: Timeout during background ESS command ’.ess.getHelpAliases()’ |
|                                                                ||                                                                              |
| > ?ls                                                          ||   Retrieving RDS aliases...                                                  |
|                                                                ||   ess-command: Timeout during background ESS command ’.ess.getHelpAliases()’ |
|                                                                ||                                                                              |
| > ?ls                                                          ||   (success)                                                                  |
|----------------------------------------------------------------||------------------------------------------------------------------------------|
[...]

PR #1080, opened 2020-12-22
Cache aliases on the R side

lionel-:
These aliases are currently all read again when the search path has changed. The number of RDS files to read grows with the numbers of packages attached and @r2evans has reported in #1025 that it takes between 5 to 10 seconds to get help.

To limit the delays, the aliases are cached per package on the R side. ~The cache is reset when `sp-for-help-changed` is set on the inferior.~


lionel-:
> The cache is reset when sp-for-help-changed is set on the inferior.

oops this defeats the purpose. We only need to reset the cache when the user has requested it with a prefix arg, but not everytime a change to the search path is detected. Fixed in last commit.

PR #1099, opened 2021-01-24
WIP: Determine links from installed Rd AST

### LINKED ISSUE #1095: ESS hangs when opening R help
I just upgraded to ESS version 20210117.1706. Having loaded {mgcv} via `library('mgcv')`, when I type `?gam` in a running R buffer I observe:

* the help page is not displayed (it is opened in a new buffer that I can switch to using the Buffer menu, but that buffer is no longer raised to be the visible buffer)
* I see the following in the Messages buffer:

        Retrieving RDS aliases...done
        ess-command: Timeout during background ESS command ‘names(formals(gam(formula,family=gaussian(),data=list))’

* the R prompt is not displayed (you can hit return in the R console buffer and the prompt displays).

This is with Emacs 27.1 (release 2.fc33 if that matters) on Fedora 33 with ESS version 20210117.1706. The version of {mgcv} is 1.8.33 (the latest CRAN release).

### LINKED ISSUE #1098: Issues caused by linkification in help buffers
I did some research on the help issues we have been facing for a while. They are caused by a new feature that adds button links to keywords in `ess-help` buffers.

Here is the chronology:

- **PR**: Add links to help buffers https://github.com/emacs-ess/ESS/pull/748
- **PR**: Only insert help links that link somewhere https://github.com/emacs-ess/ESS/pull/772
- **Issue**: False-positive links in help buffers for functions documented in the current topic https://github.com/emacs-ess/ESS/issues/847
- **PR**: Don't link functions documented in the current topic https://github.com/emacs-ess/ESS/pull/848
- **Issue**: False-positive links for documented arguments like `data` https://github.com/emacs-ess/ESS/issues/855
- **PR**: Don't link arguments from functions documented in the current topic https://github.com/emacs-ess/ESS/pull/891

As implemented, linkification works by looking up "fancy quotes" in the help buffer, i.e. quotes that look like this:

```
‘rename()’
```

These fancy quotes are the result of `\code{}` markup in .Rd files. The first PR replaces all these quoted elements by Emacs buttons. Clicking on the button attempts to look up the help for the quoted text. Since there are many `\code{}` elements in the help page for which there is no help available, the linkification was excessive and distracting. The next three PRs filter the quoted elements to only retain those:

- For which a help topic is available. Uses `ess-help-get-topics` to retrieve the list of topics.
- And which don't correspond to a function documented in the current topic or one of their formal arguments.

Linkification is a nice new feature that makes it explicit that a word in the help page has contextual help available. Unfortunately it has several problems:

1. As Martin mentioned, ideally we'd use the parsed Rd files to determine the links, rather than figuring out by ourselves where to add them.

2. The use of `ess-help-get-topics` causes very long hangs on Windows. The more packages attached on the search path, the longer the hangs; @r2evans reported hangs over 5 seconds. I have improved the caching of these topics on the R side in #1080 but this doesn't fix the hangs completely.

3. The current approach to determining the functions and arguments documented is brittle. We regularly send incomplete inputs to R via `ess-command`, which causes infinite hangs (see #1025 and #1095). I have added a timeout to `ess-command` in #1078 to limit the issue and give a friendly error message but we need to fix the hangs in the first place. @r2evans kindly contributed a fix in #1083 but it turns out it doesn't completely solve the issue.

4. The filtering of functions and arguments documented in the current topic currently doesn't work correctly if the relevant functions are not reachable from the global environment. This would be the case with namespace-qualified requests like `?foo::bar`, or when multiple packages export the same function.

(3) can be fixed robustly and I have already started to work on a fix. However, now that I have a better overview of these problems, I wonder if we should just revert the linkification for this release and work on a better approach for the next one. I worry that problems (2) and (4) are going to cause a subpar experience for many users, so I tend to think we should revert. We could turn off that feature by default and keep it in ESS until (1) is solved. This way it would still be available for those that want to use the bleeding edge feature.

What do you think @mmaechler and @jabranham?

lionel-:
Closes #1098.
Closes #1095.

Uses a few internal `utils:::` helpers in the implementation. This should still be robust as a `tryCatch()` swallows any error or warning.

The link targets are retrieved from the topic's parsed Rd. If fancy-quoted words match these targets, we transform these to buttons. This still isn't perfect, some links will be missing and namespaced links are not disambiguated - ESS will prompt for a location instead of linking to the right place. However, this fixes the major problems and we get to keep this nice feature implemented by Alex.

To support this, we now keep track of some metadata in the help buffer: object name and package name.


lionel-:
I'll test this locally for a while.

@mmaechler Can you please review the R side and see if the approach and implementation are reasonable?

mmaechler:
Congratulations, @lionel- , for taking the "deep dive"!     I've looked at only the R-code.  It does seem "reasonable" .. and it agrees with my gut feeling that we couldn't have gone there, without using hidden and hence non-API properties of the current Rd parse implementation.  I've wanted to think / discuss about exporting some of the Rdparse functionality, in the past but never got to that yet, anyway...

lionel-:
Thank you Martin.

I browsed a few help pages from base and stats this morning. It looks good overall, but there are some false positives when a help page includes a link to a function that has the same name as an argument. This would be the case with `df` in `?qt` for example.

So I think we'll still need to reintroduce some filtering as Alex did. This should be easy now that we have access to the Rd AST, we can just look up what are the documented arguments. After the filtering `df` will become a false negative in the `seealso` section: it won't be linkified even though it should. As discussed before it is preferable to have a few false negatives than a lot of false positives.

mmaechler:
agreed.  I'm happy if you merge ((I'm not happy working with ESS branches; the master is variable enough already ..))

lionel-:
Now also preserves trailing parentheses in link buttons.

vspinu:
I am late to the party but this seems great! Great job!!

I have only ne question, does this work on remotes? I see that there are some local package path manipulation in the code.

lionel-:
> I have only ne question, does this work on remotes? I see that there are some local package path manipulation in the code.

Good question. I think it will because we're only inspecting the path where the package is installed to figure out the package name. The directory structure of installed packages should be the same everywhere. I haven't tested on remotes though as I don't use those. Would be great to have some tests for remotes but currently remote ESSR loading takes like 20 seconds on my machine. I tried to investigate why that's the case but maybe it's just Emacs IPC that's really slow (maybe only on some platforms).

Issue #1104, opened 2021-02-01
Timeout during background ESS command ?


vspinu:
When I do `C-c C-d C-e` on a relatively large object I am getting `Timeout during background ESS command` (implemented in ee0266e03c85)

I don't think 1s global default timeout is a good idea. "instantaneous" is a relative concept. It can easily span > 1 s on remote machines and it almost always spans > 1s for `C-c C-d C-e`

I think the default should be 30s. C-g should do the job and it's up to the developer to decide whether to use shorter delays.  For eldoc completion, like the cause of #1062, it can be 1s indeed, but then some calls will never be cached, so no completion. Not ideal either. 

<details>

```
Debugger entered--Lisp error: (error #("Timeout during background ESS command ‘str(jsons, ..." 43 48 (tb-index 10 fontified t)))
  signal(error (#("Timeout during background ESS command ‘str(jsons, ..." 43 48 (tb-index 10 fontified t))))
  error("Timeout during background ESS command `%s'" #("str(jsons, list.len=1000)" 4 9 (fontified t tb-index 10)))
  (if (ess-wait-for-process proc nil wait force-redisplay timeout) nil (error "Timeout during background ESS command `%s'" (ess--strip-final-newlines cmd)))
  (if no-prompt-check (sleep-for 0.02) (if (ess-wait-for-process proc nil wait force-redisplay timeout) nil (error "Timeout during background ESS command `%s'" (ess--strip-final-newlines cmd))) (goto-char (point-max)) (delete-region (point-at-bol) (point-max)))
  (save-current-buffer (set-buffer out-buffer) (ess-setq-vars-local proc-alist) (setq buffer-read-only nil) (erase-buffer) (set-marker (process-mark proc) (point-min)) (inferior-ess-mark-as-busy proc) (process-send-string proc rich-cmd) (if no-prompt-check (sleep-for 0.02) (if (ess-wait-for-process proc nil wait force-redisplay timeout) nil (error "Timeout during background ESS command `%s'" (ess--strip-final-newlines cmd))) (goto-char (point-max)) (delete-region (point-at-bol) (point-max))))
[...]

Issue #1102, opened 2021-01-29
Current buffer has no process


jthaman:
On Windows 10 / Emacs 27.1, 26.3 / ESS 20210126, I occasionally edit files on network drives, and run into some odd ESS behavior. It's hard to pin down the issue, but after working interactively with R scripts on network drives, the iESS buffer loses connection, and shows `no process` in the modeline. 

In the messages buffer, I can only find the phase `user-error: Current buffer has no process`. If I try to continue to interact with the broken process, I get the message`Error running timer ‘ess--idle-timer-function’: (wrong-type-argument stringp nil)`

When this happens I have to restart the R process, but I inevitably run into it again.

The *ESS* buffer shows the output

```
(R): ess-dialect=R, buf=JTH/v3_data_read.R, start-arg=nil
 current-prefix-arg=nil
(inferior-ess: waiting for process to start (before hook)
(inferior-ess 3): waiting for process after hook(R): inferior-ess-language-start=options(STERM='iESS', str.dendrogram.last="'", editor='emacsclient', show.error.locations=TRUE)
 (ess-search-list ... ) after 'search()
', point-max=294
(nil): created new alist of length 13
```

I've only run into this `no process` issue while working interactively with R scripts on network drives under Windows 10, but I'm not sure that's part of the problem.

Here is my complete ESS configuration: 

```
(use-package ess
  :defer t
  :ensure t
  :hook
  (ess-help-mode . evil-normal-state)
  :custom
  (inferior-ess-fix-misaligned-output t)
  (ess-eldoc-show-on-symbol t)
  (ess-gen-proc-buffer-name-function 'ess-gen-proc-buffer-name:projectile-or-directory)
  (ess-eval-visibly 'nil "Don't hog Emacs")
  (ess-style 'RStudio)
  (ess-use-flymake nil "Syntax checking is usually not helpful")
  (ess-tab-complete-in-script nil "Do not interfere with Company")
  (ess-use-ido nil "Prefer Ivy/Counsel")
  (ess-history-directory (expand-file-name "ESS-history/" no-littering-var-directory))
  (inferior-R-args "--no-save")
  (ess-R-font-lock-keywords
   (quote
    ((ess-R-fl-keyword:keywords . t)
     (ess-R-fl-keyword:constants . t)
     (ess-R-fl-keyword:modifiers . t)
     (ess-R-fl-keyword:fun-defs . t)
     (ess-R-fl-keyword:assign-ops . t)
[...]

PR #1106, opened 2021-02-02
Attempt at fixing process loss

lionel-:
From @vspinu in #1102:

> I think I know where this comes from. On the timeout-interrupt (#1104) the process is somehow stuck in the temporary buffer and not returned back to the original buffer. It doesn't happen always, only sometimes. This is why it's so difficult to debug.

With this patch we now nest two `unwind-protect` to ensure the process cleanup is run even when the interrupt cleanup exits early.

@vspinu Since you could reproduce the issue, would you mind trying this branch with a small default timeout to see if it actually fixes the problem?

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1293,43 +1293,44 @@ wrapping the code into:
         ;; Swap the process buffer with the output buffer before
         ;; sending the command
         (unwind-protect
-            (progn
-              (when use-sentinel
-                (process-put proc 'ess-output-sentinel sentinel))
-              (set-process-buffer proc out-buffer)
-              (set-process-filter proc 'inferior-ess-ordinary-filter)
-              (with-current-buffer out-buffer
-                (ess-setq-vars-local proc-alist)
-                (setq buffer-read-only nil)
-                (erase-buffer)
-                (set-marker (process-mark proc) (point-min))
-                (inferior-ess-mark-as-busy proc)
-                (process-send-string proc rich-cmd)
-                ;; Need time for ess-create-object-name-db on PC
-                (if no-prompt-check
-                    (sleep-for 0.02)   ; 0.1 is noticeable!
-                  (unless (ess-wait-for-process proc nil wait force-redisplay timeout)
-                    (error "Timeout during background ESS command `%s'"
-                           (ess--strip-final-newlines cmd)))
-                  ;; Remove prompt. If output is cat(..)ed without a
-                  ;; final newline, this deletes the last line of output.
+            (unwind-protect
+                (progn
+                  (when use-sentinel
+                    (process-put proc 'ess-output-sentinel sentinel))
+                  (set-process-buffer proc out-buffer)
+                  (set-process-filter proc 'inferior-ess-ordinary-filter)
+                  (with-current-buffer out-buffer
+                    (ess-setq-vars-local proc-alist)
+                    (setq buffer-read-only nil)
+                    (erase-buffer)
+                    (set-marker (process-mark proc) (point-min))
+                    (inferior-ess-mark-as-busy proc)
+                    (process-send-string proc rich-cmd)
+                    ;; Need time for ess-create-object-name-db on PC
+                    (if no-prompt-check
+                        (sleep-for 0.02) ; 0.1 is noticeable!
+                      (unless (ess-wait-for-process proc nil wait force-redisplay timeout)
+                        (error "Timeout during background ESS command `%s'"
+                               (ess--strip-final-newlines cmd)))
+                      ;; Remove prompt. If output is cat(..)ed without a
+                      ;; final newline, this deletes the last line of output.
+                      (goto-char (point-max))
+                      (delete-region (point-at-bol) (point-max))))
+                  (setq early-exit nil))
+              (when early-exit
+                (with-current-buffer out-buffer
+                  (goto-char (point-min))
+                  (when (and use-sentinel
+                             (not (re-search-forward
+                                   (inferior-ess--sentinel-start-re sentinel)
+                                   nil t)))
+                    ;; CMD probably failed to parse if the start sentinel
+                    ;; can't be found in the output. Disable the sentinel
+                    ;; before interrupt to avoid a freeze.
+                    (process-put proc 'ess-output-sentinel nil))
                   (goto-char (point-max))
-                  (delete-region (point-at-bol) (point-max))))
-              (setq early-exit nil))
+                  (ess-interrupt))))
           ;; Restore the process buffer in its previous state
-          (when early-exit
-            (with-current-buffer out-buffer
-              (goto-char (point-min))
-              (when (and use-sentinel
-                         (not (re-search-forward
-                               (inferior-ess--sentinel-start-re sentinel)
-                               nil t)))
-                ;; CMD probably failed to parse if the start sentinel
-                ;; can't be found in the output. Disable the sentinel
-                ;; before interrupt to avoid a freeze.
-                (process-put proc 'ess-output-sentinel nil))
-              (goto-char (point-max))
-              (ess-interrupt)))
           (process-put proc 'ess-output-sentinel nil)
           (set-process-buffer proc oldpb)
           (set-process-filter proc oldpf)
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1336,7 +1336,7 @@ wrapping the code into:
                     ;; before interrupt to avoid a freeze.
                     (process-put proc 'ess-output-sentinel nil))
                   (goto-char (point-max))
-                  (ess-interrupt)))
+                  (ess--interrupt proc 1)))
             ;; Restore the process buffer in its previous state
             (process-put proc 'ess-output-sentinel nil)
             (set-process-buffer proc oldpb)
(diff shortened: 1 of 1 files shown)


lisp/ess-inf.el
```diff
@@ -1293,43 +1293,44 @@ wrapping the code into:
...
+                (with-current-buffer out-buffer
+                  (goto-char (point-min))
+                  (when (and use-sentinel
+                             (not (re-search-forward
+                                   (inferior-ess--sentinel-start-re sentinel)
+                                   nil t)))
+                    ;; CMD probably failed to parse if the start sentinel
+                    ;; can't be found in the output. Disable the sentinel
+                    ;; before interrupt to avoid a freeze.
+                    (process-put proc 'ess-output-sentinel nil))
```
vspinu, line 1330: proc variables need not actually be namespaced. In the entire codebase they are not.

lisp/ess-inf.el
```diff
@@ -1293,43 +1293,44 @@ wrapping the code into:
...
+                    (setq buffer-read-only nil)
+                    (erase-buffer)
+                    (set-marker (process-mark proc) (point-min))
+                    (inferior-ess-mark-as-busy proc)
+                    (process-send-string proc rich-cmd)
+                    ;; Need time for ess-create-object-name-db on PC
+                    (if no-prompt-check
+                        (sleep-for 0.02) ; 0.1 is noticeable!
+                      (unless (ess-wait-for-process proc nil wait force-redisplay timeout)
+                        (error "Timeout during background ESS command `%s'"
```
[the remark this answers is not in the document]
lionel-: I think it's a good idea to increase the timeout but to me it makes sense to throw when the R process doesn't respond. If we can't interrupt when requested it doesn't seem safe to continue normal execution, the calling context of an `ess-command` is likely to assume it can interact with the process.

lisp/ess-inf.el
```diff
@@ -1293,43 +1293,44 @@ wrapping the code into:
...
+                                   (inferior-ess--sentinel-start-re sentinel)
+                                   nil t)))
+                    ;; CMD probably failed to parse if the start sentinel
+                    ;; can't be found in the output. Disable the sentinel
+                    ;; before interrupt to avoid a freeze.
+                    (process-put proc 'ess-output-sentinel nil))
                   (goto-char (point-max))
-                  (delete-region (point-at-bol) (point-max))))
-              (setq early-exit nil))
+                  (ess-interrupt))))
```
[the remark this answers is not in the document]
lionel-: We use `ess-interrupt` so we block until prompt shows up. If we used `interrupt-process` we wouldn't block and output would spill in the process buffer.

lisp/ess-inf.el
lionel-: I noticed they were not but I think we should start namespacing them if ESS really intends to be generic.

vspinu: It's our own process, so we ca default to non-namespaced symbols. If other tooling use our process plist, they have to namespace.

lionel-: They are not necessarily our own process since we have `ess-remote`.

lisp/ess-inf.el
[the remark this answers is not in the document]
vspinu: I am not saying we shouldn't interrupt. I am saying we can avoid throwing. We are erroring there just to catch it two line latter by ourselves. The only reason to do that is to send the message to the user.

lisp/ess-inf.el
```diff
@@ -1293,43 +1293,44 @@ wrapping the code into:
...
-                ;; Need time for ess-create-object-name-db on PC
-                (if no-prompt-check
-                    (sleep-for 0.02)   ; 0.1 is noticeable!
-                  (unless (ess-wait-for-process proc nil wait force-redisplay timeout)
-                    (error "Timeout during background ESS command `%s'"
-                           (ess--strip-final-newlines cmd)))
-                  ;; Remove prompt. If output is cat(..)ed without a
-                  ;; final newline, this deletes the last line of output.
+            (unwind-protect
+                (progn
```
vspinu, line 1297: How about adding unwind-protect just on the sentinel+ess-interrupt block instead? Seems cleaner and easier to reason about to me.

lisp/ess-inf.el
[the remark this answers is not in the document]
lionel-: Timeouts are not a code smell to me, they are essential to concurrency with external entities. I really think we need to block with a timeout. The user may have already typed `C-g` and if ESS doesn't respond to the first interrupt consistently, sometimes requiring multiple interrupts, this deteriorates the user sense of being in control of the process.

I think using `ess-interrupt` rather than `interrupt-process` is cleaner from the standpoint of code reuse consolidating internals.

I don't know exactly what you're referring to with the `"\n"` but if it's tangential I'd prefer to focus on the main points.

lisp/ess-inf.el
lionel-: IIUC that was how I initially wrote it but something fell off using unwind-protect in the protect block. I tried to google if it was a normal pattern but couldnt find anything. Happy to change it back if you think it's better.

lisp/ess-inf.el
vspinu: ess-remote hijacks the process. There is no way back afterwards. 

Anyways, a tiny issue. I would keep small keywords for simplicity.

lionel-:
@vspinu Can you try this branch and see if it fixes the process losses for you please?

lisp/ess-inf.el
lionel-: There could also be minor modes interacting with the process? I agree it's a tiny issue and I don't mind removing the prefix.

lisp/ess-inf.el
vspinu: I think it's cleaner as it can be nested into `(when (and use-sentinel early-exit) ...)` block. We don't really need `unwind-protect` there, simple `condition-case-unless-debug` would do, no?

lisp/ess-inf.el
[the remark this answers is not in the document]
lionel-: oops I thought we were discussing about the timeout error inside `ess-interrupt` but you're suggesting we don't timeout-error in `ess-command` altogether?

I strongly think a timeout in `ess-command` should abort everything. We had a lot of hanging issues lately (most of them now solved) and having a clean error improved ux _a lot_.

> The only reason to do that is to send the message to the user.

It is also to interrupt the caller of `ess-command`. If the command failed, that code should be aborted. Why should it continue if the command could not be run to completion?

Also in the future we'll bubble R errors to emacs errors here.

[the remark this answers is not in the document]
lionel-: > just to catch it two line latter by ourselves

We don't catch the error by the way, we let it go through.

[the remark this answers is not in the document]
vspinu: Arh, right. I missed that part. I guess I am getting elisp-rusty a bit :/

lisp/ess-inf.el
vspinu: > There could also be minor modes interacting with the process?
Maybe, but those will use their own prefixes. 

>  I agree it's a tiny issue and I don't mind removing the prefix.

Well, the main thing is to keep it consistent. Either we keep all un-prefixed, or all prefixed. Changing now into prefixed seems an overkill of the work for little gain (little loss, actually, because it's uglier :).  Also who knows, maybe some dude/external pakage is using 'busy marker for something in their code. So I am afraid we are stuck with unprefixed keywords.

lisp/ess-inf.el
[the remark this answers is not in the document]
vspinu: It's not tangential. Look at the `ess-interrupt` code. 

> I think using ess-interrupt rather than interrupt-process is cleaner from the standpoint of code reuse and consolidating internals.

Sometimes using lower level utilities in critical code is cleaner and beter conveys the meaning. `ess-interrupt` in this case does a bunch of things that are intended for interactive use. There is no place for things like `(inferior-ess-force)`, `(ess-get-process)` `(timeout 1)` `(process-send-string nil "\n")` `(goto-char (process-mark proc))` in `ess-command` IMO.

lisp/ess-inf.el
lionel-: s/dude/person. Actually the use of prefixes would allow us to use `ess--` for private internal state. Let's just consider that all these variables are private anyway. I'll rename to `output-sentinel` after we merge or close the PR.

vspinu:
I will try but I cannot replicate things reliably, so it may take a while till I stumble on an issue. 

Regarding "sentinel" in the names. Can we call it ess-command-output-delimiter or something? I don't like "sentinel", too generic, doesn't convey the meaning precisely and can be confused with `process-sentinel`, `ess-process-sentinel`.

lionel-:
Thanks! "Output delimiter" sounds good to me.

lisp/ess-inf.el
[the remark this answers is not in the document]
lionel-: Are you sure that `"\n"` is not needed on Windows terminals? Since we are blocking for prompt it seems risky not to send it? While I see your point, I think my point holds as well, considering the uncertainty about cross-platform behaviour.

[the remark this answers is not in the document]
lionel-: I have refactored `ess-interrupt` into interactive and non-interactive functions. `ess-command` now uses the non-interactive one. I kept the newline thing in the core routine because of the uncertainty regarding its purpose.

PR #1107, opened 2021-02-03
Rewrite load ESSR into remote process routine

### LINKED ISSUE #1047: company object completion not working on remotes
While runing R on a remote server using TRAMP the following error appears in a new help[R] buffer every time I try to open the help file on some function:

```
+ Error in base::as.environment("ESSR") : 
  no item called "ESSR" on the search list
```

The problem just started after updating to ESS 20200905.822 and only when using TRAMP.

The error appears whether I type `?function_name` or I call `ess-display-help-on-object` with my keybinding.

I am using Spacemacs.

dapritchard:
This PR is an attempt to fix #1047. It is still a WIP so as to add some testing and perhaps better error handling.

The new implementation as it currently stands copies the ESSR files over each time that a new R session is started on the remote. This seems kind of silly, however I think you would want to check each time that you have the current version of the files on the remote, and I'm not sure that it's worth the effort right now since simply copying the files may be fast enough.


vspinu:
This will work only with TRAMP connections. We also have ess-remote, which need not rely on tramp at all. 

> however I think you would want to check each time that you have the current version

If we go this way then we could consider keeping ESSR.rds as part of the repo, and sending that file only. Checking for version on the remote is trivial. It's already implemented in the current approach. 

Not sure though about the entire initiative. It's feels like a marginal improvement and potentially opens a can of issues with the copying of all those files to the remote.

vspinu:
For completeness, one approach which hasn't been attempted yet is to send RDS through stdout. It's about 12kb, very small. One can split it in short strings and send them one by one to the remote which will check the digest and piece them together. If this work it would be an uniform solution for all our problems.

lionel-:
I like that approach, but I thought we'd need some sort of encoding to send a binary file over the process connection, e.g. base64? But then we need something on the other end to decode it.

dapritchard:
What if we inject the ESSR objects into the global environment on the remote process in the manner that is being currently performed, and then simply move them into an `ESSR` environment after the fact? This would eliminate the need to try and copy files into the remote filesystem and then read them into R.

So running something like the following code in R once the ESSR objects have been created in the global environment
```r
local({
    ESSR <- new.env()
    essr_nms <- grep("\\.(ess|ESS)", ls(.GlobalEnv, all.names = TRUE), value = TRUE)
    for (nm in essr_nms) {
        ESSR[[nm]] <- get(nm, pos = .GlobalEnv)
    }
    rm(list = essr_nms, pos = .GlobalEnv)
    attach(ESSR)
})
```
but taking a little more care to make sure that everything is back-compatible with older versions of R such as is done in `.load.R`.

vspinu:
> I like that approach, but I thought we'd need some sort of encoding to send a binary file over the process connection, e.g. base64? 

Precisely. This where I got stuck last time. Also we need an algorithm for checksums. 

Though, I think we can have a very basic implementations of these. Instead of base64 we can send decimal representation (raw vector). For checksums can use a bunch of modulos. Not a rocket science. Can give this a try tomorrow. 

@dapritchard 
> So running something like the following code in R once the ESSR objects have been created in the global environment

Yes, we can do that. But it's still a hack IMO. The core problem is the transmission of our files to the remote.

vspinu:
I recalled what lead to the current logic.. I think sending large files through TRAMP was ok and pretty fast, so there was no need to bother with workarounds. Thus by default with TRAMP we source file by file. With `ess-remote` we also source file by file, but it's chunked to avoid sending large amounts of code. This worked ok in my experience, but has the drawback that people who write ESSR code have to take care and split the files with ^L delimiters. 

I will add a few comments in the code. 

> but taking a little more care to make sure that everything is back-compatible with older versions of R such as is done in .load.R.

Bound good but can we keep that code as part of ESSR itself and reuse existing code, not as string in elisp?

dapritchard:
@vspinu I've updated this PR based on this prior comment:

> Sound good but can we keep that code as part of ESSR itself and reuse existing code, not as string in elisp?

With this PR, I think that all codepaths should result in the ESSR objects being loaded into an environment. Previously one of the codepaths for the ESSR load on remote machines resulted in the ESSR objects being placed in the global environment, which lead to the original issue with the R help in #1047 after some other changes in the codebase.

A few somewhat unrelated notes:
1. Since the ESSR code has been changed, the ESSR version number should probably be bumped (and the corresponding rds object rebuilt and git tag added). I haven't done any of that.
2. I refactored the `ess-r--fetch-ESSR-remote` function very slightly to make it more amenable to testing. Previously, if the ESSR GitHub load failed it would call `ess-r--load-ESSR-remote` directly. Now it returns `t` if the load was successful and `nil` otherwise, and relies on a wrapper function to call `ess-r--load-ESSR-remote` upon failure. This makes it easier to test whether a failure occurred rather than having to check the *Messages* buffer (or stdout in the batch case).
3. The test for `ess-r--fetch-ESSR-remote` is skipped on the CI, but can be run locally.
4. I think this PR satisfies the original issue with #1047, but there seems to be a secondary problem that has come up so that issue should probably not get automatically closed.

etc/ESSR/R/.load.R
```diff
@@ -30,14 +30,25 @@
     VERSION <- "1.7"
     assign(".ess.ESSRversion", VERSION, envir = ESSR)
 
+    ## either source the ESSR files if a directory is provided, or otherwise try
+    ## to move any existing R objects into the ESSR environment
+    if(!is.null(dir)) {
```
vspinu, line 35: It's a bit confusing as two distinct functionalities are packed in one function, especially if the name of the function does not represent what it does. 

What I meant with my comment is to extract the moving-to-ESSR-enviornment into a function. Place it in this file along side .ess.load.ESSR.  Maybe `.ess.collect.ESSR.objects` (should contain dots only to be compatible with the pre-historic versions of R).

dapritchard: I've added some commits that split out the parts of `.ess.load.ESSR` that are needed for reuse in the new `.ess.collect.ESSR.objects` function, and then create the new function.

vspinu:
Finally got down to it. Looks good. Thanks a bunch! The tests are great. I have added a few minor simplifications here and there.

dapritchard:
If this is waiting on me for anything please let me know. Happy to make any further changes as needed.

zauster:
Thanks @dapritchard, works like a charm!

PR #1108, opened 2021-02-05
Interrupt process asynchronously when background command exits early

### LINKED ISSUE #1091: C-level Debugging of R now quickly gets to "no process"
```
C-u M-x R  <Enter>   -d gdb  <Enter>
```
open `gdb` in the `*R*` buffer (and you must typically switch there yourself).
There, you type  `r`  at the gdb prompt  `(gdb)`  so indeed, R starts running seemingly regularly.

Now, e.g., you press `C-c C-c`  with which you go into the debugger and then e.g.,
```
   C-c C-c
Program received signal SIGINT, Interrupt.
0x00007ffff7160f7a in select () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install bzip2-libs-1.0.8-2.fc32.x86_64 libgcc-10.2.1-9.fc32.x86_64 libgfortran-10.2.1-9.fc32.x86_64 libgomp-10.2.1-9.fc32.x86_64 libicu-65.1-2.fc32.x86_64 libstdc++-10.2.1-9.fc32.x86_64 ncurses-libs-6.1-15.20191109.fc32.x86_64 pcre2-10.36-1.fc32.x86_64 readline-8.0-4.fc32.x86_64 sssd-client-2.4.0-1.fc32.x86_64 xz-libs-5.2.5-1.fc32.x86_64
(gdb) br do_names
Breakpoint 1 at 0x45cfb0: file ../../../R/src/main/attrib.c, line 1016.
(gdb) c
Continuing.

> 
```
with 'c' ('continue')  I go back to the R  prompt after having set a (C) breakpoint at the 
`do_names()` C function which is called from R's   names()  function.

This is all as it has worked for many years with ESS and R.

But since relatively recently (I discovered ~ a week ago; but I had n't tried to run R in the C level debugger for quite a few weeks),  very quickly the R session  seems to  "lose the process":
The emacs status line then suddenly changes from

>    (iESS  [R]  run   Eldoc) 

to

>    (iESS  [R]  no process   Eldoc) 

and I also get a message in the minibuffer

>  Current buffer has no process

and from then on,  I cannot run any command anymore in the  `iESS` buffer.

It may be that this only happens when I press  `[Tab]`  for R completion of function names, etc.

### LINKED ISSUE #1102: Current buffer has no process

lionel-:
We can't wait for the process to interrupt because:

1. Blocking on interrupt causes hangs when typing or moving cursor. That's because the background completion commands (eldoc, company) are aborted by `while-no-input`.
2. Remote processes might be slow to respond to the interrupt. In that case blocking might trigger a timeout which restores the user's process buffer too early and causes the output of the background command to spill in the buffer.

Follow up to #1106.
Closes #1091.
Closes #1102.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1359,9 +1359,12 @@ wrapping the code into:
 (defun ess--command-proc-restore (proc restore-alist)
   (process-put proc 'cmd-output-delimiter nil)
   (process-put proc 'cmd-restore-alist nil)
-  (set-process-buffer proc (alist-get 'old-pb restore-alist))
-  (set-process-filter proc (alist-get 'old-pf restore-alist))
-  (set-marker (process-mark proc) (alist-get 'old-pm restore-alist)))
+  (let ((old-pb (alist-get 'old-pb restore-alist))
+        (old-pf (alist-get 'old-pf restore-alist))
+        (old-pm (alist-get 'old-pm restore-alist)))
+    (set-process-buffer proc old-pb)
+    (set-process-filter proc old-pf)
+    (set-marker (process-mark proc) old-pm old-pb)))
 
 ;; TODO: Needs some Julia tests as well
 (defun ess--foreground-command (cmd &optional out-buffer _sleep no-prompt-check wait proc)
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -420,7 +420,8 @@ Return non-nil if the process is in a ready (not busy) state."
           (when (re-search-forward (concat "^\\(" sentinel "-END[\n\r]+\\)") nil t)
             (delete-region (match-beginning 0) (match-end 0))
             (process-put proc 'busy nil)
-            (process-put proc 'output-delimiter nil)))))))
+            (process-put proc 'cmd-output-delimiter nil)
+            t))))))
 
 (defun inferior-ess--sentinel-start-re (sentinel)
   (concat "^\\(" sentinel "-START$\\)"))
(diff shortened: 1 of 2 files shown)


lionel-:
Merging now to fix the dev version as I'm moderately confident this resolves the linked issues.

Feel free to leave a review or comments though!

lisp/ess-inf.el
```diff
@@ -1002,16 +1003,24 @@ Returns nil if TIMEOUT was reached, non-nil otherwise."
...
-    (unless delim
+  (ess--if-verbose-write-process-state proc string "ordinary-filter")
+  (let ((cmd-delim (process-get proc 'cmd-output-delimiter))
+        (cmd-async-restore-alist (process-get proc 'cmd-async-restore-alist))
+        (flush (lambda () (with-current-buffer (process-buffer proc)
+                       (insert string)))))
+    (if cmd-delim
+        (progn
+          (funcall flush)
+          (inferior-ess--set-status-sentinel proc (process-buffer proc) cmd-delim)
```
vspinu, line 1014: I would still say that while things are changing so much now, let's not be redundant here. Basically, 

```
(inferior-ess--set-status proc string)
(inferior-ess-run-callback proc string)
(when cmd-delim 
  ... do cmd specific stuff)
```
Kinda want to figure this part while it's hot now, and not break user's code later.

vspinu:
Confirming that this seem to have solved the issue. Both with short timeout, and with user interrupts on C-g. Hurrah!!

vspinu:
Actually not, on very quick C-g C-g the proc buffer is still not returned :(

PR #1111, opened 2021-02-06
Small stylistic simplifications to the process-restore logic

vspinu:
@lionel-  I am trying to catch-up. Before we move further on this I would like to propose a few stylistic simplifications.

I think restoring in ordinary filter is the right aproach, but it is still not quite working and it's unclear why.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -1003,11 +1003,12 @@ Returns nil if TIMEOUT was reached, non-nil otherwise."
     (< elapsed timeout)))
 
 (defun inferior-ess-ordinary-filter (proc string)
+  "Filter used in `ess-command'.
+PROC and STRING are as in `ess-command'."
   (ess--if-verbose-write-process-state proc string "ordinary-filter")
   (let ((cmd-delim (process-get proc 'cmd-output-delimiter))
-        (cmd-async-restore-alist (process-get proc 'cmd-async-restore-alist))
         (flush (lambda () (with-current-buffer (process-buffer proc)
-                       (insert string)))))
+                            (insert string)))))
     (if cmd-delim
         (progn
           (funcall flush)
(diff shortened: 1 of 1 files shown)


lionel-:
> @lionel- I am trying to catch-up. Before we move further on this I would like to propose a few stylistic simplifications.

Thank you. Unfortunately I've already moved further in https://github.com/lionel-/ESS/commit/3cabdc51 :) Would you mind waiting for this work to be merged?

A few comments already:

- I think it makes sense to keep a `cmd-` prefix for everything related to `ess-command`. I've added more of these cmd prefixed properties in the linked commit.
- I like plist for user-facing data structures, but for the backend alist are more robust to programming errors. It's harder to inverse keys and values for instance.

vspinu:
> I think it makes sense to keep a cmd- prefix for everything related to ess-command

The process-restore property might be used in the future from other places as well. I don't see it as `ess-command` specific. It's more like `oridinary-filter` specific.

> t's harder to inverse keys and values for instance.

 For complex nested data structures, yes of course, but here not relevant for 3 element plist. The benefit of plist+keywords in this concrete example is that it's much nice for the code. 

> Unfortunately I've already moved further in lionel-@3cabdc5 :) Would you mind waiting for this work to be merged?

Sure go head, but I wanted to discuss that exact piece of code first. My hunch is that we can do without that search, but it's just a hunch. The start string also is problematic when command output is displayed to the user. (e.g. C-c C-e C-e).

Maybe it would be a bit easier to rebase for you, as I have more small changes here. Not a big deal either way.

lionel-:
> The process-restore property might be used in the future from other places as well. I don't see it as ess-command specific. It's more like oridinary-filter specific.

Good point but I think it's unclear that this data structure will be generic for other cases. How about we store a closure to be called once the proc is available again?

> For complex nested data structures, yes of course, but here not relevant for 3 element plist. The benefit of plist+keywords in this concrete example is that it's much nice for the code.

ok let's switch to plists.

vspinu:
> How about we store a closure to be called once the proc is available again?

Not bad, cleaner and more generic. Basically moving `ess--command-proc-restore` into a lambda locally. No plists/alists required. I like it.

PR #1128, opened 2021-03-17
Keep trailing slash on default-directory from ess-synchronize-dirs

BrendanHalpin:
default-directory should have a trailing slash. This is restored by wrapping the value ess-synchronize-dirs returns with file-name-as-directory.


lionel-:
Thanks!

PR #1178, opened 2022-01-19
Set background mode envvar on process startup

lionel-:
To be used by formatting packages like cli (https://github.com/r-lib/cli/) to select correct colours for the background shade (light or dark) used in Emacs.

cc @gaborcsardi.

### DIFF
--- a/lisp/ess-inf.el
+++ b/lisp/ess-inf.el
@@ -594,6 +594,11 @@ process-less buffer because it was created with
          (nconc
           (list "STATATERM=emacs"
                 (format "PAGER=%s" inferior-ess-pager))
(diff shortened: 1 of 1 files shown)


vspinu:
Would it make sense to use a more generic name like `R_CLI_BACKGROUND_MODE`?

lionel-:
hmm, this may be less generic? For instance Julia packages could also make use of that envvar.

vspinu:
Indeed. In fact it would make sense to have this emacs-wide maybe.

vspinu:
@lionel- can we merge this?

mmaechler:
before it becomes irrelevant ... -- merged *after* release of  ess 25.01.0 -- to be tested now...

PR #1181, opened 2022-01-25
[Fix #1179] Add support for R projects

### LINKED ISSUE #1179: Add support for R projects and integrate with `project.el`
We currently have project.el integration for R packages as part of [ess-r-package.el](https://github.com/emacs-ess/ESS/blob/3a919560e701c3fc19dc7e03e49f0950eb120b6c/lisp/ess-r-package.el#L1) and we have experimental support for starting R in project root [here](https://github.com/emacs-ess/ESS/blob/3a919560e701c3fc19dc7e03e49f0950eb120b6c/lisp/ess-inf.el#L629-L631)

Proposal: 

  1. Add project.el support for arbitrary R projects (those containing `.Rprofile` or `XYZ.Rproj`) 
  2. Make R (and others) start in project root by default
 
Once the above is done can look into session management functionality as provided by sesman (#914, #1172 cc @tpeacock19).

@lionel- proposed (2.) years ago and if I recall correctly we even tried it once, but there were a few edge cases. I think the edge cases are really edge and the benefits outweigh the drawbacks. 

A quick local hack which works for me

```el


;;;*;;; Project detection

(defvar-local ess-r-project--info-cache nil
  "Current package info cache.
See `ess-r-project-info' for its structure.")

(defun ess-r-project (&optional dir)
  "Return the current project as an Emacs project instance.
R project is a directory containing either .Rprofile or XYZ.Rproj
file. Return a list of the form (:name \"XYZ\" :root
\"/path/to/project\"). If DIR is provided, the project is
searched from that directory instead of `default-directory'."
  (let ((info (ess-r-project-info dir)))
    (when (car info)
      (cons 'ess-r-project (plist-get info :root)))))

(cl-defmethod project-roots ((project (head ess-r-project)))
  "Return the project root for ESS R projects."
  (list (cdr project)))

(cl-defmethod project-root ((project (head ess-r-project)))
  "Return the project root for ESS R projects."
  (cdr project))

(defun ess-r-project-info (&optional dir)
  "Get the description of the R project in directory DIR.
Return an plist with the keys :name and :root. When not in a
project return '(nil). This value is cached buffer-locally for
efficiency reasons."
  (if (and (null dir)
           (car ess-r-project--info-cache))
      ess-r-project--info-cache
    (let ((dir (locate-dominating-file (or dir
                                           (buffer-file-name)
                                           default-directory)
                                       ".Rprofile")))
      (when dir
        (let ((dir (directory-file-name dir)))
          (setq ess-r-project--info-cache
                (list :name (file-name-nondirectory dir) :root dir)))))))
```

vspinu:
Project XYZ is defined by `XYZ/.Rpfofile`, `XYZ/DESCRIPTION` or `XYZ/XYZ.Rproj` unless it's the user's home directory.

### DIFF
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -588,6 +588,9 @@ will be prompted to enter arguments interactively."
                   inferior-R-args " "   ; add space just in case
                   start-args))
          (debug (string-match-p " -d \\| --debugger=" r-start-args))
+         (project-find-functions (if (memq 'ess-r-project project-find-functions)
+                                     project-find-functions
+                                   (cons 'ess-r-project project-find-functions)))
          use-dialog-box)
     (when (or ess-microsoft-p
               (eq system-type 'cygwin))
@@ -617,21 +620,6 @@ will be prompted to enter arguments interactively."
   (interactive "P")
   ;; FIXME: Current ob-R expects current buffer set to process buffer
   (set-buffer (run-ess-r start-args)))
-
-(defun inferior-ess-r--adjust-startup-directory (dir dialect)
-  "Adjust startup directory DIR if DIALECT is R.
-If in a package project, prefer the tests directory but only if
-the package directory was selected in the first place."
-  (if (string= dialect "R")
-      (let* ((project-dir (cdr (ess-r-package-project)))
-             (tests-dir (expand-file-name (file-name-as-directory "tests")
-                                          project-dir)))
-        (if (and project-dir
-                 (string= project-dir dir)
-                 (string= default-directory tests-dir))
-            tests-dir
-          dir))
-    dir))
 
 (defun inferior-ess-r--init-callback (_proc _name)
   (R-initialize-on-start))
@@ -870,7 +858,7 @@ efficiency reasons."
                                       (file-exists-p (expand-file-name "DESCRIPTION" dir))
                                       (let ((nm (file-name-nondirectory (directory-file-name dir))))
                                         (file-exists-p (expand-file-name (concat nm ".Rproj") dir))))))))
-                      (when dir 
+                      (when dir
                         (let ((dir (directory-file-name dir)))
                           (unless (member dir (list "~" (getenv "HOME")))
                             (list :name (file-name-nondirectory dir)
--- a/lisp/ess-r-mode.el
+++ b/lisp/ess-r-mode.el
@@ -796,6 +796,7 @@ top level functions only."
   (add-hook 'completion-at-point-functions #'ess-r-package-completion nil 'local)
   (add-hook 'completion-at-point-functions 'ess-filename-completion nil 'local)
   (add-hook 'xref-backend-functions #'ess-r-xref-backend nil 'local)
+  (add-hook 'project-find-functions #'ess-r-project -90 'local)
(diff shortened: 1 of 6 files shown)


lisp/ess-r-mode.el
```diff
@@ -796,6 +796,7 @@ top level functions only."
   (add-hook 'completion-at-point-functions #'ess-r-package-completion nil 'local)
   (add-hook 'completion-at-point-functions 'ess-filename-completion nil 'local)
   (add-hook 'xref-backend-functions #'ess-r-xref-backend nil 'local)
+  (add-hook 'project-find-functions #'ess-r-project -90 'local)
```
lionel-, line 799: This is probably worth a newfeat item.

How is the `-90` depth determined?

lionel- (approved):
Should we also look into turning on `ess--enable-experimental-projects` by default to set the correct startup directory? I've been using this for a few years now.

IIRC the concern was that if people have their home directory versioned, and thus recognised by `(project-current)`, ESS always starts in `~` if there is no inner projects. Maybe we could call these new functions instead of `project-current`. Or have our own configurable list of project finding functions.

vspinu:
> Should we also look into turning on ess--enable-experimental-projects by default to set the correct startup directory? I've been using this for a few years now.

That's the idea indeed.  I will add that with the doc and newfeat changes once we agree on the details of the project lookup.

> ESS always starts in ~ if there is no inner projects

This is taken care by explicitely blacklisting user's home [here](https://github.com/emacs-ess/ESS/pull/1181/files#diff-7f5fba8d706aa5303d67b819b820d1a659a2d2cefec30b56407a18c9a7347944R875). I think it's fine for now. Later we maybe could add an explicit blacklist of directories.

vspinu:
> IIRC the concern was that if people have their home directory versioned,

This is a strange use case but we could indeed either just call our function directly or maybe better blacklist `~` after `project-current` call.

lisp/ess-r-mode.el
vspinu: It's arbitrary, the doc says the default priority is 90, so I just put -90 to make sure that our function comes first. Someone else can impose higher priority with values < -90.

vspinu: I am also thinking it should be part of global `project-find-functions` but not very sure.

vspinu:
@lionel- I think I am done with activating it by default. 

I removed the super ugly special case for starting R in test directory. Whenever starting in test dir is required the user can navigate to test dire manually on process startup or with setwd or with `C-c C-e /`. Not a big deal I think.

lionel-:
> I removed the super ugly special case for starting R in test directory. Whenever starting in test dir is required the user can navigate to test dire manually on process startup or with setwd or with C-c C-e /. Not a big deal I think.

I agree, this is what I'm doing as well. I have keybindings for setting wd to the project dir or to the current dir, which I use relatively frequently.

lionel- (approved):

mmaechler:
I think it is since this change that I get a warning message (here cited from `*Messages*`), i.e., the text *and* a bit of a timeout:
```
Making project-find-functions buffer-local while locally let-bound!
```

Issue #1163, opened 2021-10-03
ess-remote causes various `unexpected symbols` errors in R evaluations.


bryce-carson:
I use `ess-remote` in my daily R work on a cluster. I need to allocate resources for interactive jobs, so I can't use other methods of interacting with remote R processes, so:

1. `M-x shell` and indicate `/bin/bash` as the shell type.
2. I load an environment module so that I have `R`, and then I call `R`.
3. `M-x ess-remote` and I select R as the process type, and then the following errors are produced.

The previous workspace is _not_ the culprit of the issue, as I have tried `R --vanilla` before and I get the same behaviour. Usually I simply call `M-x ess-remote` and select R until I can do so and none of the errors are returned in the `*shell*` process buffer.

```R
[Previously saved workspace restored]

> options(pager='cat')
options(pager='cat')
> 
> ## Hacked help.start() to use wit
> .ess_help_start <
+     home <- if (is.nul
+                 port <- 
Error: unexpected symbol in:
"    home <- if (is.nul
                port"
>                 if (port > 
+                     if (up
+                         make.packages.html(temp=TRUE)
Error: unexpected symbol in:
"                    if (up
                        make.packages.html"
>    
>         
>                 else stop(".ess_help_start() requires the HTTP server to be runni
Error: unexpected 'else' in "                else"
>  
> 
>     paste0(home, "/doc/html/index.html")
Error in paste0(home, "/doc/html/index.html") : object 'home' not found
> }
Error: unexpected '}' in "}"
> 
> 
> 
> .ess.rdired <- f
Error: object 'f' not found
>     cat('\n')

> 
>     objs <- 
+     if (length(ob
+         cat("No object
Error: unexpected symbol in:
"    if (length(ob
        cat"
>         return(invisible
+     }
Error: unexpected '}' in:
"        return(invisible
    }"
> 
>     names(objs) <- objs
Error: object 'objs' not found
>     objs <- lapply(objs, get, envir = .GlobalEnv)
Error in lapply(objs, get, envir = .GlobalEnv) : object 'objs' not found
> 
>     mode <- sapply(objs, data.class)
Error in lapply(X = X, FUN = FUN, ...) : object 'objs' not found
>     length <- sapply(objs, l
+  
+ 
+  
+     var.names <- row.names(d)
Error: unexpected symbol in:
" 
    var.names"
>     
>     ## If 
>     quotes <- rep
>     spaces <- grep(' ', var
+     if (any(spaces))
[...]

Issue #1142, opened 2021-04-13
Error whie running `ess-remote`


ledohod:
Hi,

When open a `shell`, `ssh` into a remote server, then start `R` and run `ess-remote`, I get the following error output in the `iESS` buffer. It looks like some libraries are missing or something. I've installed `ESS` using `straight`. The installation details are at the end. This makes Emacs unusable as Emacs keeps hanging every few seconds for a long time every time.
```

R version 3.6.1 (2019-07-05) -- "Action of the Toes"
Copyright (C) 2019 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> options(pager='cat')
> Error: object 'unquote' not found
> Error: unexpected '}' in "    }"
> Error: object '.ess.command' not found
> Error: object 's' not found
> Error: object 'i' not found
> + + + Error: unexpected symbol in:
"        if (sinked
            sink"
> Error: unexpected '}' in "    }"
> Error: object 'wri' not found
> Error: object 'func' not found
> + > Error: unexpected '}' in "    }"
> Error in withVisible(expr) : object 'expr' not found
> Error: unexpected ')' in "    )"
> + + + + + Error: unexpected '}' in:
" 
}"
> Error: object '.ess.' not found
> Error: unexpected '}' in "}"
> + + > + + > > Error: unexpected '}' in "}"
> + Error: object 'classesNs' not found
> + Error: object 'newClasses' not found
> Error: object 'allMethodTables' not found
> Error in as.character(x) : 
  cannot coerce type 'closure' to vector of type 'character'
> Error: object 'methodsN' not found
> + Error: unexpected symbol in:
"    for (i in seq
        table"
> Error: object 'tableE' not found
> + + + + Error: unexpected '}' in:
"              
        }"
> + Error: unexpected 'if' in:
"            inserted <- .ess.ns_insertMethods(ta
            if"
> + Error: unexpected '}' in:
"                newMethods <
        }"
> + Error: unexpected symbol in:
[...]

PR #1182, opened 2022-01-31
Improve support for `R` sessions started with `ess-remote`

dankessler:
As discussed in #1163 and #1142, there are some issues with `R` sessions invoked via `ess-remote` as [discussed in the manual](https://ess.r-project.org/Manual/ess.html#ESS_002dremote) (i.e., launch a shell within emacs using `M-x shell`, ssh to a server, run `R`, and then do `M-x ess-remote` to get things setup).

The most pressing issue is that some literal control codes in files that are injected during configuration can get intercepted by the shell which screws up subsequent parsing and can have weird consequences. A simple fix for this is to use octal codes rather than literals in the associated R files, and this is all that 2ce1846488544fe9afe6ce24c502185fab6ddff0 does. However, that alone is not sufficient to get the associated functionality working correctly, but if the rest of this PR requires further discussion and a quick fix is desired, I'm happy to submit a PR that includes* only this small commit.

The subsequent commits in this PR are aimed at improving MPI support for `R` sessions started with `ess-remote` as discussed above. I enumerate the changes/commits below, since in general they are all severable.

* In d9443009347caaa779420986191dd1a830db4893, I switch to different control codes and in particular use those associated with separators to reduce the chances of a shell ascribing them special meaning. Note: I use literals in the `elisp` code so in Github it renders like an empty string, but it's easier to see in emacs.

* In a5451413b939aebd1e629ca189510b7b27e9d2d8, change the message handler to override the `read-only` property which prevents it from cleaning up after itself when it runs in a `*shell*` buffer.

* In f5f54f27238d0058b5346cbe2b76751030ee45a1, I (conditionally) invoke `ess-tracebug` when setting things up in a manner analogous to that done by `R-initialize-on-start` (which `ess-remote` doesn't call for what I assume are good reasons unknown to me). This is necessary to set off a chain of activity which ultimately results in `ess-mpi-handle-message` getting run automatically in `R` sessions started with `M-x ess-remote` in the same way that it does in `R` sessions started with `M-x R`.

### DIFF
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -1257,7 +1257,8 @@ ends with an incomplete message."
                           (apply handler payload))
                       (error "No handler defined for MPI message '%s" head))
                   (goto-char mbeg0)
-                  (delete-region mbeg0 mend0)))
+                  (let ((inhibit-read-only t))
+                    (delete-region mbeg0 mend0))))
             (setq out :incomplete))))
       out)))
 
--- a/lisp/ess-tracebug.el
+++ b/lisp/ess-tracebug.el
@@ -1183,9 +1183,9 @@ Kill the *ess.dbg.[R_name]* buffer."
 
 ;; http://jkorpela.fi/chars/c0.html
 ;; https://en.wikipedia.org/wiki/ANSI_escape_code#Escape_sequences
-(defvar ess-mpi-message-start-delimiter "_")
+(defvar ess-mpi-message-start-delimiter "")
 (defvar ess-mpi-message-field-separator "")
-(defvar ess-mpi-message-end-delimiter "\\")
+(defvar ess-mpi-message-end-delimiter "")
 
 (define-obsolete-variable-alias 'ess-mpi-alist 'ess-mpi-handlers "ESS 19.04")
 (defvar ess-mpi-handlers
@@ -1229,7 +1229,7 @@ value from EXPR and then sent to the subprocess."
 
 (defun ess-mpi-handle-messages (buf)
   "Handle all mpi messages in BUF and delete them.
-The MPI message has the form TYPEFIELD... where TYPE is the
+The MPI message has the form TYPEFIELD... where TYPE is the
 type of the messages on which handlers in `ess-mpi-handlers' are
 dispatched. And FIELDs are strings. Return :incomplete if BUF
 ends with an incomplete message."
--- a/etc/ESSR/R/mpi.R
+++ b/etc/ESSR/R/mpi.R
@@ -7,7 +7,7 @@
         else as.character(el)
     })
     payload <- paste(dots, collapse = "")
-    cat(sprintf("\033_%s036%s\033\\", head, payload))
+    cat(sprintf("\034%s\036%s\035", head, payload))
 }
 
 .ess_mpi_message <- function(msg){
--- a/etc/ESSR/R/mpi.R
+++ b/etc/ESSR/R/mpi.R
@@ -7,7 +7,7 @@
         else as.character(el)
     })
     payload <- paste(dots, collapse = "")
-    cat(sprintf("_%s%s\\", head, payload))
+    cat(sprintf("\033_%s036%s\033\\", head, payload))
 }
 
 .ess_mpi_message <- function(msg){
--- a/lisp/essd-els.el
+++ b/lisp/essd-els.el
@@ -154,7 +154,8 @@ DIALECT is the desired ess-dialect. If nil, ask for dialect"
       (ess-eval-linewise (format "options(pager='%s')\n"
                                  (or inferior-ess-remote-pager inferior-ess-pager))
                          nil nil nil 'wait)
-      (ess-r-load-ESSR))
+      (ess-r-load-ESSR)
+      (when ess-use-tracebug (ess-tracebug 1)))
 
     (when (equal ess-dialect "S+")
       (ess-command ess-S+--injected-code))


vspinu:
Thanks! Looks good. But why these characters and in that order? Have you tried the ones proposed by Lionel 001 - start of header, 002 - start of text, 003 - end of text? It would semantically map to the MPI protocol. 

Though the GS,  RS and US codes were designed to send tabular data https://stackoverflow.com/a/18782271/453735

How about this protocol `GS RS head US payload RS GS`? This way we could add more records in the future if needed.

Could you also better document the protocol and include the links to relevant info? 

https://stackoverflow.com/a/18782271/453735
https://www.sciencebuddies.org/science-fair-projects/references/ascii-table

dankessler:
> Have you tried the ones proposed by Lionel 001 - start of header, 002 - start of text, 003 - end of text? It would semantically map to the MPI protocol.

I hadn't tried because I was nervous about the shell being confused by `\003` specifically, since it's what is sent by `C-c` to interrupt a process. However, it was lazy of me to not test, so I went ahead and tried it and it seems to work fine (tested with `M-x R` and `M-x shell` using both `bash` and `zsh` on a MacBook Pro).

So, I propose we go with the protocol suggested by @lionel- , specifically `SOH head STX payload ETX`

On the other hand, we could go further down the road proposed by @vspinu and consider the more extensible protocol of `GS RS head US payload RS GS`, although I don't know if there's value in being able to pack multiple messages into a single group that doesn't have its own header as opposed to just sending them atomically (although perhaps the overhead is different if there are many, many messages to send). My hesitation with this protocol (and the reason I had done something semantically weird to do something like `ETX` without actually using it) was concern about using the same delimiter for both start and end and a fear that the handler might get confused and bad things could happen if it thinks an end is a start and vice versa, but maybe that anxiety is misplaced and/or can be addressed. 


In my opinion, the `SOH head STX payload ETX` approach seems clearer (and can of course be revisited down the road if `MPI` starts seeing a lot of use), but I'm happy to do whatever is preferred.

> Could you also better document the protocol and include the links to relevant info?
> https://stackoverflow.com/a/18782271/453735 https://www.sciencebuddies.org/science-fair-projects/references/ascii-table

Sure; once we settle on delimiters and a protocol I can do that. Would the [docstring for `ess-mpi-handle-messages`](https://github.com/emacs-ess/ESS/blob/f5f54f27238d0058b5346cbe2b76751030ee45a1/lisp/ess-tracebug.el#L1231-L1235) be a good place to put this documentation, or are you thinking something more formal like the manual?

Also, while I'm at it, I'm planning to use octal codes in the elisp too where I can, just so that people looking at the code on e.g., GitHub are less confused by the non-printing characters.

lionel-:
> Sure; once we settle on delimiters and a protocol I can do that. Would the docstring for ess-mpi-handle-messages be a good place to put this documentation, or are you thinking something more formal like the manual?

Just a note that I think all of this should be treated as internal for now, so the manual would not be a good place. Probably comments would be fine for this doc. (For the same reason I'd prefer things like the mpi regex variables to be double-slash prefixed.)

dankessler:
>  (For the same reason I'd prefer things like the mpi regex variables to be double-slash prefixed.)

Do you mean just for display purposes, or would you like the regex to actually include a literal slash? I.e., do you mean like this?

```
(defvar ess-mpi-message-start-delimiter "\001")
(defvar ess-mpi-message-field-separator "\002")
(defvar ess-mpi-message-end-delimiter "\003")
```

or like this?
```
(defvar ess-mpi-message-start-delimiter "\\\001")
(defvar ess-mpi-message-field-separator "\\\002")
(defvar ess-mpi-message-end-delimiter "\\\003")

```

lionel-:
Sorry, I meant double-dashed, e.g. `ess--mpi-message-start-delimiter`.

dankessler:
Ah, that makes more sense and I can make that change while I'm touching this code anyway (I presume the convention is that the `ess--` prefix indicates things that are "internal").

dankessler:
I've pushed some new commits that
1. switch to the `SOH header STX payload ETX` protocol
2. adhere to the double-dash convention (at least for any variables/functions I touched; I did not comprehensively refactor)
3. update/add documentation in elisp comments and doctoring for `ess--mpi-handle-messages`

I think this addresses all of the outstanding issues raised in the discussion of this PR and is perhaps good to go.

Feel free to go ahead and accept if this looks good on your end, but let me know if you prefer that I squash this all into a single commit or rebase to tidy things up and force push to update the PR.

lisp/ess-tracebug.el
```diff
@@ -1227,28 +1227,30 @@ value from EXPR and then sent to the subprocess."
    ((string= el "t") t)
    (t el)))
 
-(defun ess-mpi-handle-messages (buf)
+(defun ess--mpi-handle-messages (buf)
```
vspinu, line 1230: Can you leave this name unchanged? If we adopt the convention that internal means non-user visible then everything has to be changed. We are adhering to the narrow meaning of "internal" as "internal details extracted for convenience".

dankessler:
I just realized we never closed the loop on this. How does this protocol  (which @vspinu proposed earlier)?

```
GS RS head US payload RS GS
```

This could later be extended to include multiple key/value pairs like

```
GS RS head1 US payload1 RS head2 US payload2 RS GS
```

So long as we stick to the convention that there's a `RS` *before* the first pair and *after* the last pair, then the start/end delimiters are distinct (`GS RS` and `RS GS`, respectively), which addresses my fear of the delimiters getting unbalanced.

If we like it, I'll go ahead and update the PR to implement this.

aramirezreyes:
It would be interesting to have this merged. The loading of ess-tracebug would give a consistent experience across ess. See, for example #1265

PR #1190, opened 2022-02-25
fix: terminate auto-mode-alist entries

nverno:
Anchor  `auto-mode-alist` entries to match until end of file extension


vspinu:
Thanks!

PR #1202, opened 2022-06-16
new feature: Rd-mode-insert-data-skeleton

iago-pssjd:
Added function `Rd-mode-insert-data-skeleton` to allow insert `Rd`-skeleton for datasets documentation using `C-c C-d`.


vspinu:
Thanks!

PR #1234, opened 2022-12-19
Show number of rows in rdired 

DavidMatthews83:
uses nrow to get the number of rows of an object and display in the R dired buffer


mmaechler:
I've not been using `rdired` myself.  Any volunteers confirming this works for them,  please ?

PR #1244, opened 2023-04-15
Disable background evals when process init failed

lionel-:
These changes are aimed to prevent background eval hangs when ESSR failed to load.

- Add a `bg-eval-disabled` process property. It's set to `t` when ESSR couldn't be loaded.
- Add a `(ess-can-eval-in-background)` defun that checks for the global variable as well as the process property. Now used throughout ESS instead of just checking for the global var.
- Add a `(ess-get-next-available-bg-process)` defun to get a process for which bg evals are allowed. Used in completion helpers.

### DIFF
--- a/doc/newfeat.texi
+++ b/doc/newfeat.texi
@@ -4,6 +4,11 @@
 Changes and New Features in 19.04 (unreleased):
 @itemize @bullet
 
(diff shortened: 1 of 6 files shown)


lionel-:
Merging now but comments are welcome.

PR #1245, opened 2023-04-15
Fix compilation of package installed from Git; plus cosmetics

lionel-:
Patch from @monnier on 2022-11-01. Needs rebase (edit: now done) and some work (see fixmes).

```
In `test/*.el`: Provide "test/etest/etest" to `require` so Emacs can find the file when the package is compiled.

In `*.el`: Prefer #' to quote named functions.
Use `lexical-binding` in all the remaining files.
Use lexical-binding also in `eval`.
Remove redundant let-bindings of obsolete variable `inhibit-point-motion-hooks`.
Replace uses of `point-at-bol` and `point-at-eol`, both marked obsolete in Emacs-29.  Use `defalias` rather than `fset` to define a function. Fix occasional incorrect uses of ' in docstrings.
Remove some redundant `:group` arguments.
Add leading `_` to unused args.

In addition to the above, a few, more specifc, tweaks:

* lisp/ess-r-mode.el (ess-offset, ess-offset-type): Strength-reduce `eval` to `symbol-value`.
(ess-rutils-rsitesearch): Hoist the common `browse-url` out of the `if`.

* lisp/ess-r-package.el: Don't test `fboundp` before `advice-add` since we require Emacs≥25.  Don't require `shell` since `advice-add` also works when the function is not yet defined.

* lisp/ess-r-syntax.el (ess-parser--backward): Give it a global nil value, both to declare it as dynamically scoped and to simplify the code.  All users adjusted not to bother with `bound-and-true-p`. (ess-node): Hoist common `gethash` out of the `if`.

* lisp/ess-sas-a.el (ess-sas-rtf-portrait, ess-sas-rtf-landscape): Add FIXME.

* lisp/ess-sas-d.el (ess-electric-run-semicolon): Re-indent.  Add FIXME.

* lisp/ess-sas-l.el (beginning-of-sas-statement): Rename statically scoped arg to avoid conflict with dynamically scoped var. (sas-prev-line): Make (unused) arg optional.
(ess-imenu-SAS): Use [:alpha:] and [:alnum:].

* lisp/ess-tracebug.el (ess--tb-start, ess--tb-stop): Use `advice-add/remove` to redefine `ess-parse-errors`.

* lisp/ess.el (ess-version-string): Don't presume `point-min` is `1`.

* lisp/obsolete/ess-mouse.el (ess-mouse-me-eval-expanded): Remove unused var `page-scommand`.

* lisp/obsolete/mouseme.el (mouse-me-get-string, mouse-me-find-grep): Remove unnused vars `p`, `beg`, and `end`.

* test/ess-test-inf.el (ess-test-inferior-local-start-args): Use `*proc*` rather than undocumented `*inf-buf*`. (ess-inf-send-fn-test, ess-inf-send-cat-some.text-test): Remove unused var `output-nowait`.

* test/ess-test-literate.el (chunk-end): Remove `defvar`. (elt-process-next-chunk, elt-process-next-subchunk, elt-process-case) (elt-process-code): Pass/take `chunk-end` as argument instead.

* test/ess-test-org.el (test-org-R-ouput): Remove unused var `inf-proc`.

* test/ess-test-r-utils.el (with-ess-test-file): Add FIXME. (ess-test-sleep-while, ess-test-unwind-protect): Move before first use. Split the bulk of the code into a function, to ease debugging, avoid name captures, and improve error messages.
(with-r-running): Split the bulk of the code into a function. Document the `*proc*` var.  Add FIXME.
(output, face-at, token=): Add FIXME.
(output=): Remove redundant `eval`.

* test/ess-test-r.el (ess-r-inherits-prog-mode-test): Turn global `ess-test-prog-hook` var into local `pmh-was-run`.

* test/literate/fontification.el (face-at-point): Remove unused function.

* test/literate/roxy.el (face-at-point, faces-at-point): Remove unused functions.
```


lionel-:
Thanks Stefan!

monnier:
Thank you, Lionel

PR #1257, opened 2023-08-03
fix: quote ess-julia-customize-alist

nverno:
Without quoting this is causing an error when evaling in `ess-setq-vars-local`.


lionel-:
Thanks!

PR #1260, opened 2023-08-11
tracebug: add keybindings for remaining browser commands

dsweber2:
Mirroring the remaining [browser functions](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/browser). I tested this by running these lines in my config, and it worked in the file I had open. There's probably a better way to do help so that it lists these commands too, but hopefully the pattern is clear enough.

The bindings are of course just suggestions. They were unbound in my work environment and roughly followed the conventions of the other tracebug browser controls.

Hopefully this proves useful to someone else.


mmaechler:
It seems "nobody" uses  tracebug ... but the code "looks promising" ... so let's try to put it into (development version of) ess, now that ess 25.01.0 is released.

Please tracebug fans, test the new features (new keys!) and   give  praises  (and bug reports)!

PR #1281, opened 2024-02-29
+ \dontdiff{.}

bastistician:
A new built-in Rd macro `\dontdiff` has been added to the trunk in r85975. The output of enclosed example code will be ignored when comparing to saved reference output (`<pkg>-Ex.Rout.save`).


lionel-:
Thanks!

PR #1288, opened 2024-04-27
Add more available roxygen2 tags to ess-roxy-tags-param

wenjie2wang:
This PR tries to complete the built-in `ess-roxy-tags-params` with a few more roxygen2 tags that are documented in the latest {Roxygen2} package vignettes: <https://roxygen2.r-lib.org/articles/namespace.html> and <https://roxygen2.r-lib.org/articles/rd.html>.  To be more specific, the following five tags are added:

- `@evalNamespace`
- `@example`
- `@examplesIf`
- `@exportS3Method`
- `@returns`

With this PR, ESS will highlight and complete those tags if I understand the variable `ess-roxy-tags-params` correctly. Thanks!

### DIFF
--- a/lisp/ess-custom.el
+++ b/lisp/ess-custom.el
@@ -1427,19 +1427,21 @@ Used to decide highlighting and tag completion."
   :type '(repeat string))
 
 (defcustom ess-roxy-tags-param '("author" "aliases" "concept" "details"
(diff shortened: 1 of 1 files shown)


lionel-:
Thanks!

PR #1298, opened 2024-07-25
fix loading ESSR in R 2.5

bastistician:
Loading ESSR failed in R 2.5.0 (not that I run this regularly,  but it was annoying when I wanted to try something in some older R versions).

The first problem was in `.ess.ESSR.load()` which relied on `nzchar` from R >= 2.6.0. Avoiding this would still not give me a working prompt because `.ess.command()`, which is needed by `ess-r-initialize`, used `paste0()` which "only" exists since R 2.15.0. The corresponding fixes are trivial and attached here.

NB: This PR does *not* address the current incompatibility of ESSR with even older versions of R.

- R < 1.8.0 fails because `:::` in `.ess.fetchParsedRd()` gives syntax errors.
- R < 2.3.0 fails because `emptyenv()` for `.ess.environment.state` does not exist.
- R < 2.5.0 fails because `withVisible()` in `.ess.command()` does not exist.

So just *loading* ESSR is still broken in these old R versions.
It turns out the header comment in `ESSR/R/.basic.R`, "Should work on *all* versions of R", is increasingly hard to fulfil.

### DIFF
--- a/etc/ESSR/R/.basic.R
+++ b/etc/ESSR/R/.basic.R
@@ -80,7 +80,7 @@
 .ess.helpLinks <- function(topic, package) {
     tryCatch(
         warning = function(...) NULL,
-        error = function(...) NULL,
+        error = function(...) NULL, # needs R >= 2.13.0
         {
             ast <- .ess.fetchParsedRd(topic, package)
             .ess.findLinks(ast)
@@ -156,7 +156,7 @@
 }
 
 .ess.strip.error <- function(msg, srcfile) {
-    pattern <- paste0(srcfile, ":[0-9]+:[0-9]+: ")
+    pattern <- paste(srcfile, ":[0-9]+:[0-9]+: ", sep = "")
     sub(pattern, "", msg)
 }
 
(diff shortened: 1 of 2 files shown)


mmaechler:
>>>>> Sebastian Meyer 
>>>>>     on Thu, 25 Jul 2024 05:51:45 -0700 writes:

    > Loading ESSR failed in R 2.5.0 (not that I run this regularly,  but it was annoying when I wanted to try something in some older R versions).
    > The first problem was in `.ess.ESSR.load()` which relied on `nzchar` from R &gt;= 2.6.0. Avoiding this would still not give me a working prompt because `.ess.command()`, which is needed by `ess-r-initialize`, used `paste0()` which &quot;only&quot; exists since R 2.15.0. The corresponding fixes are trivial and attached here.

    > NB: This PR does *not* address the current incompatibility of ESSR with even older versions of R.

    > - R &lt; 1.8.0 fails because `:::` in `.ess.fetchParsedRd()` gives syntax errors.
    > - R &lt; 2.3.0 fails because `emptyenv()` for `.ess.environment.state` does not exist.
    > - R &lt; 2.5.0 fails because `withVisible()` in `.ess.command()` does not exist.

    > So just *loading* ESSR is still broken in these old R versions.

    > It turns out the header comment in `ESSR/R/.basic.R`, &quot;Should work on *all* versions of R&quot;, is increasingly hard to fulfil.

Yes, I wrote the above and I had spent dozens of hours of my
life to keep ESS working even with R 0.16 (IIRC), 
but my other ESS core colleagues did not care so much at the
time.

I agree it's worth the effort to at least go back older versions
than you found working now... and am happy to help.

Martin

Issue #1303, opened 2024-09-02
Symbol's function definition is void: julia-latexsub


samJcrawford:
Seems like [this commit](https://github.com/JuliaEditorSupport/julia-emacs/commit/4d2047666fea43e053ead0c5501781b08520bd98) removed the `julia-latexsub` function which is called by `ess-julia-latexsub-completion`.

`julia-mode` does provide its own pair of capfs for this purpose, namely `julia-mode-latexsub-completion-at-point-before` and `julia-mode-latexsub-completion-at-point-around`. Perhaps ESS could just add both of these to `completion-at-point-functions` for both `ess-julia-mode` and `inferior-ess-julia-mode`?

PR #1304, opened 2024-09-02
Replace julia-latexsub with appropriate capf functions

samJcrawford:
A naive solution to #1303. Not sure if this is the correct way to do things, but it seems the simplest change to make, and has the desired effect on my machine.

### DIFF
--- a/lisp/ess-julia.el
+++ b/lisp/ess-julia.el
@@ -42,7 +42,8 @@
 ;; Don't require `julia-mode' to compile this file.
 (when t (require 'julia-mode))
 (declare-function julia-mode "julia-mode" ())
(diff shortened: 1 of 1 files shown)


mmaechler:
Thank you, Sam.  This looks good,  but as I haven't used  Julia for several years, nor looked at its  emacs / ESS interface recently, I "can" not merge (i.e. "should not"), because I would not (easily be able to) test the change, etc.

Is there any other volunteer, please,  to confirm if the above patch fixes the problem (julia-latex command completion) in #1303 ?   Thanks a lot in advance!

mmaechler:
As we did not find any volunteer,  I have applied to patch now (*after* releasing ess 25.01.0; so "we" can test here, before it gets released).

PR #1325, opened 2025-07-29
Pass buffer to text= for clarity

MichaelChirico:
Also explicitly request `parse_settings=TRUE` for back-compatibility, see https://github.com/r-lib/lintr/pull/2863.

### DIFF
--- a/lisp/ess-r-flymake.el
+++ b/lisp/ess-r-flymake.el
@@ -27,7 +27,7 @@
 ;; Flymake is the built-in Emacs package that supports on-the-fly
 ;; syntax checking.  This file adds support for this in ess-r-mode by
 ;; relying on the lintr package, available on CRAN and currently
-;; hosted at https://github.com/jimhester/lintr.
+;; hosted at https://github.com/r-lib/lintr.
 
 ;;; Code:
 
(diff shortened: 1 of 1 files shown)


lisp/ess-r-flymake.el
```diff
@@ -76,28 +76,26 @@ each element is passed as argument to `lintr::linters_with_defaults'."
...
     } else {
-        if (packageVersion('lintr') <= '3.0.0') {
-            cat('@@error: @@Need `lintr` version > v3.0.0')
-        } else {
-            tryCatch(lintr::lint(commandArgs(TRUE), ...),
-                    error = function(e) {
-                       cat('@@warning: @@', conditionMessage(e))
-                    })
-        }
+        tryCatch(lintr::lint(text=commandArgs(TRUE), ..., parse_settings=TRUE),
```
bastistician, line 90: While at it I think this should be changed to `text = str`, so the function argument is actually used.

MichaelChirico, line 90:
```suggestion
        tryCatch(lintr::lint(text=str, ..., parse_settings=TRUE),
```

mmaechler (approved):

